PR #9, opened 2026-05-26
docs: add design spec for change detection and conflict resolution

sumedhdk22:
## Summary
This pull request introduces the architectural design specification for local change detection and remote conflict resolution in the Metadata as Code tool (`kcmd`). 

The proposed design balances performance, minimal remote network footprint, and workspace cleanliness while ensuring local changes are safely synchronized without data loss.

## Key Proposals & Architecture Details

- **State Database (`.catalog_state.json`)**: Recommends a single, flat, system-managed JSON state database at the workspace root (Option A) over directory-mirroring or co-located checksum files to maximize Node.js native parsing speed and minimize clutter.
- **Pull Workflow & Safety Guard**: Details a strict "dirty workspace guard" that aborts local pull updates if untracked or unpushed modifications are present. Includes semantic normalization logic (recursive alphabetical key sorting, standardized whitespaces, and column order preservation in table schemas) to ensure checksums are resilient to formatting-only changes.
- **Push Workflow**: Outlines a scale-aware, lookup-based push sequence utilizing custom views and a throttled concurrency pool to avoid socket exhaustion, protected by an all-or-nothing integrity guard.
- **Conflict Resolution Matrix**: Defines exact system actions (`Safe to Pull`, `Safe to Delete Locally`, `Safe to Push`, `Safe to Delete Remotely`, `Safe to Create`) for all combinations of local, remote, and missing aspect states.
- **Dry Run (`--dry-run`) Mechanics**: Fully specifies the pull/push dry-run workflows to safely preview local workspace modifications or remote mutation sequences without executing actual disk writes or remote API mutations.

### DIFF
--- a/toolbox/mdcode/docs/checksum_design.md
+++ b/toolbox/mdcode/docs/checksum_design.md
@@ -15,32 +15,51 @@ To enable offline capabilities, minimal network updates, and robust conflict det
 
 ---
 
-## Comprehensive Conflict Resolution Matrix
+## Conflict Resolution
 
-A **conflict** occurs when the local filesystem is not in sync with the Catalog's state of the resource.
+A **conflict** occurs when there are changes made to both the local and catalog copies of the same resource metadata since the last sync.
 
-In the event of a conflict, the user has three choices:
-* **Manual Merge (via Force Pull)**: Back up local modifications, run a force pull (`kcmd pull --force`) to fetch remote changes, and manually merge the desired local changes.
-* **Abandon Local Changes (via Force Pull)**: Run a force pull (`kcmd pull --force`) to discard local modifications and align completely with the remote Catalog state.
-* **Force Update (via Force Push)**: Run a force push (`kcmd push --force`) to overwrite the remote Catalog state with the local workspace state.
+In the event of a conflict, the resolution options depend on the direction of synchronization:
 
-The conflict resolution matrix covers aspect modifications, creations, and deletions:
+### Pull Operation Choices
+* **Manual Merge**: Back up local modifications, run a force pull (`kcmd pull --force`) to fetch remote changes, and manually merge the desired local changes.
+* **Abandon Local Changes**: Run a force pull (`kcmd pull --force`) to discard local modifications and align completely with the remote Catalog state.
 
-| Local Aspect State | Remote Aspect State | Outcome / Action | Description |
-| :--- | :--- | :--- | :--- |
-| **Unchanged** | **Unchanged** | **Skip** | No actions needed. |
-| **Unchanged** | **Modified** | **Safe to Pull** | Safe to pull. Remote changes can overwrite local file on next pull. |
-| **Unchanged** | **Deleted** | **Safe to Delete Locally** | Safe to delete locally on next pull. |
-| **Does Not Exist** | **Exists Remotely** | **Safe to Pull** | Safe to pull/create locally. Local aspect files are written and base checksums are recorded. |
-| **Modified** | **Unchanged** | **Safe to Push** | Push local modifications. Base checksum is updated. |
-| **Modified** | **Modified** | **Conflict** | Conflict! Abort push. User must pull to resolve or use force push. |
-| **Modified** | **Deleted (Entry Deleted)** | **Conflict** | Conflict! Remote entry itself was deleted. Options: (1) Abandon local changes via force pull to delete local files, or (2) Force push to recreate the entry and aspects. |
-| **Modified** | **Deleted (Aspect Deleted)** | **Conflict** | Conflict! Only the aspect was deleted remotely. Options: (1) Abandon local changes via force pull to delete local aspect, or (2) Force push to recreate/push the aspect remotely. |
-| **Deleted** | **Unchanged** | **Safe to Delete Remotely** | Push aspect deletion to remote catalog. Clean up local state entry. |
-| **Deleted** | **Modified** | **Conflict** | Conflict! Remote was modified while local was deleted. |
-| **Deleted** | **Deleted** | **Skip** | Aspect was deleted on both sides. Clean up local state. |
-| **Created (New)** | **Exists Remotely** | **Conflict** | Conflict! Local aspect created but already exists on remote. |
-| **Created (New)** | **Does Not Exist** | **Safe to Create** | Safe to push/create on remote. Base checksum is recorded. |
+### Push Operation Choices
+* **Manual Resolution**: Back up local modifications, run a force pull (`kcmd pull --force`) to retrieve remote updates, manually merge and apply local changes, and push again.
+* **Force Update**: Run a force push (`kcmd push --force`) to overwrite the remote Catalog state with the local workspace state.
+
+The synchronization behaviors and conflict resolutions are split below into operations impacting pull and push workflows.
+
+### 1. Pull Conflict Resolution Matrix
+
+During a `pull` operation, the main objective is to bring remote changes into the local workspace. To protect unpushed work, the standard `pull` command implements a strict "dirty guard" check on the entire workspace first.
+
+| Local Aspect State | Remote Aspect State | Standard Pull (`kcmd pull`) | Force Pull (`kcmd pull --force`) | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| **Unchanged** | **Unchanged** | **Skip** | **Skip** | No changes on either side. |
+| **Unchanged** | **Modified** | **Safe to Pull** | **Safe to Pull** | Remote modifications are safely written to the local workspace. |
+| **Unchanged** | **Deleted** | **Safe to Delete Locally** | **Safe to Delete Locally** | Remote deletion is propagated locally; the local aspect file is deleted. |
+| **Does Not Exist** | **Exists Remotely** | **Safe to Pull** | **Safe to Pull** | The remote aspect is pulled down and created in the local workspace. |
+| **Modified** / **Created** / **Deleted** | **Any State** | **Abort (Dirty Workspace)** | **Overwritten by Remote** | Local changes exist in the workspace. Standard pull aborts immediately to prevent overwriting unpushed local work. Force pull discards local changes and overwrites them with the remote state. |
+
+### 2. Push Conflict Resolution Matrix
+
+During a `push` operation, local modifications are sent to the remote catalog. A standard `push` will abort if any conflict is detected (i.e., if the remote state has changed since the last synchronization).
+
+| Local Aspect State | Remote Aspect State | Standard Push (`kcmd push`) | Force Push (`kcmd push --force`) | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| **Unchanged** | **Unchanged** | **Skip** | **Skip** | No changes on either side. |
+| **Unchanged** | **Modified** | **Skip** | **Skip** | No local changes to push. Remote changes remain until the next pull. |
+| **Modified** | **Unchanged** | **Safe to Push** | **Safe to Push** | Pushes local modifications to the remote catalog. Base checksum is updated. |
+| **Modified** | **Modified** | **Conflict (Abort)** | **Force Overwrite** | Both copies were modified. Standard push aborts. Force push overwrites the remote aspect. |
+| **Modified** | **Deleted (Entry Deleted)** | **Conflict (Abort)** | **Force Create/Overwrite** | The remote entry itself was deleted. Force push recreates the entry and aspect on the remote. |
+| **Modified** | **Deleted (Aspect Deleted)** | **Conflict (Abort)** | **Force Create/Overwrite** | The remote aspect was deleted. Force push recreates/updates the aspect on the remote. |
+| **Deleted** | **Unchanged** | **Safe to Delete Remotely** | **Safe to Delete Remotely** | Pushes the aspect deletion to the remote catalog. Base state is cleaned up. |
+| **Deleted** | **Modified** | **Conflict (Abort)** | **Force Delete Remotely** | The aspect was modified remotely while being deleted locally. Force push deletes the aspect remotely. |
+| **Deleted** | **Deleted** | **Skip** | **Skip** | The aspect was deleted on both sides. Local state base checksum is cleaned up. |
+| **Created (New)** | **Exists Remotely** | **Conflict (Abort)** | **Force Overwrite** | Local aspect created but already exists on the remote. Force push overwrites remote aspect. |
+| **Created (New)** | **Does Not Exist** | **Safe to Create** | **Safe to Create** | Pushes the new aspect to the remote catalog. Base checksum is recorded. |
 
 ---
 
@@ -72,42 +91,35 @@ graph TD
 
 ---
 
-## The State Database (`.catalog_state.json`)
+## The State Database (`catalog-state.json`)
 
 A flat JSON database maps each tracked entry and its corresponding aspects to their last synchronized state.
 
 ### Checksum Scope & Modifiability Constraints
 
 Before discussing database layout options, we define the scope and constraints of the tracked metadata:
-* **Aspect-Level Only**: This tool does not support entry-level metadata modifications (e.g., `displayName`, `description` defined directly on the Entry resource). Therefore, no entry-level checksums are calculated or tracked.
+* **Entry-Level & Aspect-Level Checksums**: The tool tracks both a unified **Entry-Level Checksum** and individual **Aspect-Level Checksums**. This is critical for user-managed or custom entries where the entry itself can be created or deleted, and for uningested entry groups where entry-level metadata (such as `entrySource`, `displayName`, or `description`) is modifiable.
 * **Aspect `checksum`**: These checksums are calculated for the key-value properties inside each individual aspect type (e.g., `dataplex-types.global.overview`, `dataplex-types.global.descriptions`).
-* **Static Modifiability Enforcement**: The `catalog.yaml` validation layer statically ensures that only modifiable aspects (e.g., non-required aspects for ingested entries, or any aspect for custom/user-managed entries) are configured to be synchronized and published.
+* **Static Modifiability Enforcement**: The `catalog.yaml` validation layer statically ensures that only modifiable aspects and entry properties are configured to be synchronized and published.
 
-### Proposed Design: Flat State File (`.catalog_state.json`)
+### Proposed Design: Flat State File (`catalog-state.json`)
 
-A single JSON file at the root of the workspace containing a map of all tracked files, their base aspects, and their checksums.
+A single JSON file at the root of the workspace containing a map of all tracked entries, their unified entry-level base checksum, their last sync timestamps, and individual base aspect checksums.
 
 > [!NOTE]
 > **Why JSON over YAML for system state?**
-> Since `.catalog_state.json` is entirely system-managed and gitignored, human readability is secondary to execution performance and parsing reliability. Node.js has native, highly-optimized C++ support for parsing and serializing JSON (`JSON.parse` / `JSON.stringify`). Utilizing YAML would require external JavaScript-based parser libraries, which introduce significant CPU and execution latency during CLI startup as the catalog scales.
+> Since `catalog-state.json` is entirely system-managed and gitignored, human readability is secondary to execution performance and parsing reliability. Node.js has native, highly-optimized C++ support for parsing and serializing JSON (`JSON.parse` / `JSON.stringify`). Utilizing YAML would require external JavaScript-based parser libraries, which introduce significant CPU and execution latency during CLI startup as the catalog scales.
 
 ```json
 {
   "version": "1",
   "entries": {
     "ecommerce-prod.ecommerce-dataset/orders": {
+      "entryChecksum": "e93fa854cd329fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b857",
       "lastSyncTime": "2026-05-26T06:18:49Z",
       "aspects": {
-        "dataplex-types.global.overview": {
-          "checksum": "a8f5c1e39834c2219efbf4c8996fb92427ae41e4649b934ca495991b7852b855",
-          "format": "MARKDOWN",
-          "file": "orders.overview.md"
-        },
-        "dataplex-types.global.descriptions": {
-          "checksum": "5b30c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b856",
-          "format": "YAML",
-          "file": "orders.yaml"
-        }
+        "dataplex-types.global.overview": "a8f5c1e39834c2219efbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        "dataplex-types.global.descriptions": "5b30c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b856"
       }
     }
   }
@@ -116,10 +128,12 @@ A single JSON file at the root of the workspace containing a map of all tracked 
 
 * **Pros:**
   * **High Performance:** Native V8 JSON parsing (`JSON.parse`/`JSON.stringify`) is extremely fast and incurs zero external parsing dependency overhead.
+  * **Startup Optimization / Mapping:** The state file maps entry IDs directly to their base hashes and structures. This acts as an in-memory catalog index, allowing the startup phase to construct a map of the local catalog's content without recursively enumerating, scanning, and parsing every file in the workspace directory.
+  * **Derivable File Paths:** File locations and formats are derived from entry names based on the `CatalogSource` and `CatalogLayout` implementations, allowing the state file to omit `format` and `file` fields for a highly compact representation.
   * **Zero Clutter:** Keeps the user's active catalog folders completely clean of system-managed state files.
-  * **Easy Git Ignoring:** The single `.catalog_state.json` file is easily ignored globally via `.gitignore`.
 * **Cons:**
-  * **Parallelization Bottleneck (Lock Contention):** Operating on a single centralized file requires strict concurrent process locking to prevent state corruption. During large/bulk export or import operations, this single-file lock becomes a contention point that limits parallel execution capabilities across processes.
+  * **State Desynchronization & Atomicity Risk:** Storing authorable metadata in the `catalog/` directory and its base sync state in `catalog-state.json` separates the data from its tracking ledger. Because writing to two separate files cannot be executed as a single atomic OS transaction, any process interruption (such as a crash, manual cancel, or disk-full error) in between operations can leave the workspace in an inconsistent state. This requires the CLI tool to implement strict write-ordering sequences or write-ahead journaling to detect and self-heal inconsistencies.
+  * **Global Write Amplification:** Every single metadata modification, creation, or deletion—even to a single aspect or description—requires the CLI to read, deserialize, modify, reserialize, and rewrite the *entire* centralized `catalog-state.json` database. At a large catalog scale, this introduces massive I/O overhead for minor incremental changes and creates a global execution bottleneck if multiple automated runs attempt concurrent updates.
 
 ---
 
@@ -172,7 +186,7 @@ path/to/root/
 
 ### Concurrent Process Locking
 
-To prevent state database corruption when multiple instances of the `kcmd` CLI or automated runners (such as CI/CD pipelines or local agent tools) run concurrently, the tool implements a lightweight, cross-platform file lock on `.catalog_state.json` during read/write sequences. If the state file is locked, concurrent operations will fail-fast with a concurrency conflict error.
+To prevent state database corruption when multiple instances of the `kcmd` CLI or automated runners (such as CI/CD pipelines or local agent tools) run concurrently, the tool implements a lightweight, cross-platform file lock on `catalog-state.json` during read/write sequences. If the state file is locked, concurrent operations will fail-fast with a concurrency conflict error.
 
 > [!IMPORTANT]
 > **Aspect Creations & Type Registration Constraints**
@@ -191,14 +205,21 @@ sequenceDiagram
     autonumber
     participant Tool as CLI Sync Tool
     participant Disk as Local Workspace
-    participant State as State File (.catalog_state.json)
+    participant State as State File (catalog-state.json)
     participant Remote as Knowledge Catalog
 
     Note over Tool: User executes "pull"
     Tool->>Disk: Read Local Workspace Files
     Tool->>State: Load Base Checksums
-    Tool->>Tool: Calculate Current Local Checksums
-    Tool->>Tool: Compare with Base Checksums
+    Tool->>Tool: Calculate Current Local Entry Checksums
+    
+    rect rgb(45, 55, 72)
+    Note over Tool: Optimization Layer: Check Entry Checksum First
+    Tool->>Tool: Compare Entry Checksum with Base
+    alt Entry Checksum Changed
+        Tool->>Tool: Calculate and Compare Granular Aspect Checksums
+    end
+    end
     
     alt Local Modifications Detected & No --force Flag
         Note over Tool: DIRTY WORKSPACE! Abort pull.
@@ -209,16 +230,17 @@ sequenceDiagram
 
         loop For Each Entry
             Tool->>Tool: Normalize (Semantic & whitespace normalization)
-            Tool->>Tool: Calculate Aspect Checksums
-            Tool->>Disk: Write Sidecar Aspect Files (YAML/Markdown)
-            Tool->>State: Write Aspect Base Checksums
+            Tool->>Tool: Calculate Entry & Aspect Checksums
+            Tool->>Disk: Write Sidecar Aspect/Core Files (YAML/Markdown)
+            Tool->>State: Write Entry & Aspect Base Checksums
         end
     end
 ```
 
 1. **Workspace Dirty Check (Pull-Safety Guard - Strict All-or-Nothing):**
-   - Scan all local aspect files and compute their current local checksums.
-   - Compare with the `Base Checksum` in the state file.
+   - Compute current local entry-level checksums for all catalog entries.
+   - Compare each computed checksum against the `entryChecksum` in the state file.
+   - For any entry where the entry-level checksum differs, perform granular checks on its individual local aspects and core fields to confirm actual modifications.
    - If *any* local change is detected anywhere in the workspace and the `--force` flag is absent, the entire pull operation aborts immediately. This strict all-or-nothing constraint protects unpushed work across the entire workspace.
    
    > [!NOTE]
@@ -238,7 +260,7 @@ sequenceDiagram
 
 4. **Save State:**
    - Save the normalized aspect content to disk (either in aspect YAML or sidecar `.md` files).
-   - Update the computed aspect checksums as the new **Base Checksums** in `.catalog_state.json`.
+   - Update the computed entry-level and aspect checksums as the new **Base Checksums** in `catalog-state.json`.
 
 ---
 
@@ -249,13 +271,21 @@ sequenceDiagram
     autonumber
     participant Tool as CLI Sync Tool
     participant Disk as Local Workspace
-    participant State as State File (.catalog_state.json)
+    participant State as State File (catalog-state.json)
     participant Remote as Knowledge Catalog
 
     Note over Tool: User executes "push"
     Tool->>Disk: Read Local Workspace Files
     Tool->>State: Load Base Checksums
-    Tool->>Tool: Identify Locally Modified/Created/Deleted Aspects
+    
+    rect rgb(45, 55, 72)
+    Note over Tool: Optimization Layer: Check Entry Checksum First
+    Tool->>Tool: Calculate Current Local Entry Checksums
+    Tool->>Tool: Compare with Base Entry Checksums
+    alt Entry Checksum Changed
+        Tool->>Tool: Identify Granular Locally Modified/Created/Deleted Aspects & Core Fields
+    end
+    end
     
     Note over Tool: Fetch Remote State: Individual Lookup
     loop For Each Entry with Modifications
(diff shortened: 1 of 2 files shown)


toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
+# Change Detection & Conflict Resolution Design Spec
+
+This document outlines the design for local change detection and remote conflict resolution in the **Metadata as Code** tool.
```
sapank-sys: nit: library not tool

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Design Goals
+* **Local Change Detection**: Swiftly isolates fields or aspects modified locally since the last `pull` or `push`.
+* **Minimizing Network Footprint**: Only transmits aspects that have changed locally, omitting unmodified metadata.
+* **Zero-Workspace Clutter**: System revisions, sync timestamps, and base checksums are stored entirely in the state database, keeping authorable source YAML/Markdown files clean for Git history.
+
+---
+
+## Comprehensive Conflict Resolution Matrix
+
```
sapank-sys: Can we add a short summary on what is a conflict, and choices for a user in case of a conflict.

Conflict : Local filesystem is not in sync with "Catalog's" state of the resource. 

Choices : User manually merges latest pull information with desired changes OR Abandons local changes OR forces an update as per local state to Remote.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+The conflict resolution matrix covers aspect modifications, creations, and deletions:
+
+| Local Aspect State | Remote Aspect State | Outcome / Action | Description |
```
sapank-sys: For deleted there are 2 scenarios : 

what happens when local entry has changes but remote entry itself is deleted.  
what happens when local entry has changes but only modified aspect in remote entry is deleted.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+| **Modified** | **Deleted** | **Conflict** | Conflict! Remote was deleted while local was modified. |
+| **Deleted** | **Unchanged** | **Safe to Delete Remotely** | Push aspect deletion to remote catalog. Clean up local state entry. |
+| **Deleted** | **Modified** | **Conflict** | Conflict! Remote was modified while local was deleted. |
+| **Deleted** | **Deleted** | **Skip** | Aspect was deleted on both sides. Clean up local state. |
+| **Created (New)** | **Exists Remotely** | **Conflict** | Conflict! Local aspect created but already exists on remote. |
+| **Created (New)** | **Does Not Exist** | **Safe to Create** | Safe to push/create on remote. Base checksum is recorded. |
+
+---
+
+## The State Database (`.catalog_state.json`)
```
sapank-sys: this feels like an abrupt jump from the state to the implementation . 
   Can we describe , how we get here in a brief explanation of the design.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+          "format": "YAML",
+          "file": "orders.yaml"
+        }
+      }
+    }
+  }
+}
+```
+
+#### Option B: Mirror State Directory (`.catalog_state/`)
```
sapank-sys: since options B and C are not preferred, do we just move them below to a alternatives considered section,

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,290 @@
...
+| **Deleted** | **Modified** | **Conflict** | Conflict! Remote was modified while local was deleted. |
+| **Deleted** | **Deleted** | **Skip** | Aspect was deleted on both sides. Clean up local state. |
+| **Created (New)** | **Exists Remotely** | **Conflict** | Conflict! Local aspect created but already exists on remote. |
+| **Created (New)** | **Does Not Exist** | **Safe to Create** | Safe to push/create on remote. Base checksum is recorded. |
+
+---
+
+## The State Database (`.catalog_state.json`)
+
+A flat JSON database maps each tracked entry and its corresponding aspects to their last synchronized state.
```
sapank-sys, line 186: Will using a single file for a large / bulk export import become a hindrance in parallelization ?

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+* **Option C (Co-located Checksum Files):**
+  * *Pros:* Highly localized and self-contained. If a directory or file is renamed/deleted by the user, clean-up is intuitive and localized.
+  * *Cons:* Clutters the user workspace by doubling the number of files in the catalog folders, hindering human legibility.
+
+**Recommendation:** We proceed with **Option A (Flat JSON State File)** to optimize read/write speeds in Node.js and prioritize a clean workspace.
+
+
+### Concurrent Process Locking
+
+To prevent state database corruption when multiple instances of the `kcmd` CLI or automated runners (such as CI/CD pipelines or local agent tools) run concurrently, the tool implements a lightweight, cross-platform file lock on `.catalog_state.json` during read/write sequences. If the state file is locked, concurrent operations will fail-fast with a concurrency conflict error.
```
sapank-sys: See comment above on concurrency bottlenecks for bulk push/pull

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,290 @@
...
+   * **Aspect Modifications**: Aspects to be updated.
+   * **Aspect Creations**: New aspects to be added.
+   * **Aspect Deletions**: Aspects to be removed.
+5. **Execution Safeguard**: **No write operations** are performed on the remote Dataplex service (`modifyEntry` is not called) and `.catalog_state.json` remains unchanged.
+
+---
+
+## Future Scope
+
+### 1. State File Resiliency & Atomic Writes
```
sapank-sys: Add ability for automatically **merging non-overlapping** changes within a file as a Future scope item

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Done

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Added in the table

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: added a section 'Design Rationale: Why a State Database is Required'

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Yes, it might. Added this in con. As well as added in future scope to consider addressing this

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Added

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Done. Thanks

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Done

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+### Design Goals
+* **Local Change Detection**: Swiftly isolates fields or aspects modified locally since the last `pull` or `push`.
+* **Minimizing Network Footprint**: Only transmits aspects that have changed locally, omitting unmodified metadata.
+* **Zero-Workspace Clutter**: System revisions, sync timestamps, and base checksums are stored entirely in the state database, keeping authorable source YAML/Markdown files clean for Git history.
+
+---
+
+## Comprehensive Conflict Resolution Matrix
+
+A **conflict** occurs when the local filesystem is not in sync with the Catalog's state of the resource.
```
nikhilk: Its a bit more restrictive. Its when there are changes made to both local and service copies of the metadata.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+
+In the event of a conflict, the user has three choices:
+* **Manual Merge (via Force Pull)**: Back up local modifications, run a force pull (`kcmd pull --force`) to fetch remote changes, and manually merge the desired local changes.
+* **Abandon Local Changes (via Force Pull)**: Run a force pull (`kcmd pull --force`) to discard local modifications and align completely with the remote Catalog state.
+* **Force Update (via Force Push)**: Run a force push (`kcmd push --force`) to overwrite the remote Catalog state with the local workspace state.
+
+The conflict resolution matrix covers aspect modifications, creations, and deletions:
```
nikhilk: Can you split the table into behaviors that impact pull and separately behaviors that impact push.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+
+    Local ===|1. Detect Local Edits| Base
+    Base ===|2. Detect Remote Edits| Remote
+    
+    style Base fill:#2d3748,stroke:#3182ce,stroke-width:2px,color:#fff
+```
+
+---
+
+## The State Database (`.catalog_state.json`)
```
nikhilk: The hidden file has appeal. However, the risk is its easy to miss if the user copies files around on local machine, or forgets to add them to their git repo.

Alternative, similar to package-lock.json alongside package.json, a catalog-state.yaml alongside catalog.yaml would be interesting.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+
+A flat JSON database maps each tracked entry and its corresponding aspects to their last synchronized state.
+
+### Checksum Scope & Modifiability Constraints
+
+Before discussing database layout options, we define the scope and constraints of the tracked metadata:
+* **Aspect-Level Only**: This tool does not support entry-level metadata modifications (e.g., `displayName`, `description` defined directly on the Entry resource). Therefore, no entry-level checksums are calculated or tracked.
```
nikhilk: This is needed for entrygroups that are not ingested, as entry.EntrySource is modifiable, and because entries are creatable/deletable.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+* **Aspect `checksum`**: These checksums are calculated for the key-value properties inside each individual aspect type (e.g., `dataplex-types.global.overview`, `dataplex-types.global.descriptions`).
+* **Static Modifiability Enforcement**: The `catalog.yaml` validation layer statically ensures that only modifiable aspects (e.g., non-required aspects for ingested entries, or any aspect for custom/user-managed entries) are configured to be synchronized and published.
+
+### Proposed Design: Flat State File (`.catalog_state.json`)
```
nikhilk: We should make sure this file can also be used to build a map of the contents of the local catalog, so the catalog snapshot does not have to recursively enumerate all files at startup and parse them enough to populate the in-memory index on startup.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+
+A single JSON file at the root of the workspace containing a map of all tracked files, their base aspects, and their checksums.
+
+> [!NOTE]
+> **Why JSON over YAML for system state?**
+> Since `.catalog_state.json` is entirely system-managed and gitignored, human readability is secondary to execution performance and parsing reliability. Node.js has native, highly-optimized C++ support for parsing and serializing JSON (`JSON.parse` / `JSON.stringify`). Utilizing YAML would require external JavaScript-based parser libraries, which introduce significant CPU and execution latency during CLI startup as the catalog scales.
```
nikhilk: Curious at what scale this matters and is true.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+{
+  "version": "1",
+  "entries": {
+    "ecommerce-prod.ecommerce-dataset/orders": {
+      "lastSyncTime": "2026-05-26T06:18:49Z",
+      "aspects": {
+        "dataplex-types.global.overview": {
+          "checksum": "a8f5c1e39834c2219efbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+          "format": "MARKDOWN",
+          "file": "orders.overview.md"
```
nikhilk: Is there a reason file and format are needed? Those should be derivable from the name of the entry based on the CatalogSource and CatalogLayout implementations.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+      }
+    }
+  }
+}
+```
+
+* **Pros:**
+  * **High Performance:** Native V8 JSON parsing (`JSON.parse`/`JSON.stringify`) is extremely fast and incurs zero external parsing dependency overhead.
+  * **Zero Clutter:** Keeps the user's active catalog folders completely clean of system-managed state files.
+  * **Easy Git Ignoring:** The single `.catalog_state.json` file is easily ignored globally via `.gitignore`.
```
nikhilk: It should not be. Like package-lock.json, this should be committed into the repo.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+* **Cons:**
+  * **Parallelization Bottleneck (Lock Contention):** Operating on a single centralized file requires strict concurrent process locking to prevent state corruption. During large/bulk export or import operations, this single-file lock becomes a contention point that limits parallel execution capabilities across processes.
```
nikhilk: I don't think this applies. The CatalogSnapshot should be thread-local.
What does matter is every create/update/delete/push/pull operation will need to update this file.

The other tricky thing (vs. state in the entry file itself) is being able to update both metadata and its checksum atomically.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+   * **Aspect Deletions**: Aspects to be removed.
+5. **Execution Safeguard**: **No write operations** are performed on the remote Dataplex service (`modifyEntry` is not called) and `.catalog_state.json` remains unchanged.
+
+---
+
+## Future Scope
+
+### 1. State File Resiliency & Atomic Writes
+To prevent state file corruption when writing updates to `.catalog_state.json` (e.g., if the CLI process is interrupted, crashes, or the disk runs out of space):
+* **Atomic Write Mechanism**: Future versions of the CLI should implement atomic file updates. Instead of writing directly to the target `.catalog_state.json` file, the tool will write the updated state to a temporary file (e.g., `.catalog_state.json.tmp`) in the same directory, and then use an atomic rename operation (`fs.rename` or equivalent OS-level rename) to replace the original file.
```
nikhilk: I would expect in-memory state -> persisted state is essentially given you that.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+* **Automatic Backup**: Maintain a `.catalog_state.json.bak` copy before executing any state transitions to facilitate self-healing and recovery options for corrupted states.
+
+### 2. State Scalability & Parallelization
+To resolve the lock contention bottleneck of a single centralized state file during large-scale or bulk import/export operations:
+* **Transactional State Indexer**: Explore transitioning the state storage to a transactional database. Utilizing a lightweight embedded engine like SQLite is one potential option to enable row-level concurrency control rather than file-level locking; however, the final choice of approach or storage engine should be thoroughly evaluated and decided based on performance and architectural requirements at the time of implementation.
```
nikhilk: Not sure how this helps, unless the entry file is also managed in the DB. The inconsistency is a result of separation of state from the entry. And moving it into the DB means its still separate.

toolbox/mdcode/docs/checksum_design.md
```diff
@@ -0,0 +1,349 @@
...
+
+### 3. Fine-Grained Change Isolation & Automatic Merging
```
nikhilk: I think this is needed sooner to minimize the conflicts.

nikhilk (changes requested):

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack. Modified

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Done.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack. Changed the file to catalog-state.json.

Replied to other comment related to yaml vs json.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack. Modified to have a entry level checksum as well.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack, the file can be used for startup optimization. Added this as a Pro for the proposed design as well.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Tried to verify this using gemini, following are the results:

For a catalog containing 10,000 entries (resulting in a file size of approximately 4.8 MB), running a local benchmark in Node.js yields the following results:


Operation | JSON (JSON.parse / JSON.stringify) | YAML (yaml npm library version 2.8.4) | Performance Difference
-- | -- | -- | --
Parsing (Read) | 12.4 ms | 3767 ms (3.76 seconds) | JSON is ~303x faster
Serialization (Write) | 18.9 ms | 428.4 ms | JSON is ~22x faster

Scale Impact
- At small scale (100 entries, ~50 KB): The difference is negligible in absolute time (less than 5ms).
- At medium scale (1,000 entries, ~500 KB): YAML parsing takes ~350ms, introducing some lag.
- At large scale (10,000 entries, ~4.8 MB): YAML parsing takes nearly 4 seconds, which is quite some lag. Using JSON keeps the startup parsing overhead completely imperceptible (12ms).

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack, removed the formats and file names from state file.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack. Removed this Pro from the list.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Yes, makes sense. Thanks.

Updated the con to 'State Desynchronization & Atomicity Risk' and 'Global Write Amplification'

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Not sure if the in-memory operations prevents the filesystem-level corruption.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Agree. This was originally added because of the 'Parallelization Bottleneck' issue which isn't applicable (discussed in another comment).

Removing this section.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Agree, moving this as first point. We need to think about this soon.

toolbox/mdcode/docs/checksum_design.md
nikhilk: Could you do a quick test using Bun's built-in YAML parser as well? I haven't switched to using it yet, but I expect it to be faster than the current library.

https://bun.com/docs/runtime/yaml

toolbox/mdcode/docs/checksum_design.md
nikhilk: I would expect that when node process writes to the file, if there is a failure mid-way, and hence file handle is not flushed/closed, there will be no partial save. Otherwise, no file writing code would work without having to implement this write-to-alternate-file-and-then-move-or-rename strategy.

nikhilk:
Nit -- could you please move this to docs/features/checksum.md -- to keep the top-level directory somewhat clean.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: The conclusion is: Bun's yaml parsers is significantly better than NPM's yaml. Overall JSON is better but we can use YAML with Bun's yaml parser.

Here is the similar comparison:

For a catalog containing **10,000 entries**, running a local benchmark in Bun yields the following absolute results:

### Performance Comparison Table

| Operation | JSON (Native `JSON.parse/stringify`) | Bun.YAML (Native C Parser) | YAML (NPM `yaml` library) | Performance Comparison |
| :--- | :--- | :--- | :--- | :--- |
| **Parsing (Read)** | **8.53 ms** | **79.46 ms** | **1,995.30 ms** (1.99s) | JSON is **9.3x** faster than `Bun.YAML`, and **233x** faster than NPM `yaml` |
| **Serialization (Write)** | **3.53 ms** | **44.23 ms** | **439.48 ms** | JSON is **12.5x** faster than `Bun.YAML`, and **124x** faster than NPM `yaml` |

---

### Scale Impact

*   **At small scale (100 entries, ~50 KB):** The difference is negligible in absolute time (less than 2ms across all options).
*   **At medium scale (1,000 entries, ~500 KB):** 
    *   `JSON` takes **< 1 ms**.
    *   `Bun.YAML` takes **~8 ms** (completely imperceptible).
    *   `NPM yaml` takes **~200 ms**, introducing noticeable lag.
*   **At large scale (10,000 entries, ~3.6 MB):** 
    *   `JSON` takes **8.5 ms** (completely imperceptible).
    *   `Bun.YAML` takes **79.4 ms** (extremely fast, virtually imperceptible).
    *   `NPM yaml` takes nearly **2.0 seconds**, creating a highly frustrating startup delay.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Ack. Removing the future scope for now. We can consider it or any other scope later if required.

toolbox/mdcode/docs/checksum_design.md
sumedhdk22: Addressed this in the latest commit. Removing this from future scope since it is now part of the design.

sumedhdk22:
@nikhilk , I have updated the design with fine-grained change isolation. PTAL.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+### Design Goals
+* **Local Change Detection**: Swiftly isolates fields or aspects modified locally since the last `pull` or `push`.
+* **Minimizing Network Footprint**: Only transmits aspects that have changed locally, omitting unmodified metadata.
+* **Zero-Workspace Clutter**: System revisions, sync timestamps, and base checksums are stored entirely in the state database, keeping authorable source YAML/Markdown files clean for Git history.
+
+---
+
+## Logical Entry Composition & Sidecar Mapping
+
+To provide a clean, readable layout for unstructured text (such as rich-text documentation or Markdown profiles), the synchronization engine operates on a logical **Entry** boundary rather than treating filesystem files in isolation.
```
sapank-sys: The engine terminology seems new, Can we call it tool or library to remain consistent with other docs.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+    classDef activeNode fill:#e1f5fe,stroke:#0288d1,stroke-width:2px,color:#01579b;
+    class Main,SidecarMD,Normalize,Checksum,StateDB lightNode;
+    class Compose activeNode;
+```
+
+---
+
+## Conflict Resolution
+
+A **conflict** occurs when both the local workspace copy and the remote catalog copy of the same aspect or core metadata have been modified since the last synchronization.
```
sapank-sys: Can we add definitions of synchronization. 
IIUC : 
A pull synchronizes local with remote,
A push synchronizes remote with local.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+## Conflict Resolution
+
+A **conflict** occurs when both the local workspace copy and the remote catalog copy of the same aspect or core metadata have been modified since the last synchronization. 
+
+To handle these scenarios safely and flexibly, the sync engine provides fine-grained CLI configurations and distinct workflow strategies.
+
+### 1. CLI Synchronization Configurations
+The synchronization engine supports three core flags to customize conflict handling in both interactive (local developer) and automated (CI/CD) environments:
+
+*   **`--allow-partial` (Boolean, default: `false`):**
```
sapank-sys: does this flag only apply to pull, or also to push?

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+
+### 1. CLI Synchronization Configurations
+The synchronization engine supports three core flags to customize conflict handling in both interactive (local developer) and automated (CI/CD) environments:
+
+*   **`--allow-partial` (Boolean, default: `false`):** 
+    *   When `false` (strict mode): Any unresolved conflict aborts the entire operation immediately (non-zero exit code).
+    *   When `true` (partial mode): The engine pushes/pulls all non-conflicting aspects, but skips and warns about the conflicting aspects. The CLI exits with code `0` (success) but logs a warning summary.
+*   **`--conflict-resolution=<strategy>` (Enum, default: `manual`):** Specifies the programmatic resolution policy:
+    *   `manual`: Conflicts are flagged for human intervention (skipped when `--allow-partial` is active; otherwise aborts).
+    *   `accept-remote` (Remote Wins): Automatically resolves conflicts by overwriting the local workspace with the remote catalog state.
```
sapank-sys: A push with accept-remote, will ignore the local changes . in that case the push will still return back success, or will the user get any indication of the adopted resolution?

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+To handle these scenarios safely and flexibly, the sync engine provides fine-grained CLI configurations and distinct workflow strategies.
+
+### 1. CLI Synchronization Configurations
+The synchronization engine supports three core flags to customize conflict handling in both interactive (local developer) and automated (CI/CD) environments:
+
+*   **`--allow-partial` (Boolean, default: `false`):** 
+    *   When `false` (strict mode): Any unresolved conflict aborts the entire operation immediately (non-zero exit code).
+    *   When `true` (partial mode): The engine pushes/pulls all non-conflicting aspects, but skips and warns about the conflicting aspects. The CLI exits with code `0` (success) but logs a warning summary.
+*   **`--conflict-resolution=<strategy>` (Enum, default: `manual`):** Specifies the programmatic resolution policy:
+    *   `manual`: Conflicts are flagged for human intervention (skipped when `--allow-partial` is active; otherwise aborts).
```
sapank-sys: why is manual skipped when allow-partial is active ?

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+    *   `kcmd pull --force` $\equiv$ `kcmd pull --conflict-resolution=accept-remote`
+
+---
+
+### 2. Workflow Resolution Strategies
+
+Depending on the synchronization direction, developers can apply these CLI flags to execute the following workflow choices:
+
+#### Pull Conflict Strategies
+*   **Manual Resolution (Default):** Back up your local edits, run `kcmd pull --force` (which accepts the remote state), and then manually merge your backed-up edits back into the workspace.
```
sapank-sys: Lets remove the manual . Instead call out the use of allow-partial, similar to Push

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+    *   `kcmd push --force` $\equiv$ `kcmd push --conflict-resolution=accept-local`
+    *   `kcmd pull --force` $\equiv$ `kcmd pull --conflict-resolution=accept-remote`
+
+---
+
+### 2. Workflow Resolution Strategies
+
+Depending on the synchronization direction, developers can apply these CLI flags to execute the following workflow choices:
+
+#### Pull Conflict Strategies
```
sapank-sys: Conflict Resolution Strategies

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+*   **Manual Resolution (Default):** Back up your local edits, run `kcmd pull --force` (which accepts the remote state), and then manually merge your backed-up edits back into the workspace.
+*   **Abandon Local Changes:** Run `kcmd pull --force` to cleanly discard all local modifications and completely align the workspace with the remote catalog.
+
+#### Push Conflict Strategies
+*   **Manual Resolution (Default):** Abort the push, pull the remote catalog changes, resolve the conflict locally in your editor, and push again.
```
sapank-sys: can be removed.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+
+> [!NOTE]
+> **Partial Pull Behavior (`--allow-partial`):**
+> If `--allow-partial` is set during a standard pull, any aspect triggering an **Abort (Dirty Workspace)** will be cleanly skipped with a warning, allowing all other unmodified/safe aspects to be pulled successfully.
+
+| Local Aspect State | Remote Aspect State | Standard Pull (`kcmd pull`) | Force Pull (`kcmd pull --force`) | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| **Unchanged** | **Unchanged** | **Skip** | **Skip** | No changes on either side. |
+| **Unchanged** | **Modified** | **Safe to Pull** | **Safe to Pull** | Remote modifications are safely written to the local workspace. |
+| **Unchanged** | **Deleted** | **Safe to Delete Locally** | **Safe to Delete Locally** | Remote deletion is propagated locally; the local aspect file is deleted. |
```
sapank-sys: Remote aspect deletions are always propagated locally (even when allow-partial is true).

sapank-sys: on similar lines, do we want to add a row for remote entry deletion as well.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+| **Does Not Exist** | **Exists Remotely** | **Safe to Pull** | **Safe to Pull** | The remote aspect is pulled down and created in the local workspace. |
+| **Modified** / **Created** / **Deleted** | **Any State** | **Abort (Dirty Workspace)** | **Overwritten by Remote** | Local changes exist in the workspace. Standard pull aborts immediately to prevent overwriting unpushed local work. Force pull discards local changes and overwrites them with the remote state. |
+
+### 2. Push Conflict Resolution Matrix
+
+During a `push` operation, local modifications are sent to the remote catalog. A standard `push` will abort if any conflict is detected (i.e., if the remote state has changed since the last synchronization).
+
+> [!NOTE]
+> **Partial Push Behavior (`--allow-partial`):**
```
sapank-sys, line 108: Since the remote catalog doesn't have etag support yet,  there is a possibility of even a deemed safe change actually overriding latest catalog entry.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+> If `--allow-partial` is set during a standard push, any aspect triggering a **Conflict (Abort)** will be skipped and logged as a warning, allowing the rest of the non-conflicting aspects to be successfully published.
+
+| Local Aspect State | Remote Aspect State | Standard Push (`kcmd push`) | Force Push (`kcmd push --force`) | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| **Unchanged** | **Unchanged** | **Skip** | **Skip** | No changes on either side. |
+| **Unchanged** | **Modified** | **Skip** | **Skip** | No local changes to push. Remote changes remain until the next pull. |
+| **Modified** | **Unchanged** | **Safe to Push** | **Safe to Push** | Pushes local modifications to the remote catalog. Base checksum is updated. |
+| **Modified** | **Modified** | **Conflict (Abort)** | **Force Overwrite** | Both copies were modified. Standard push aborts. Force push overwrites the remote aspect. |
+| **Modified** | **Deleted (Entry Deleted)** | **Conflict (Abort)** | **Force Create/Overwrite** | The remote entry itself was deleted. Force push recreates the entry and aspect on the remote. |
```
sapank-sys: Recreate will only be allowed for entries that are not directly managed by GCP (1p)

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+---
+
+## Deletion Mechanics & Lifecycle
+
+Deletion represents a critical state transition in metadata synchronization. The engine enforces strict rules to distinguish between deleting a specific aspect and deleting an entire catalog entry.
+
+### 1. Entry-Level Deletion
+The presence of the **Core Entry File** (e.g., `orders.yaml`) represents the existence of the Entry.
+*   **Local Entry Deletion:** If a developer deletes `orders.yaml` from their workspace, the sync engine interprets this as an intentional request to delete the entire entry and all its aspects. During `push`, the entry is removed from the remote catalog.
+*   **Remote Entry Deletion:** If an entry is deleted remotely, a `pull` operation will remove `orders.yaml` and all its sidecars from the local workspace.
```
sapank-sys: if there is no conflict.  In case of conflict, see table .

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+
+### 2. Aspect-Level Deletion (Sidecar Removal)
```
sapank-sys: this section should be for both sidecar and inline aspects

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+## The State Database (`catalog-state.json`)
+
+A flat JSON database maps each tracked entry and its corresponding aspects to their last synchronized state.
+
+### Checksum Scope & Modifiability Constraints
+
+Before discussing database layout options, we define the scope and constraints of the tracked metadata:
+* **Entry-Level & Aspect-Level Checksums**: The tool tracks both a unified **Entry-Level Checksum** and individual **Aspect-Level Checksums**. This is critical for user-managed or custom entries where the entry itself can be created or deleted, and for uningested entry groups where entry-level metadata (such as `entrySource`, `displayName`, or `description`) is modifiable.
+* **Aspect `checksum`**: These checksums are calculated for the key-value properties inside each individual aspect type (e.g., `dataplex-types.global.overview`, `dataplex-types.global.descriptions`).
+* **Static Modifiability Enforcement**: The `catalog.yaml` validation layer statically ensures that only modifiable aspects and entry properties are configured to be synchronized and published.
```
sapank-sys: the validation should refer to the publishable aspects/properties.  User is allowed to synchronize/pull non-modifiable properties locally.

toolbox/mdcode/docs/features/checksum.md
```diff
@@ -0,0 +1,454 @@
...
+   - **Concurrency Control:** Calls are executed using a throttled concurrent request pool to prevent socket exhaustion and service-side rate limiting (HTTP 429).
+
+3. **Verify & Push:**
+   - The conflict resolution behavior is fully customized by the CLI configuration:
+     1. **Force Overwrite:** If `--force` (or `--conflict-resolution=accept-local`) is specified, all conflict checks are bypassed. Local states overwrite the remote catalog.
+     2. **Default Strict (allow-partial = false):** If any aspect or core metadata fails conflict verification (i.e., both local and remote have modified it since the base sync), the entire push operation is aborted immediately. No mutations are committed, and the CLI exits with code `1`.
+     3. **Partial Publish (allow-partial = true):** The engine relaxes the strict check. Conflicting aspects are skipped, leaving their local files dirty and their base state unchanged in `catalog-state.json`. All non-conflicting edits are committed, and the CLI exits with code `0` with warning logs listing the skipped aspects.
```
sapank-sys: base state checksums in partial push should be preferably be updated with the final remote state to allow future push to detect pending modifications.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Ack. Calling the operation as 'synchronization' and the library at places where the design is talking about tool (since there will be MCP tools in the future for this as well.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Done. Added.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Added clarification that it is applicable for both the operations.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Removed the combined flags behavior. For manual, Conflicts are flagged for human intervention.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Added clarification, that 'accept-remote' is for pull and 'accept-local' is for push operation

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Modified.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Done

toolbox/mdcode/docs/features/checksum.md
sumedhdk22:
```
Remote aspect deletions are always propagated locally (even when allow-partial is true).
```
Yes. We are not bringing --allow-partial flag behavior in the table. It will always come into picture when there are conflicts and there is flag set.
In this scenario specifically, it is clean delete since local is unchanged and remote is delete.


```
on similar lines, do we want to add a row for remote entry deletion as well.
```
Done, added.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Makes sense. Added warning.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Yes, added in the description section of this.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Updated

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Yes, updated the section

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Yes, updated the text.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Yes, updated the doc accordingly.

toolbox/mdcode/docs/features/checksum.md
sumedhdk22: Done

PR #12, opened 2026-06-03
Add EntryLinks support

eeshadutta1408:
This PR implements full support for Dataplex **EntryLinks** (both top-level and schema column-inlined links) in the `kcmd` tool. It enables catalog curators to manage relationships (such as definition/glossary linkages or schema joins) locally in a structured format and synchronize them with the cloud catalog.

---

### Key Features
* **Manifest & Alias Mapping:** Added `entryLinks` configuration blocks to snapshot/publishing schemas in `manifest.ts`, supporting custom and global/system aliases (e.g. `definition`, `schema-join`).
* **Dataplex API Integration:** Extended `dataplex.ts` with Glossary and EntryLinks endpoints (`lookupEntryLinks`, `createEntryLink`, `deleteEntryLink`, `listEntryLinks`), with built-in CRM project ID/number normalization.
* **Bi-directional Synchronization:** Implemented paginated, cached pulling for entry group scopes and **Symmetrical Reference Matching** on pushing in `sync.ts` to prevent accidental link deletion or duplicate creation.

---

### EntryLink Examples

Below are examples of how entry links are serialized locally in an entry's YAML file:

#### 1. Schema-Join Link
For undirected relationships containing aspect metadata, such as join keys and inference attributes:
```yaml
links:
  dataplex-types.global.schema-join:
    - target: myzh-test-project-1.froyotech_dataset.component
      id: bigquery.googleapis.com/projects/myzh-test-project-1/datasets/froyotech_dataset/tables/component
      schema-join:
        joins:
          - source:
               name: product
               fields:
                 - name
            target:
               name: component
               fields:
                 - name
            type: JOIN
            inferenceSource: QUERY_HISTORY
        userManaged: true
        job:
          runTime: 2026-02-27T13:24:32Z
          name: Dataplex Catalog Usage Signals
```

#### 2. Glossary definition Link
For directional references mapped to glossary term path targets (resolved dynamically using display names):
```yaml
links:
  dataplex-types.global.definition:
    - target: dc-cuj-staging-playground.global.UtilityTest.UtilityTerm1
      id: projects/dc-cuj-staging-playground/locations/global/glossaries/pephjajda-c7c4b951-9c078501-4f35fa82/terms/imbaoooej-b14cc513-abdaf298-1b8843c2
```


toolbox/mdcode/src/libts/gcp/dataplex.ts
```diff
@@ -177,12 +205,115 @@ export class CatalogClient extends api.ApiClient {
     } while (pageToken);
   }
 
+  async lookupEntryLinks(
```
nikhilk, line 208: Does this need to page through the list of entry links to retrieve them all similar to list?

toolbox/mdcode/src/libts/gcp/dataplex.ts
```diff
@@ -177,12 +205,115 @@ export class CatalogClient extends api.ApiClient {
...
+    }
+    link.aspects = fixedAspects;
+  }
 }
 
 
 // Fix all entries and aspects to consistently use project id. Its currently a mess with an
 // inconsistent mix of project ids and unusable project numbers.
-async function _fixEntry(entry: Entry, ctx: context.ApiContext): Promise<void> {
+export async function _fixEntry(entry: Entry, ctx: context.ApiContext): Promise<void> {
```
nikhilk: This function is expected to be used only within the service wrapper, the idea being all responses returned from this client layer are 'fixed', and any downstream user of this client should not have to think about such details.

toolbox/mdcode/src/libts/manifest.ts
```diff
@@ -8,26 +8,139 @@ import * as gcp from './gcp';
 import { CatalogSource, createSource, Sources } from './source';
 
 
+export const SYSTEM_LINK_ALIASES: Record<string, string> = {
```
nikhilk: Couple of things -

1. Alias functionality would probably be better added in future alongside support for entry type and aspect type aliases. Being worked on in https://github.com/GoogleCloudPlatform/knowledge-catalog/pull/20/changes ... 
2. Still looking ahead, but we should support all link types without having to list them all here. The set of link types to be retrieved or created/modified should depend on the types listed in catalog manifest snapshot and publishing configuration.

toolbox/mdcode/src/libts/gcp/dataplex.ts
eeshadutta1408: Yes, we should. Updated `lookupEntryLinks` to use a `do-while` loop that pages through all available entry links using `nextPageToke`n (similar to the logic in `listEntryLinks`).

toolbox/mdcode/src/libts/gcp/dataplex.ts
eeshadutta1408: Resolved. I have unexported `_fixEntry` and `_fixEntryLink` from `dataplex.ts` and encapsulated them as private helpers within the module. To allow `CatalogClientMock` and the scenario tests to clean/fix mock values without violating this encapsulation, I added public `fixEntry` and `fixEntryLink` delegation methods directly on the `CatalogClient` class.

toolbox/mdcode/src/libts/manifest.ts
eeshadutta1408: I have removed the predefined `SYSTEM_LINK_ALIASES` mapping and custom `entryLink` alias overrides from the manifest schemas. Instead, `resolveEntryLinkType` and `findAliasForType` now run a generic fallback mapping (e.g. prefix appending/stripping), allowing the sync engine to dynamically support any link types defined in the catalog manifest snapshot without having to maintain a list of known types.

I also refined the local link id extraction: the local link id now stores the target's entire Entry ID (e.g. `bigquery.googleapis.com/projects/...` for schema-joins and `projects/.../glossaries/...` for glossary links). This ensures symmetric representation on pull/push, with dynamic resolution of both `@dataplex` and `@bigquery` entry group namespaces.

PR #13, opened 2026-06-03
feat(mdcode): complete local change detection, sync workflows, and 3-way conflict resolution

sumedhdk22:
## Summary
This PR implements the complete architecture for local change detection, state hashing, and remote conflict resolution for `kcmd`. It combines all the phases of the checksum design.

## Key Features & Implementations
- **State Tracking (`.catalog_state.json`)**: Implemented a flat, system-managed JSON state database to track workspace sync states securely using lockfiles.
- **Stable Checksums**: Integrated the `ohash` library to replace custom JSON stringification. Checksums are resilient to formatting-only changes and safely handle missing/undefined YAML fields.
- **`kcmd status`**: Implemented local modification tracking to compare local workspace changes against the tracked base state.
- **Pull Workflow & Safety Guard**: Integrated a strict "dirty workspace guard" that aborts local updates if untracked/unpushed modifications are present. Added `--allow-partial` and `--force` flags.
- **Push Workflow & 3-Way Merge**: Implemented granular push mechanics with 3-way conflict detection. Automatically detects if aspects were modified remotely before safely pushing local modifications.
- **Dry-Run Mechanics**: Full `--dry-run` support to safely preview mutation sequences without disk writes or API calls.


toolbox/mdcode/src/libts/sources/bq-dataset.ts
nikhilk, line 29: Might have been addressed upstream if I recall. Hopefully will show up when you pull/rebase.

toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -7,22 +7,26 @@ import * as path from 'node:path';
...
-  constructor(catalogPath: string) {
+  constructor(catalogPath: string, source: CatalogSource) {
     this._catalogPath = catalogPath;
+    this._source = source;
   }
 
   async init(): Promise<void> {
     this._index.clear();
+    // Temporary reference to satisfy TS compiler
+    this._source;
```
nikhilk, line 37: I vaguely recall removing the member field. Hopefully that will come through when you pull/rebase.

toolbox/mdcode/src/libts/checksum.ts
```diff
@@ -0,0 +1,66 @@
+import * as crypto from 'node:crypto';
+import { Entry } from './metadata';
+
+/**
+ * Normalizes aspect data recursively by sorting object keys alphabetically.
+ * Preserves the order of array elements, but normalizes objects within arrays.
+ */
+export function normalizeAspectData(data: any): any {
```
nikhilk: Since this is fairly general purpose (not specific to aspect data), perhaps just call this `normalizeJson` (initially when I saw Aspect in the name, I thought data would always be of type Record). And move to a util directory.

toolbox/mdcode/src/libts/checksum.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  if (Array.isArray(data)) {
+    return data.map(item => normalizeAspectData(item));
+  }
+
+  const sortedKeys = Object.keys(data).sort();
+  const normalizedObj: Record<string, any> = {};
+  for (const key of sortedKeys) {
+    normalizedObj[key] = normalizeAspectData(data[key]);
```
nikhilk: Should the normalized representation be an array, so you essentially have a list of key-sorted tuples `key:value` to avoid the assumption of JSON.stringify serializing in the order that keys were set?

toolbox/mdcode/src/libts/checksum.ts
```diff
@@ -0,0 +1,66 @@
...
+export function calculateAspectChecksum(aspectData: any): string {
+  const normalized = normalizeAspectData(aspectData);
+  return sha256(JSON.stringify(normalized));
+}
+
+/**
+ * Calculates the unified entry-level checksum of an Entry by combining the hashes
+ * of its normalized core fields and individual aspect checksums.
+ */
+export function calculateEntryChecksum(entry: Entry): string {
```
nikhilk, line 20: Can you look at https://www.npmjs.com/package/ohash -- this seems to provide functionality we can use to implement computing checksums for entries and aspects.

nikhilk (commented):
Realized I didn't actually click submit after entering comments; PTAL, and thanks.

toolbox/mdcode/src/libts/sources/bq-dataset.ts
sumedhdk22: Yes, resolved during latest rebase.

toolbox/mdcode/src/libts/layouts/documents.ts
sumedhdk22: Yes, resolving after rebasing

toolbox/mdcode/src/libts/checksum.ts
sumedhdk22: I actually replaced the custom normalization logic with 'ohash' per your suggestion, so normalizeAspectData (along with the manual sorting logic) is not required now. Hence it is removed in the latest commit.

toolbox/mdcode/src/libts/checksum.ts
sumedhdk22: By switching to ohash, the stable array/object key ordering serialization is handled natively, so we don't have to manually format tuples anymore.

toolbox/mdcode/src/libts/checksum.ts
sumedhdk22: Thanks. This is very useful. It replaced our custom hashing/normalization logic entirely. Added in the latest commit. PTAL.

sumedhdk22:
The rebasing and git conflict handling was tedious with all the 7 PRs, merged all of them in this single PR.

toolbox/mdcode/src/libts/sync.ts
```diff
@@ -51,75 +126,392 @@ export class CatalogSync {
...
+        if (mdEntry.aspects) {
+           for (const [aspectKey, aspectData] of Object.entries(mdEntry.aspects)) {
+              const localChecksum = calculateAspectChecksum(aspectData);
+              localAspectChecksums[aspectKey] = localChecksum;
+              
+              if (!stateEntry || stateEntry.aspects?.[aspectKey] !== localChecksum) {
+                 // Locally modified aspect
+                 let aspectConflict = false;
+                 
+                 if (remoteEntry && remoteEntry.aspects?.[aspectKey]) {
```
eeshadutta1408, line 342: If an aspect is modified locally (`stateEntry.aspects?.[aspectKey] !== localChecksum`) but has been deleted remotely (`remoteEntry.aspects?.[aspectKey]` is undefined/falsy), this condition block is skipped. As a result, the standard push would incorrectly treat it as safe, push the modified aspect, and silently overwrite the remote deletion right? If yes, should we flag this as a conflict if the aspect existed in the base state (`baseChecksum !== undefined`)?

toolbox/mdcode/src/libts/sync.ts
```diff
@@ -51,75 +126,392 @@ export class CatalogSync {
...
-      if (exist.status != 200 || !exist.result) {
-        const entryGroup = nameParts[5];
-        const entryId = nameParts.slice(7).join('/');
-        const createEntryRes = await this._catalog.createEntry(project, location, entryGroup, entryId, entry);
-        if (createEntryRes.status != 200 || !createEntryRes.result) {
-          return { success: false, details: `Failed to create entry ${entry.name}: ${createEntryRes.message || createEntryRes.status}` };
+        const existRes = await this._catalog.lookupEntry(project, location, entry.name, [...this._snapshot.aspectTypes.keys()]);
+        const entryExists = existRes.status === 200 && !!existRes.result;
+
+        if (!entryExists) {
```
eeshadutta1408, line 258: IIUC, if an entry existed in the last synchronization state (`stateEntry` is defined) but no longer exists remotely (`entryExists` is false), it indicates the entry was deleted remotely. Since the user has modified it locally (which is why we are in the push loop), this represents a conflict.
Instead of recreating the entry remotely under a standard push, should we abort with a conflict error here?

sumedhdk22: Yes, thanks for catching.

Previously, a local update would silently recreate an entry that had been deleted remotely.

I've updated the push logic to implement 3-way merge conflict detection using the base checksums in .catalog_state.json. Now, if an entry exists in the local state database (meaning it was previously tracked) but is missing from the remote catalog, kcmd correctly flags it as a 'Local Update vs Remote Delete' conflict and aborts the push. I've also added explicit test coverage for this scenario in push_conflict_remote_delete.yaml to ensure it's properly caught!"

toolbox/mdcode/src/libts/sync.ts
sumedhdk22: Yes, thanks.

The exact same issue applied here at the aspect level. I've mirrored the 3-way merge conflict fix here as well. If an aspect is tracked in the local state but is missing remotely, it will trigger an aspect-level conflict instead of silently overwriting the remote deletion. We've also included new tests in the YAML scenarios specifically to verify aspect-level remote deletions

PR #14, opened 2026-06-03
feat: Pushing local created files to catalog service by creating new Entry.

qitianyu29:
Call entry.create() during `kcmd.push` for those entries.yaml who created locally.


toolbox/mdcode/src/libts/sync.ts
```diff
@@ -76,6 +76,20 @@ export class CatalogSync {
       const project = nameParts[1];
       const location = nameParts[3];
 
+      const exist = await this._catalog.lookupEntry(project, location, entry.name);
+      if (exist.status != 200 || !exist.result) {
```
nikhilk, line 83: Should we look for 404?

toolbox/mdcode/tests/scenarios/push_new_entry.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - action: push
+
+assert:
+  fileSystem:
+    catalog/custom-entry.yaml:
+      - contains: "name: custom-entry"
+      - contains: "description: Hello world"
+  catalog:
+    entries:
+      - name: projects/p/locations/us/entryGroups/custom-group/entries/custom-entry
```
nikhilk, line 45: Shouldn't we create a new entry with a different name? And ensure it exists as a result of push?

toolbox/mdcode/tests/scenarios/push_new_entry.yaml
```diff
@@ -0,0 +1,44 @@
...
+  fileSystem:
+    catalog.yaml: |
+      scope: entryGroup.p.us.custom-group
+      snapshot:
+        entries:
+          - p.global.custom-type
+
+actions:
+  - action: pull
+  - action: updateEntry
```
nikhilk: Need a createEntry action in the test

toolbox/mdcode/src/libts/sync.ts
```diff
@@ -96,6 +110,7 @@ export class CatalogSync {
       if (res.status !== 200 || !res.result) {
         return { success: false, details: `Failed to update entry ${name}: ${res.message || res.status}` };
       }
+      console.log(`Successfully pushed entry ${entry.name}`);
```
nikhilk: I think eventually we want plug in a verbose output mode, but for now can we strip out debugging statements.

toolbox/mdcode/src/libts/sync.ts
qitianyu29: Yes I initially used that. However the actual status code is 403 for non-exist, which also covers permission denied, so I feel it's would be better to use 200 to represent this case.

nikhilk: OK, seems bit questionable service behavior if the user has permission to the EG to create... but if thats the service behavior, then so be it. Can you add a comment to capture the reasoning in brief.

qitianyu29: Sure, addressed in the latest commit

toolbox/mdcode/tests/scenarios/push_new_entry.yaml
qitianyu29: Done

toolbox/mdcode/src/libts/sync.ts
qitianyu29: Done, removed.

toolbox/mdcode/tests/scenarios/push_new_entry.yaml
qitianyu29: Done

PR #19, opened 2026-06-04
Add OKF GA4 demo

amirhormati:
## Summary

This PR bundles fixes that make the Documents Layout work end-to-end with OKF-style markdown bundles, plus a new `demo/okf/` example that exercises the full round-trip against a real `kb`-scope EntryGroup.

**`sync.ts` — entry ID truncation in push.** Was using `nameParts[7]` to derive the entry ID, which silently dropped everything after the first path segment. Hierarchical names like `tables/events_` or `references/metrics/event_count` were creating entries named just `tables` / `references`. Switched to `nameParts.slice(7).join('/')`.

**`layouts/documents.ts` — accept bare markdown and custom entry types.** Several changes so the layout works with OKF-style markdown bundles (where files have title/description/tags but no `catalogEntry.name` or Dataplex-typed `type`):
- Entry name is now derived from file path unconditionally — no longer ever read from or written to frontmatter.
- Custom entry types are allowed (no longer restricted to `dataplex-types.*`), falling back to `dataplex-types.global.generic` when the frontmatter `type` is missing or not a valid 3-part dotted reference.
- Required type aspect auto-injected into `aspects` so `createEntry` does not 400 on "Missing required Aspect(s): generic".
- Cleaned up a stray dual `console.log` ("Failed" + "Successfully") on the create path and returned a structured error instead.

**`demo/okf/`.** New end-to-end example that pulls/pushes a GA4 OKF wiki bundle (17 markdown files: indexes + references + dataset + table) into a Knowledge Catalog entry group via the Documents Layout.

## Test plan

- [x] Round-tripped 17 OKF-style markdown files through `kcmd push` against a real `kb`-scope EntryGroup; all entries created with correct hierarchical IDs.
- [x] Verified overview aspect content persisted on each entry via `gcloud dataplex entries describe ... --view CUSTOM`.
- [x] `bun setup.ts && kcmd push` in `demo/okf/` creates the entry group and pushes all GA4 markdown files (indexes + references + dataset + table).
- [x] `npm test` — 17/17 mock scenarios pass, including:
  - `create_entry_hierarchy` (covers the `nameParts.slice(7).join('/')` fix — creates `aaa/bbb` and asserts it lands at `catalog/aaa/bbb.yaml`).
  - `push_kb` (covers the demo's `pull` → `updateEntry(overview)` → `push` round-trip with the modified Documents Layout).
- [x] Existing `tests/scenarios/pull_kb.yaml` still passes after dropping the `name:` frontmatter assertions (file path is now the naming assertion).
- [x] Malformed/custom `type:` values fall back to `dataplex-types.global.generic` instead of failing manifest validation.

### DIFF
--- a/toolbox/mdcode/src/libts/layouts/documents.ts
+++ b/toolbox/mdcode/src/libts/layouts/documents.ts
@@ -142,10 +142,7 @@ export function parseMarkdown(content: string): { entry: md.Entry|null; body: st
   const body = lines.slice(endIndex + 1).join('\n');
 
   const entry = (metadata.catalogEntry ?? {}) as md.Entry;
-  const declaredType = metadata.type;
-  entry.type = (typeof declaredType === 'string' && declaredType.startsWith('dataplex-types.'))
-    ? declaredType
-    : DEFAULT_ENTRY_TYPE;
+  entry.type = metadata.type;
   entry.resource = entry.resource ?? {}
   entry.resource.displayName = metadata.title;
   entry.resource.description = metadata.description;
--- a/toolbox/mdcode/src/libts/layouts/documents.ts
+++ b/toolbox/mdcode/src/libts/layouts/documents.ts
@@ -36,15 +36,8 @@ export class DocumentsLayout implements CatalogLayout {
     });
 
     for (const localPath of matches) {
-      try {
-        const content = await fs.promises.readFile(localPath, 'utf8');
-        const { entry } = parseMarkdown(content);
-        const name = entry?.name ?? deriveEntryNameFromPath(localPath, this._catalogPath);
-        this._index.set(name, localPath);
-      }
-      catch (err) {
-        // Skip unreadable/invalid files during indexing
-      }
+      const name = deriveEntryNameFromPath(localPath, this._catalogPath);
+      this._index.set(name, localPath);
     }
   }
 
@@ -66,9 +59,7 @@ export class DocumentsLayout implements CatalogLayout {
     const { entry: parsed, body } = parseMarkdown(content);
 
     const entry: md.Entry = parsed ?? ({ type: DEFAULT_ENTRY_TYPE, resource: {} } as md.Entry);
-    if (!entry.name) {
-      entry.name = name;
-    }
+    entry.name = name;
 
     // Ensure the entry's type aspect is present — Dataplex create requires it.
     entry.aspects = entry.aspects ?? {};
@@ -184,6 +175,7 @@ export function toMarkdown(entry: md.Entry, body: string): string {
     catalogEntry: entryClone
   };
 
+  delete entryClone.name;
   delete entryClone.resource.displayName;
   delete entryClone.resource.description;
   delete entryClone.resource.updateTime;
--- a/toolbox/mdcode/src/libts/layouts/documents.ts
+++ b/toolbox/mdcode/src/libts/layouts/documents.ts
@@ -9,6 +9,7 @@ import * as md from '../metadata';
 import { CatalogLayout } from '../layout';
 
 const OVERVIEW_ASPECT_KEY = 'dataplex-types.global.overview';
+const DEFAULT_ENTRY_TYPE = 'dataplex-types.global.generic';
 
 
 export class DocumentsLayout implements CatalogLayout {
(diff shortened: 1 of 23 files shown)


toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -132,7 +142,10 @@ export function parseMarkdown(content: string): { entry: md.Entry|null; body: st
   const body = lines.slice(endIndex + 1).join('\n');
 
   const entry = (metadata.catalogEntry ?? {}) as md.Entry;
-  entry.type = metadata.type;
+  const declaredType = metadata.type;
+  entry.type = (typeof declaredType === 'string' && declaredType.startsWith('dataplex-types.'))
```
nikhilk: We should just use the type. It doesn't have to be in `dataplex-types`. It can be a custom type.

toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -63,12 +63,17 @@ export class DocumentsLayout implements CatalogLayout {
       throw new Error(`Entry not found: ${name}`);
     }
     const content = await fs.promises.readFile(entryPath, 'utf8');
-    const { entry, body } = parseMarkdown(content);
+    const { entry: parsed, body } = parseMarkdown(content);
 
-    if (!entry) {
```
nikhilk, line 68: Still need to check for this as parseMarkdown can return null... or change that implementation. See other comment below.

toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -117,6 +122,11 @@ export class DocumentsLayout implements CatalogLayout {
...
 }
 
+function deriveEntryNameFromPath(absolutePath: string, catalogPath: string): string {
+  const rel = path.relative(catalogPath, absolutePath);
+  return rel.replace(/\.md$/, '');
+}
+
 export function parseMarkdown(content: string): { entry: md.Entry|null; body: string } {
   const lines = content.split(/\r?\n/);
   if (lines[0] !== '---') {
```
nikhilk, line 123: Will need to change if we want to support front-matter-less markdown files.

toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -38,9 +39,8 @@ export class DocumentsLayout implements CatalogLayout {
       try {
         const content = await fs.promises.readFile(localPath, 'utf8');
         const { entry } = parseMarkdown(content);
-        if (entry && entry.name) {
-          this._index.set(entry.name, localPath);
-        }
+        const name = entry?.name ?? deriveEntryNameFromPath(localPath, this._catalogPath);
```
nikhilk: I wonder if the glob call above can return relative paths (right now its being instructed to return absolute path) and if thats cleaner to derive names, and build absolute paths only to read files.

nikhilk: Additionally, I think we will want to switch to the behavior where entry name is _always_ computed from the file path, rather than sometimes.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: I think it works fine now:

When parsed is null (file has no --- frontmatter, or no closing ---), the ?? falls back to { type: DEFAULT_ENTRY_TYPE, resource: {} }. Then:
  - entry.name gets set from the file path (since it's missing)
  - entry.aspects becomes {} 
  - The generic aspect gets populated (aspects['dataplex-types.global.generic'] = {})
  - The entire file body becomes the overview aspect content
  
  So no-frontmatter files are treated as plain markdown overviews with a default type. That should works properly. Did i miss something?

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: see above.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: I'll add a todo for the glob.

also made the name always be the path.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: done.

amirhormati: ok...so this is more complicated. To use custom types they have to be in the format of project-id.namespace.type

My suggestion is to change the logic here to fall back to generic if the type is not in the right format. Take a look at the new logic and tell me what your thoughts are.

PR #24, opened 2026-06-05
Link Documents Layout entries by parent in Dataplex

amirhormati:
## Summary

Documents Layout entries (the `kb.*` scope) were being pushed flat under the EntryGroup, with no `parentEntry` set. The Dataplex UI showed bundle files as siblings instead of a tree.

This PR derives `parentEntry` from each file's path so bundles render as a tree, with one synthetic root entry per directory. The mapping is:

- A file's parent is its directory's `index` entry (e.g. `references/metrics/event_count.md` → parent `references/metrics/index`).
- An `index.md` file's parent is the parent directory's `index` (e.g. `references/metrics/index.md` → parent `references/index`).
- If a directory has no `index.md` on disk, the layout exposes a **synthetic** index entry with no backing file, so every directory has an index and every entry has a parent.
- The root `index` entry has no parent and becomes a top-level entry under the EntryGroup.

No manifest config is required &mdash; the name `index` is the convention, the rest is automatic. Synthesizing means there's no walk-up logic for missing intermediate indexes; the tree is always fully connected.

### Implementation

- **`layouts/documents.ts`** &mdash; new `_synthesizeIndexEntries()` is run after `init`, `saveEntry`, and `deleteEntry`. It collects every directory containing entries and inserts a synthetic index entry (`null` path) wherever an `index.md` is missing. `loadEntry` returns an empty `dataplex-types.global.generic` entry for synthetic names. `deriveParentLocalName` is now a pure path calculation since every directory is guaranteed to have an index. `toMarkdown` strips `resource.parent` from persisted frontmatter (it's derived from path, not user-editable).
- **`snapshot.ts`** &mdash; resolves the local-name parent to a full Dataplex resource name via `source.serviceName(...)` before push.
- **`sync.ts`** &mdash; adds `parent_entry` to the update mask (currently a no-op on the API side, but harmless and future-proof).
- **`manifest.ts`, `layout.ts`** &mdash; no manifest changes were needed; behavior is intrinsic to the Documents Layout.

## Test plan

- [x] `npm test` &mdash; 17/17 scenarios pass.
- [x] Local layout sanity check against the real OKF tree (every directory has a real `index.md`): 17 entries, single root, full parent chain.
- [x] Local layout check with intermediate `index.md` files removed: synthetic entries inserted for the missing directories, parent chain still walks to the single root.
- [x] Live push against `xxokf_ga4` with all `index.md` files present: 17 entries materialized in Dataplex with the expected 3-level tree:
  ```
  index (root)
  +-- datasets/index
  |   +-- datasets/ga4_obfuscated_sample_ecommerce
  +-- tables/index
  |   +-- tables/events_
  +-- references/index
      +-- references/joins/index
      |   +-- references/joins/events___ads_clickstats
      +-- references/metrics/index
          +-- references/metrics/{event_count, user_count, ...}
  ```
- [x] Live push against `xxokf_ga4` with **all `index.md` files deleted locally**: same 17 entries, identical tree shape &mdash; all 6 directory indexes were synthesized on the fly and created in Dataplex.

## Known limitation

The Dataplex `modifyEntry` API silently ignores `parent_entry` in the update mask, so already-existing entries created before this change won't get `parentEntry` backfilled by a re-push &mdash; they need to be deleted and re-created. Fresh pushes against a new EntryGroup work correctly because `createEntry` accepts `parentEntry` in the body.

### DIFF
--- a/toolbox/mdcode/src/libts/layouts/documents.ts
+++ b/toolbox/mdcode/src/libts/layouts/documents.ts
@@ -10,18 +10,19 @@ import { CatalogLayout } from '../layout';
 
 const OVERVIEW_ASPECT_KEY = 'dataplex-types.global.overview';
 const DEFAULT_ENTRY_TYPE = 'dataplex-types.global.generic';
+const INDEX_NAME = 'index';
 
 
 export class DocumentsLayout implements CatalogLayout {
 
   private _catalogPath: string = '';
-  private readonly _directoryIndex?: string;
 
-  private readonly _index = new Map<string, string>();
+  // Maps entry name -> local file path, or null for synthetic directory-index
+  // entries that don't have a backing file on disk.
+  private readonly _index = new Map<string, string | null>();
 
-  constructor(catalogPath: string, directoryIndex?: string) {
+  constructor(catalogPath: string) {
     this._catalogPath = catalogPath;
-    this._directoryIndex = directoryIndex;
   }
 
   async init(): Promise<void> {
@@ -41,11 +42,12 @@ export class DocumentsLayout implements CatalogLayout {
       const name = deriveEntryNameFromPath(localPath, this._catalogPath);
       this._index.set(name, localPath);
     }
+
+    this._synthesizeIndexEntries();
   }
 
   entryExists(name: string): boolean {
-    const entryPath = this._index.get(name);
-    return !!entryPath && fs.existsSync(entryPath);
+    return this._index.has(name);
   }
 
   listEntries(): string[] {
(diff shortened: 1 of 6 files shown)


toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -10,13 +10,16 @@ import { CatalogLayout } from '../layout';
...
+const INDEX_NAME = 'index';
 
 
 export class DocumentsLayout implements CatalogLayout {
 
   private _catalogPath: string = '';
 
-  private readonly _index = new Map<string, string>();
+  // Maps entry name -> local file path, or null for synthetic directory-index
+  // entries that don't have a backing file on disk.
```
nikhilk, line 21: Wondering if the implementation should just create those index files at the time of the pull operation (including the root index.md). Probably would simplify things...

toolbox/mdcode/src/libts/layouts/documents.ts
```diff
@@ -39,37 +42,50 @@ export class DocumentsLayout implements CatalogLayout {
...
       this._index.set(name, localPath);
     }
+
+    this._synthesizeIndexEntries();
   }
 
   entryExists(name: string): boolean {
-    const entryPath = this._index.get(name);
-    return !!entryPath && fs.existsSync(entryPath);
+    return this._index.has(name);
```
nikhilk, line 50: Some concerns about this; we'd want to support the user (or agent) creating files and coming into existence by virtue of that.

nikhilk (commented):
Added some questions... ptal.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: I explicitly wanted to do this at the time of push so the agent who are trying to navigate this using KC can interact with it similar to a file system (as much as possible).

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: Valid concern but i think the old implementation also didn't do this.

The old code also wouldn't see a freshly-dropped file, because _index.get(name) returns undefined for any name not picked up by init()'s glob, and the fs.existsSync check only ran after a successful map lookup. So my change made the contract more explicit without making the behavior more restrictive.

I think it's a fair design choice to support this for both creating and deleting files....but i don't think the implementation right now handles that case. I can add a TODO to fix this? WDYT?

amirhormati: I should add that the code before handled "deletion" (which is probably not as important) but it didn't handle creation of new entries.

toolbox/mdcode/src/libts/layouts/documents.ts
nikhilk: We'll need to do it in pull, if we want to have the local catalog apis mirror the service. I will tackle as a follow-up since we want to submit.

toolbox/mdcode/src/libts/layouts/documents.ts
nikhilk: The code to implement creation is in the dev branch. There is a massive amount of change there, and I think the merge is going to be quite involved. I'll figure out that as a followup.

nikhilk (commented):
I'll address things as followup. Can you sync if needed and then accept the pull request.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: Let's discuss. I am probably not understanding something.

toolbox/mdcode/src/libts/layouts/documents.ts
amirhormati: ack

PR #58, opened 2026-06-15
okf: clarify SPEC v0.1 ambiguities and add a trust/safety section

therealfatleeadama:
These are clarifying edits to `okf/SPEC.md` (OKF v0.1 — Draft). They tighten under-specified or self-conflicting points **without changing the format's intent** — every change clarifies existing behavior, and none alters a currently-conformant bundle. CLA signed.

### Must-fix
- **§6 ↔ §11 contradiction.** §6 stated index files contain *no* frontmatter; §11 permitted `okf_version` in the root `index.md` frontmatter. A conformance checker can't satisfy both. Reconciled in both sections: the single-key `okf_version` block is allowed only in the bundle-root `index.md`.

### Correctness / conformance
- **§3.1 + §9.** A plain `README.md` (no frontmatter) made an entire bundle non-conformant, because rule 1 required frontmatter in *every* non-reserved `.md`. Added `README.md`/`LICENSE.md`/`CONTRIBUTING.md` as reserved/ignored, scoped rule 1 to concept documents, and made rule 3 mechanically checkable.
- **§5.1.** A bundle-root absolute link (`/...`) resolves to the **repository** root, not the bundle root, in standard renderers (GitHub/GitLab/Obsidian) when a bundle ships as a subdirectory (§3) — which breaks the "readable without tooling" promise for the recommended link form. Documented the trade-off and corrected the stale "stable when moved" rationale.
- **§5.2a (new).** Defined how a link target maps to a Concept ID (strip `.md`; trailing slash → directory `index.md`).

### Clarity
- **§4.1 + §5.3.** `type` is a presentation/routing **hint**, not an interoperability contract (it's free-form and unregistered); and the path-as-identity trade-off is stated explicitly so the "tolerate broken links" rule reads as a deliberate choice.
- **§4.** Declared Concept IDs / link targets **case-sensitive** (a cross-platform round-trip hazard since IDs are paths).
- **§6 / §7.** Pinned a deterministic index-entry grammar (the `" - "` separator collides with hyphens in titles); clarified that git history is authoritative over `log.md`.

### Security
- **§12 (new) — Trust and safety.** Consumers MUST treat bundle contents as untrusted **data**, never instructions (prompt-injection), and OKF provides no built-in authenticity guarantee. Relevant because bundles are explicitly meant to be exchanged across orgs and consumed by agents.

Two opt-in *additions* — an optional stable `id` field and footnote-style citations — were considered but deliberately left out of this PR to keep it to clarifications only. Happy to open them as a follow-up if of interest.

### DIFF
--- a/okf/SPEC.md
+++ b/okf/SPEC.md
@@ -76,8 +76,10 @@ sense for the knowledge being captured.
 
 ```
 path/to/bundle/
+├── .okfignore                    # Optional. Marks .md files that are not concepts (§3.2).
 ├── index.md                      # Optional. Directory listing for progressive disclosure.
 ├── log.md                        # Optional. Chronological history of updates.
+├── README.md                     # Non-concept file; listed in .okfignore.
 ├── <concept>.md                  # A concept at the bundle root.
 └── <subdirectory>/               # Subdirectories organize concepts into groups.
     ├── index.md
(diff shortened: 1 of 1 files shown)


okf/SPEC.md
```diff
@@ -97,12 +97,23 @@ A bundle MAY be distributed as:
...
+| `index.md`           | Directory listing. See §6.                   |
+| `log.md`             | Update history. See §7.                      |
+| `README.md`          | Human-oriented readme. Ignored by consumers. |
+| `LICENSE.md`         | License text. Ignored by consumers.          |
+| `CONTRIBUTING.md`    | Contribution guide. Ignored by consumers.    |
+| `CODE_OF_CONDUCT.md` | Community health file. Ignored by consumers. |
+| `SECURITY.md`        | Community health file. Ignored by consumers. |
+| `CHANGELOG.md`       | Changelog. Ignored by consumers.             |
+| `CLAUDE.md`          | Agent instructions. Ignored by consumers.    |
+| `AGENTS.md`          | Agent instructions. Ignored by consumers.    |
```
marco-msg-ferrari: Hello @therealfatleeadama .
I agree with the purpose of this change, but having a long list of "whitelisted" files doens't look like a  long lived solution. Would you consider adding a sort of ".okfignore" files so everybody can manage it?

therealfatleeadama: Agreed -- the list isn't sustainable (already 3 --> 10 in this PR).

Real question: should conformance be able to flag a `.md` that's missing frontmatter? That decides whether we need a boundary at all.

**A. Concept = any `.md` with OKF frontmatter.** Drop the list, keep only `index.md`/`log.md` reserved. README/LICENSE/CHANGELOG/AGENTS have no frontmatter, so they're ignored. Nothing to maintain, no new file. Touches 3.1 and 9.1. Cost: a forgotten frontmatter block is silently ignored, not flagged.

**B. `.okfignore`.** Keeps strict detection, extensible. Cost: a control file consumers must parse, gitignore-style glob semantics to spec out, and a new failure mode if a consumer skips it.

I'd go A -- removes the list instead of moving it into a config file. B's fine if forgot-frontmatter detection matters to you. Tell me which and I'll push it.

marco-msg-ferrari: Personally I'd go B in the spirit of been able to create a "valid" OKF bundle. Having some files silently ignored IMHO it's much worst than having an optional ignore files.

therealfatleeadama: Pushed B in 30a0bac.

- 3.1: reserved list back to just `index.md`/`log.md`.
- 3.2 (new): optional root `.okfignore` -- one pattern per line, `#` comments, gitignore-style glob subset (`*` `?` `**`, a no-slash pattern matches a basename at any depth, leading `/` anchors to the bundle root). A matching `.md` is excluded: no OKF meaning, no frontmatter required. Spec ships a sample listing README/LICENSE/CHANGELOG/CONTRIBUTING/etc.
- 9: a `.md` that's neither reserved nor ignored and lacks frontmatter is now non-conformant, so a validator surfaces it instead of silently dropping it -- the "valid bundle" property you asked for.

Scoped out of v0.1 to keep it minimal (both noted in-text as a possible later minor bump): negation patterns (`!`) and per-directory `.okfignore` files. Say if you'd want either in now.

andreassiemes:
Independent confirmation for two of these from a second implementation.

I build [OPI / org-as-code](https://github.com/andreassiemes/org-as-code), a typed governance layer that ships an OKF export profile since v0.6 (OPI stays the strict source; the OKF bundle is the projection agents read). On July 4 I generated a full bundle from a worked example (22 concept files: org units, committees, decisions, one drift entry) and verified it against `SPEC.md` clause by clause — 123 checks.

Two edges stopped me, and they are exactly the first two items in this PR:

- **§6 ↔ §11:** my first validator, built from §6 read in isolation, rejected every bundle that follows §11 — including the repo's own samples. The reconciliation proposed here (single-key `okf_version` block, bundle-root `index.md` only) is what I implemented on the producer side; my spec now carries a caveat warning implementers about §6 ([opi-v0.6.md](https://github.com/andreassiemes/org-as-code/blob/main/spec/opi-v0.6.md), §1).
- **Plain `README.md`:** I dropped one into the bundle folder out of habit — 123 green checks on the payload, non-conformant bundle as a whole. Reserving the common auxiliary files and scoping rule 1 to concept documents fixes the realistic failure mode: READMEs get added by colleagues and by tooling, not by intent.

Example bundle that hit both: [examples/okf-export](https://github.com/andreassiemes/org-as-code/tree/main/examples/okf-export). Happy to re-verify against the reconciled wording once this lands — a second independent producer exercising the same clauses is cheap regression coverage for the spec.

PR #64, opened 2026-06-15
fix: align REQUIRED_FRONTMATTER_KEYS with OKF spec §4.1

tcconnally:
## Problem

`REQUIRED_FRONTMATTER_KEYS` in `document.py` was set to `("type", "title", "description", "timestamp")`, treating four fields as required. However, [OKF v0.1 §4.1](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#41-frontmatter) states that only `type` is REQUIRED. `title`, `description`, and `timestamp` are "Recommended" — not required.

This means the reference implementation was rejecting valid OKF bundles that conform to the spec. For example, a concept document with only `type` and `tags` (both spec-valid) would fail `validate()`.

## Fix

- Changed `REQUIRED_FRONTMATTER_KEYS` to `("type",)` — only `type` is required per spec.
- Updated `test_validate_rejects_missing_required_keys` → `test_validate_rejects_missing_type` to test that only missing `type` triggers validation failure.
- Added `test_validate_accepts_minimal_frontmatter` to verify that `{"type": "X"}` passes validation (spec-minimal bundle).

## Verification

All 6 document tests pass:
```
tests/test_document.py::test_roundtrip_preserves_frontmatter_and_body PASSED
tests/test_document.py::test_parse_no_frontmatter_treats_all_as_body PASSED
tests/test_document.py::test_unterminated_frontmatter_raises PASSED
tests/test_document.py::test_validate_rejects_missing_type PASSED
tests/test_document.py::test_validate_accepts_minimal_frontmatter PASSED
tests/test_document.py::test_validate_accepts_full_frontmatter PASSED
```

## Impact

This is a conformance fix. Without it, any downstream producer that emits spec-minimal bundles (type-only frontmatter) would get false validation failures from the reference validator. This is especially important for programmatic producers that may not have human-readable `title`/`description` fields available at generation time.

### DIFF
--- a/okf/SPEC.md
+++ b/okf/SPEC.md
@@ -1,6 +1,6 @@
 # Open Knowledge Format (OKF)
 
-**Version 0.2 (Draft)**
+**Version 0.2**
 
 OKF is an open, human- and agent-friendly format for representing
 *knowledge*: the metadata, context, and curated insight that surrounds


tcconnally:
Friendly ping — CLA signed and checks passing. Happy to iterate on any feedback.

tcconnally:
Closing this PR — upstream `main` has already adopted the OKF §4.1 frontmatter change (only `type` is required). I've reset the branch to current `main`; there is no remaining diff. Thanks!

PR #110, opened 2026-06-19
Allow absolute links when detecting edges

### LINKED ISSUE #48: visualize: 0 edges for spec-recommended absolute links and CommonMark link titles

seeholza:
Fixes #48 (replaces #109)

### DIFF
--- a/okf/src/enrichment_agent/viewer/generator.py
+++ b/okf/src/enrichment_agent/viewer/generator.py
@@ -51,10 +51,14 @@ def _extract_links(body: str, doc_dir: Path, bundle_root: Path) -> list[str]:
     bundle_root_resolved = bundle_root.resolve()
     for m in _LINK_RE.finditer(body):
         target = m.group(1)
-        if "://" in target or target.startswith("/"):
+        if "://" in target:
             continue
         try:
-            resolved = (doc_dir / target).resolve().relative_to(bundle_root_resolved)
+            if target.startswith("/"):
+                document_location = (bundle_root / target[1:])
+            else:
+                document_location = (doc_dir / target)
+            resolved = document_location.resolve().relative_to(bundle_root_resolved)
         except ValueError:
             continue
         rel = resolved.as_posix()


seeholza:
Ah, just saw that this is also addressed in #66.

Edit: On second inspection, the fix in #66 mixes this fix with a bigger overhaul aligning the specs across the repo. I will leave to maintainers to decide.

PR #121, opened 2026-06-20
docs(toolbox): fix the kcagent CLI name and the demo_ecommerce cleanup target

mftnakrsu:
Reduced after rebasing onto `main`.

The original PR also corrected doc paths and a CLI-name reference under
`agents/enrichment/**`, but that folder was removed upstream in #128, so
those four changes are now obsolete and have been dropped. What remains is
the still-valid `toolbox/enrichment/README.md` fix:

- The **CLI** section named the `kcenrich` tool, while every example in the
  README invokes `kcagent` — aligned the prose with the actual tool name.
- The **Clean up** step ran `bq rm -r …:demo-dataset`, but the walkthrough
  creates the `demo_ecommerce` schema, so the command removed nothing —
  pointed it at the schema that is actually created.


mftnakrsu:
@googlebot I signed it!

mftnakrsu:
@amirhormati a small `toolbox/enrichment` doc fix (the `kcagent` CLI name + the `demo_ecommerce` cleanup target). Reduced after #128 removed the `agents/` folder, so only the still-valid toolbox change remains. Checks are green and it's mergeable. Review when convenient — thanks!

PR #122, opened 2026-06-20
fix(viewer): skip unreadable/non-UTF-8 files instead of aborting the whole bundle

mftnakrsu:
## What
`enrichment_agent visualize` aborts the entire bundle if a single concept file can't be read, decoded, or parsed.

## Why
In `viewer/generator.py`, `_walk_concepts` reads every concept with `read_text(encoding="utf-8")` and parses it, but only caught `OKFDocumentError`. Three realistic failure modes still escaped and killed the whole `visualize` command:

- invalid UTF-8 bytes → `UnicodeDecodeError`
- an unreadable file → `OSError`
- a malformed frontmatter *value* — e.g. an out-of-range `timestamp: 2026-13-45` — makes PyYAML's implicit resolver raise a **bare `ValueError`** that `OKFDocument.parse` does not wrap

One bad file takes down the whole graph. The sibling `bundle/index.py::_load_doc` already tolerates all of these (`except Exception: return None`).

## Change
- Skip the offending file by catching `(ValueError, OSError)`. `ValueError` subsumes `OKFDocumentError` and `UnicodeDecodeError` (both subclasses) as well as PyYAML's bare `ValueError`; `OSError` covers read errors. The now-unused `OKFDocumentError` import is dropped.
- Add two regression tests: `test_unreadable_concept_file_is_skipped` (invalid UTF-8) and `test_malformed_timestamp_concept_file_is_skipped` (bad timestamp → bare `ValueError`), each asserting the remaining concepts still render.

Tests: `pytest okf/tests/test_viewer.py` → 8 passed.

### DIFF
--- a/okf/src/reference_agent/viewer/generator.py
+++ b/okf/src/reference_agent/viewer/generator.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
-from reference_agent.bundle.document import OKFDocument, OKFDocumentError
+from reference_agent.bundle.document import OKFDocument
 
 _INDEX_NAME = "index.md"
 _LINK_RE = re.compile(r"\]\(([^)\s]+\.md)(?:#[A-Za-z0-9_\-]*)?\)")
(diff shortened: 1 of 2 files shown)


mftnakrsu:
@amirhormati a small viewer robustness fix — skip unreadable/non-UTF-8 files instead of aborting the whole visualize run. Rebased onto the latest `main`; checks are green and it's mergeable. Would appreciate a review when you get a chance. Thanks!

Issue #158, opened 2026-06-28
Convention for maintenance signals — freshness, confidence, contradiction (additive under §4.1)


inkxel:
OKF nails the static format. What I believe is missing is a convention for the signals a knowledge base needs once it's *living* — once something's writing to it, reading from it, and trusting it weeks later. Three of them: how fresh a fact is (`timestamp` tells you when content last *changed*, not when it was last *verified* still true), how much you trust a given claim, and what happens when two concepts disagree (links are untyped, and the spec's quiet on conflicting content — no `contradicts`/`supersedes`).

Candidly, I opened this before I'd read the rest of the week's trust threads properly, and there's more overlap than I gave credit for — so let me reframe. I don't think this should be a competing proposal; the individual pieces are already further along in other hands, and I'd rather point at them than restate them:

- On **confidence**, #151 (@K4uP) is the real version — a reliability axis with production behind it. Mine should defer to that.
- On **freshness**, #97 (elevate `timestamp` for staleness) and #120 (keeping bundles healthy as they evolve) are already carrying it; my `last_verified` is the same idea.
- On **provenance** more broadly, there's already a cluster of threads (verification, SOURCES, citation); the over-time / decay angle rides alongside that work, not into it.

So what I think is actually left that's differentiated, once you strip the overlap, is two things. First, the framing: freshness, confidence, and contradiction are really one *maintenance* surface, not three unrelated asks — and it'd help to treat them as a coordinated convention (or an `EXTENSIONS.md`-style registry) so they don't land piecemeal. Second, a contradiction edge I haven't seen elsewhere: #151's conflict signal is about competing claims *within* a concept, but I'm after the *cross-concept* case — a typed link where one concept supersedes or contradicts another (an "HQ moved to Austin" note quietly obsoleting "HQ in Denver"). That's the rot that actually kills a living KB — not "how sure am I about this claim" but "this whole concept is wrong now because that one replaced it." Feels closest to #148's typed relationships.

None of this needs a core change — §4.1 already lets producers add keys and tells consumers to preserve them, so it's legal OKF today; the only thing missing is *agreed* names.

For what it's worth, I'm not theorizing — I've run a multi-domain, human-authored KB on these signals for months, and Throughline (an OKF reader/writer, https://github.com/inkxel/throughline) does freshness + confidence + contradiction end to end. Happy to be a second worked example next to #151's.

[...]

Issue #151, opened 2026-06-27
Proposal: graded confidence — an epistemic-reliability axis, distinct from integrity (#140) and citation (#92/#94)


K4uP:
Following up on my comment in #95, after reading how the trust surface developed
this week (#140 cryptographic provenance, #92/#94 SOURCES + inline citation, the
new SPEC §12 trust-and-safety in #58). Those cover three distinct senses of
"trust" — and there's a fourth still open, which is the one we've had to solve in
production.

**Four axes, so we don't talk past each other:**

| Axis | "Trust that…" | Where it lives now |
|---|---|---|
| Authenticity / integrity | the bundle is from who it claims, unaltered | #140 (signing, SHA-256 manifest) |
| Safety | content is data, not instructions (prompt-injection) | #58 §12 |
| Groundedness / citation | a claim is linked to evidence | #92 SOURCES, #94 inline citation |
| **Reliability (this proposal)** | **how much to believe the claim itself** | **— open —** |

A bundle can be perfectly signed (#140), injection-safe (#58), and have every
claim cited (#94) — and an agent still can't tell a claim *verified against the
live system* from one that merely *restates vendor documentation* (often stale or
aspirational) from one that's *inferred and unverified*. A citation proves a claim
is grounded; it doesn't grade the ground. For a retrieval agent that's tolerable.
For an agent that **acts on a real system**, it's where damage happens. (This is
the dimension #95 explicitly scoped out.)

**What we run in production** — single-domain agent KB, enterprise IT service
management, ~250 atomic notes over ~140 build cycles, agents acting on live
customer systems:
- a graded **`confidence`** value per concept in frontmatter —
  `HIGH | MEDIUM | LOW | UNVERIFIED` — orthogonal to "is it cited";
- a small **evidence-kind vocabulary** on claims (`observed-on-system` /
  `documentation` / `partner-confirmed` / `inferred`) — this types the *kind* of
  evidence, not just its URL, so it's a different sense of "provenance" than 
 #140's
  *origin* provenance;
- an explicit **trust ordering** when two cited sources disagree
  (`observed > partner-confirmed > observed+doc > doc > inferred`) and a
  **conflict** rule: *document both sides, never silently pick*.

The taxonomy is domain-specific; the *axis* is not. Every producer whose agents
*act* reinvents some version of "how sure are we", and without a shared optional
slot none of it survives the cross-org hop OKF exists for.

[...]

Issue #140, opened 2026-06-23
Provenance / verification: an additive layer for OKF bundles


Dynamicfeedai:
Thanks for OKF - the format is clean, and the "producers MAY include any additional keys" rule makes it easy to build on.

One gap shows up the moment agents act on shared bundles: a consumer has no way to verify that a bundle came from who it claims and hasn't been altered. The metadata stops at timestamp - there's no signature or integrity check. (Karpathy's original LLM-wiki gist has comments raising the same worry: stale synthesis pages that still look authoritative.)

I put together a small, additive layer that stays inside the spec:

optional provenance frontmatter keys (source, source_url, licence, measured_at) - additional keys the spec already permits
a root okf.manifest.json (not a reserved filename) that lists a SHA-256 of every file plus a provenance envelope, Ed25519-signed over its canonical form
a standalone verifier that re-hashes the files and checks the signature against the issuer's published key
Drop the manifest and it's a plain OKF bundle again - no fork, no required SDK. Repo (Apache-2.0): https://github.com/dynamicfeed/signed-okf

Sharing in case it's useful to the ecosystem, or as a reference if OKF ever wants to address provenance natively. Happy to align naming and conventions with the spec. Either way, thanks for shipping this.

K4uP:
This nails the *integrity / authenticity* axis — "is this bundle from who it claims,
and unaltered." I opened #151 for the complementary *reliability* axis: even a
perfectly signed, unaltered bundle can carry claims of very different trust
(verified-against-the-live-system vs. restating vendor docs vs. inferred). A signature
proves *who said it*; it doesn't grade *how much to believe it*.

It's a different sense of "provenance" than the cryptographic one here — happy to align
naming so the word doesn't end up carrying two meanings across the spec.

Dynamicfeedai:
Agreed, these are two distinct axes worth separating cleanly.

Integrity / authenticity (this issue): is the bundle from who it claims, and unaltered. A signature settles that, and only that. It proves who said it.

Reliability (#151): how much a consumer should believe a given claim, even when the signature is perfect. A claim verified against the live system, one restating vendor docs, and one that is inferred are all equally signable and very differently trustworthy.

On naming, I would keep "provenance" for the cryptographic axis here and use "reliability" or "assurance" for #151, so the word does not carry two meanings across the spec.

A shape that has worked for us in a live verifiable-data layer is a per-claim block that keeps the axes separate:

```json
"verification": {
[...]

Issue #94, opened 2026-06-18
Proposal: Inline citation as default convention for coupling claims to evidence


magnus919:
## Summary

Adopt inline hyperlinks on claim text (rather than footnote-style Citations sections) as the default citation convention in OKF concept bodies. This couples each claim directly to its supporting evidence, eliminating the mental mapping step between a claim in the body and a citation at the bottom.

## Motivation

OKF's current Citations section is a separate block at the bottom of a concept file. This works for academic citation conventions where a numbered reference list is expected, but it breaks the coupling between a claim and its evidence. A reader (human or agent) encountering a claim must scroll to the bottom, find the citation, and mentally map it back to the claim. For agent consumers, this means an extra parse-and-join step to reconstruct the claim-to-source mapping.

## Proposal

Recommend inline citation as the primary convention. Inline citations SHOULD be placed on the claim text itself: `[the specific claim](url)` rather than `claim text ([source](url))`. The link reads as part of the sentence and the evidence is adjacent to the assertion.

The existing Citations section remains available for:
- Grouped references that support multiple claims
- Sources that are cited indirectly (e.g., see also)
- Bibliographic metadata too long for inline placement

## Relationship to Spec

This does not require a spec change. It is a best-practice convention for body content. Permissive consumers read the body regardless of citation style. Informed consumers can extract citation graphs by parsing inline links instead of cross-referencing a separate section.

## Complementary Mechanism

Inline citations handle claim-level evidence. The proposed SOURCES body section (issue #92) handles file-level provenance between concepts. Together they form a complete evidence chain: every sentence-level claim traces to a URL, and every file-level claim traces to supporting files one layer deeper.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

magnus919:
## Prior Art

The artifact-pyramids agent skill (github.com/groktopus/artifact-pyramids) has used inline citation as its default convention since launch. In practice, the combination of SOURCES sections (file-level provenance between concepts) and inline links (sentence-level evidence for claims) means every assertion in every file traces concretely to its support. There are no orphaned claims and no footnote-chasing step for consuming agents.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

vonargo:
Coming at this from the consumer side ;) 
[...]

PR #159, opened 2026-06-29
okf: propose optional reliability frontmatter convention (reliability axis, #151)

Dynamicfeedai:
## Summary

Proposes one optional frontmatter key, `reliability`, under the existing SPEC §4.1 extension allowance ("Producers MAY include any additional keys"), for the epistemic-reliability axis discussed in #151 and #158: how much a consumer should believe a claim, even when it is correctly signed (integrity, #140) and correctly cited (groundedness, #92 / #94).

It standardizes only an agreed name and shape. `type` remains the only required field; a minimal producer ignores `reliability` entirely. This adds one self-contained proposal doc and changes nothing in the core SPEC.

## What it adds

`okf/proposals/reliability.md`, in SPEC voice, as a maturity ladder (floor `confidence` + `basis`, then opt-in corroboration and maintenance tiers), with YAML frontmatter examples and four honesty rules:

1. signed is not verified;
2. `verified: true` requires `sources >= 2`;
3. an `UNVERIFIED` band or a disputed conflict cannot be `verified`;
4. a computed `score` must be recomputable from `signals` and ordinally coherent with the band (the band also reflects a corroboration ceiling, so a single source caps at `MEDIUM`).

## Validation

Consistent with OKF referencing schemas rather than subsuming them, an optional JSON Schema (draft 2020-12) is hosted externally and referenced, not added to the tree:

- Schema: https://dynamicfeed.ai/schemas/okf-reliability-v1.json
- Worked examples: https://dynamicfeed.ai/schemas/okf-reliability-examples.json

The schema encodes the honesty rules as constraints and validates green against the three independent production shapes converging in #151 (a live-data layer, a multi-version corpus, and a multi-domain human-authored KB) plus a set of malformed cases.

## Placement

Placed as a standalone proposal doc to keep the core SPEC untouched while the axis is discussed. Happy to fold it into SPEC §4.1, or into an `EXTENSIONS.md` registry as suggested in #158, whichever the maintainers prefer.

Tracks #151 and #158.

### DIFF
--- a/okf/proposals/reliability.md
+++ b/okf/proposals/reliability.md
@@ -48,6 +48,7 @@ These keep the axis trustworthy rather than decorative:
 2. `verified: true` MUST assert `sources >= 2`.
 3. An `UNVERIFIED` band, or a disputed conflict, MUST NOT be `verified`.
 4. When `score` is present it MUST be recomputable from `signals` and ordinally coherent with `confidence`. Note the band is **not** a pure re-banding of the score: it also reflects a corroboration ceiling, so a single-source reading caps at `MEDIUM` even when its raw score is high.
+5. A disputed conflict caps `confidence` at `MEDIUM` (`HIGH` excluded). An open disagreement is a corroboration failure, so the claim cannot be `HIGH`; the position the trust ordering selects MAY still carry up to `MEDIUM` (a conservative producer MAY floor to `LOW`), while `disputed: true` stays flagged. This is the same corroboration ceiling as rule 4, applied to disputes: it gives a consumer both the dispute flag and a usable graded answer rather than discarding the prevailing reading.
 
 ## 4. Examples
 
--- a/okf/proposals/reliability.md
+++ b/okf/proposals/reliability.md
@@ -1,0 +1,109 @@
+# Reliability (optional convention)
+
+**Status:** Draft proposal, tracking [#151](https://github.com/GoogleCloudPlatform/knowledge-catalog/issues/151) and [#158](https://github.com/GoogleCloudPlatform/knowledge-catalog/issues/158). Not yet part of the SPEC; offered for discussion.
+
+This document proposes one optional frontmatter key, `reliability`, under the existing extension allowance in SPEC §4.1 ("Producers MAY include any additional keys"). It standardizes only an agreed name and shape; it does not change the required surface. `type` remains the only required field, and a minimal producer ignores `reliability` entirely.
+
+## 1. Motivation
+
+A concept can be correctly signed (integrity) and correctly cited (groundedness) and still leave a consuming agent unable to tell a claim *verified against a live system* from one that merely *restates vendor documentation* from one that is *inferred*. A citation proves a claim is grounded; it does not grade the ground. For a retrieval agent that is tolerable. For an agent that **acts** on a real system, that grade is where damage is avoided.
+
+This is the fourth trust axis, distinct from the three already in flight:
+
+| Axis | "Trust that…" | Where it lives |
+|---|---|---|
+| Integrity | the bundle is from who it claims, unaltered | #140 |
+| Safety | the content is data, not instructions | #58 §12 |
+| Groundedness | a claim is linked to evidence | #92 / #94 |
+| **Reliability** | **how much to believe the claim itself** | **this proposal, #151** |
+
+## 2. The `reliability` key
+
+`reliability` is an optional YAML mapping in a concept's frontmatter. It is a **maturity ladder, not a mandate**: a producer adopts only the tiers it can honestly populate.
+
+### 2.1 Floor
+
+- `confidence` (required within the object) — an ordinal band, one of `HIGH | MEDIUM | LOW | UNVERIFIED`. This is the interoperable surface consumers filter on.
+- `basis` — how the claim was obtained, one of `live-source | partner-attested | vendor-doc | forecast | computed | inferred`, ordered most to least authoritative when sources disagree.
+
+### 2.2 Corroboration tier
+
+- `sources` — integer count of independent sources behind the claim.
+- `verified` — boolean. See the honesty rules in §3.
+- `score` — an optional `0..1` float, present **only** when actually computed from `signals`. Absent means "not computed", not "hidden".
+
+### 2.3 Maintenance tier
+
+- `conflict` — a first-class disagreement state, distinct from merely uncorroborated. While `disputed` is true, both `positions` are retained and a `resolution` records what the trust ordering selected without discarding the other side.
+- `validity` — an applicability window (when or for which versions the claim holds), a separate clock from measurement recency, often version-keyed. Anonymous expiry (`valid_until`) lives here; **named** supersession is a typed cross-concept edge tracked separately in #158 / #148, not on this object.
+- `freshness` — measurement recency (`as_of`, `state`).
+- `signals` — the transparent inputs `confidence` and `score` are recomputed from (`signed`, `corroborated`, `fresh`).
+- `assessed_at` and `lifecycle` — for producers that persist a grade history. A recompute-per-read producer sets `assessed_at` and omits `lifecycle`; a stored-ladder producer logs transitions in `lifecycle`.
+
+## 3. Honesty rules
+
+These keep the axis trustworthy rather than decorative:
+
+1. **Signed is not verified.** `signals.signed` proves integrity only and MUST NOT be coupled to `verified`.
+2. `verified: true` MUST assert `sources >= 2`.
+3. An `UNVERIFIED` band, or a disputed conflict, MUST NOT be `verified`.
+4. When `score` is present it MUST be recomputable from `signals` and ordinally coherent with `confidence`. Note the band is **not** a pure re-banding of the score: it also reflects a corroboration ceiling, so a single-source reading caps at `MEDIUM` even when its raw score is high.
+
+## 4. Examples
+
+A live, single-source concept:
+
+```markdown
+---
+type: Live Metric
+title: US CPI inflation (annual)
+description: Headline year-over-year CPI-U inflation for the United States.
+resource: https://dynamicfeed.ai/v1/facts?tool=us_economy&indicator=cpi
+timestamp: 2026-06-29T00:00:00Z
+reliability:
+  confidence: MEDIUM
+  basis: live-source
+  score: 0.8
+  sources: 1
+  verified: false
+  freshness: { as_of: 2026-06-29T00:00:00Z, state: fresh }
+  signals: { signed: true, corroborated: false, fresh: true }
+  assessed_at: 2026-06-29T00:00:00Z
+---
+
+# Value
+
+4.25% over the 12 months ending 2026-05 (BLS, signed envelope).
+```
+
+A conflict case, where independent sources disagree on the same reading:
+
+```markdown
+---
+type: Live Metric
+title: PM2.5, station 118
+reliability:
+  confidence: LOW
+  basis: live-source
+  sources: 2
+  verified: false
+  conflict:
+    disputed: true
+    positions:
+      - { statement: "PM2.5 = 42 ug/m3", basis: live-source, source: "Sensor network A, observed 2026-06-29T10:50Z" }
+      - { statement: "PM2.5 = 11 ug/m3", basis: vendor-doc, source: "Regulatory daily summary, prior-day average" }
+    resolution: "live-source prevails under the trust ordering; both positions retained"
+  validity: { keyed_by: date, valid_from: 2026-06-29, valid_until: null }
+  freshness: { as_of: 2026-06-29T10:50:00Z, state: fresh }
+  signals: { signed: true, corroborated: false, fresh: true }
+---
+```
+
+## 5. Validation (optional)
+
+Consistent with OKF's stance that it *references* schemas rather than subsuming them (SPEC, Non-goals), an optional JSON Schema (draft 2020-12) for the `reliability` object is published at:
+
+- Schema: <https://dynamicfeed.ai/schemas/okf-reliability-v1.json>
+- Worked example bundle: <https://dynamicfeed.ai/schemas/okf-reliability-examples.json>
+
+Producers MAY validate against it. Per SPEC §4.1, consumers MUST NOT reject a document solely for omitting `reliability` or for carrying keys beyond those defined here. The schema encodes the §3 honesty rules as constraints, and has been validated against three independent production shapes (a live-data layer, a multi-version corpus, and a multi-domain human-authored KB).
--- a/okf/proposals/reliability.md
+++ b/okf/proposals/reliability.md
@@ -24,7 +24,7 @@ This is the fourth trust axis, distinct from the three already in flight:
 ### 2.1 Floor
 
 - `confidence` (required within the object) — an ordinal band, one of `HIGH | MEDIUM | LOW | UNVERIFIED`. This is the interoperable surface consumers filter on.
-- `basis` — how the claim was obtained, one of `live-source | partner-attested | vendor-doc | forecast | computed | inferred`, ordered most to least authoritative when sources disagree.
+- `basis` — how the claim was obtained, one of `live-source | partner-attested | vendor-doc | forecast | computed | inferred`, ordered most to least authoritative when sources disagree. (Open question from the human-authored-KB shape: curated or decided knowledge is neither a live/computed/forecast reading nor an inference, so it has no honest value here; an `authored` value is under discussion on the proposal thread.)
 
 ### 2.2 Corroboration tier
 
@@ -107,4 +107,4 @@ Consistent with OKF's stance that it *references* schemas rather than subsuming 
 - Schema: <https://dynamicfeed.ai/schemas/okf-reliability-v1.json>
 - Worked example bundle: <https://dynamicfeed.ai/schemas/okf-reliability-examples.json>
 
-Producers MAY validate against it. Per SPEC §4.1, consumers MUST NOT reject a document solely for omitting `reliability` or for carrying keys beyond those defined here. The schema encodes the §3 honesty rules as constraints, and has been validated against three independent production shapes (a live-data layer, a multi-version corpus, and a multi-domain human-authored KB).
+Producers MAY validate against it. Per SPEC §4.1, consumers MUST NOT reject a document solely for omitting `reliability` or for carrying keys beyond those defined here. The schema encodes the §3 honesty rules as constraints, and a portable conformance-vector suite (12 cases across all five rules, 6 valid and 6 invalid) is published alongside it (<https://github.com/dynamicfeed/df-verify/tree/main/reliability>) so independent implementations can confirm they agree. Independent runs from a live-data layer and a multi-version corpus pass in full; a human-authored-KB shape passes the honesty rules and surfaced the open `basis` question noted in §2.1.


Dynamicfeedai:
Update for reviewers: the live reference implementation now emits this object, so the schema in this PR is not aspirational. You can curl it and validate the response against the file here.

```
curl -s "https://dynamicfeed.ai/v1/facts?tool=us_economy&indicator=cpi" | jq '.facts[0].reliability'
```

returns a `reliability` block on every fact, built from the same signals and conformant by construction:

- an ordinal `confidence` band with a corroboration ceiling: a single-source live reading caps at MEDIUM even at a high score, so the band is not a pure re-banding of `score`;
- HIGH only when verified and `score >= 0.5`, UNVERIFIED only when `score < 0.5` and never verified, `verified` only with `sources >= 2`;
- a disputed conflict always carries both positions and a resolution.

Validated green against the live tool set plus a conflict case, using the schema published at https://dynamicfeed.ai/schemas/okf-reliability-v1.json. Happy to keep it in lockstep with the object as it firms up across the three shapes.

K4uP:
This faithfully captures the #151/#158 convergence — and turning it into a SPEC-voice doc this fast is much appreciated. From the multi-version-corpus side (one of the three shapes), it maps cleanly onto real practice: the `confidence` + `basis` floor is what we filter on day to day; version-keyed `validity` is how we answer "was this true at the release the user is running"; and the `validity` (anonymous `valid_until`) vs cross-concept `supersedes` split (deferred to #158/#148) is the right seam — it keeps this object from swallowing the typed-edge work.

The maturity-ladder framing in §2 is the part I'd most want to survive review: a two-field floor with corroboration and maintenance as opt-in tiers is what keeps this a gentle on-ramp rather than a gate. A producer setting just `confidence: LOW / basis: inferred` on intake and growing into the rest is exactly how a corpus matures in practice.

One clarifying question on the conflict example: it carries `confidence: LOW` while `resolution` says live-source prevails under the trust ordering. Is the intent that *any* disputed claim floors at LOW regardless of which basis wins — or can the prevailing position carry its own band (e.g. a clean live-source win capped at MEDIUM per rule 4) with `disputed: true` still flagged? Both are defensible; worth pinning down so producers grade conflicts consistently. Either way, +1 to merging this as the agreed name and shape.

K4uP:
Ran the multi-version corpus shape through `okf-reliability-v1.json` with ajv (draft 2020-12) — green across the board, and the negative controls confirm the honesty rules actually bite:

| Case | Result |
|---|---|
| HIGH / live-source / version-keyed `validity` (4.0–5.2) / `lifecycle.history` (inferred→vendor-doc→live) | valid |
| Conflict (vendor-doc vs live, both positions + resolution), version `validity` | valid |
| Maturity-ladder floor (`confidence` + `basis` only) | valid |
| `verified:true` with `sources:1` | correctly rejected (`sources must be >= 2`) |
| `UNVERIFIED` + `verified:true` | correctly rejected |
| `HIGH` + `score:0.2` | correctly rejected (`score must be >= 0.5`) |
| disputed conflict + `verified:true` | correctly rejected |

So the multi-version shape is conformant as-is — version-keyed `validity` and the stored-ladder `lifecycle` both validate unchanged. The union `basis` enum is what makes it work: our `vendor-doc`/`inferred`/`partner-attested` and the live-data `live-source`/`forecast`/`computed` coexist cleanly.

One empirical point on the conflict-banding question I raised earlier: our conflict case validates at `confidence: MEDIUM`, your example #4 uses `LOW`, and **both pass** — so the band under a dispute is currently unconstrained. That's the one rule I'd pin so producers grade conflicts consistently. Two clean options:

1. **Disputed floors the band** — `disputed:true ⇒ confidence ∈ {LOW, UNVERIFIED}`. Simplest; treats any open disagreement as low-trust.
2. **Band = the prevailing position's grade, capped at MEDIUM**, with `disputed:true` still flagged. Preserves "the live-source reading is still the best answer" without hiding the dispute.

I lean (2) — it matches the "resolution records what the trust ordering picked" intent: the consumer gets both the dispute flag and a usable graded answer. Either works; the point is to pick one. Happy to open a PR adding it as a fifth honesty rule   a test case if useful.

Dynamicfeedai:
Thank you for running the multi-version corpus through the schema with ajv, and for the negative-control table. A second of the three shapes validating green, with the honesty rules actually rejecting the bad cases, is exactly the signal this needed. Your point about the union `basis` enum is the crux: `vendor-doc`/`inferred`/`partner-attested` coexisting with `live-source`/`forecast`/`computed` is what lets one object serve a versioned corpus and a live layer without forking.

On the conflict-banding question: you are right that it should be pinned, and I agree with your option (2). The cleanest framing is that it is rule 4's corroboration ceiling applied to a dispute: an open disagreement is a corroboration failure, so the claim cannot be `HIGH`, but the position the trust ordering selects can still carry up to `MEDIUM`, with `disputed: true` always flagged. A producer that wants to be conservative can floor to `LOW`. That keeps both the dispute flag and a usable graded answer, which matches the intent that `resolution` records what the ordering picked rather than discarding the loser.

I have landed it as honesty rule 5:

- proposal §3 (this PR);
- the schema `allOf`, so ajv now rejects a disputed `HIGH`;
- both reference validators, with a good case (disputed + `MEDIUM` clean win) and a bad case (disputed + `HIGH`).

Our live producer already floors disputes to `LOW`, so it stays conformant; the rule mainly closes the gap for other producers. Your corpus's `MEDIUM` conflict case and example #4's `LOW` both still validate, which is the point: the ceiling is `MEDIUM`, not a forced floor.

A conformance test vector for the disputed-band cap from the corpus side would be a welcome addition if you would like to contribute it; happy to fold it into the validators. And +1 to your read that the §2 maturity-ladder framing is the part to protect in review: a two-field floor that grows into the corroboration and maintenance tiers is the on-ramp, not a gate.

K4uP:
Landed perfectly — rule 5 reads exactly right, and framing it as rule 4's corroboration ceiling applied to a dispute is cleaner than my either/or. Re-ran against the updated schema: our `MEDIUM` conflict still validates and a `HIGH` dispute is now correctly rejected (`confidence must be equal to one of the allowed values`), so the cap bites.

Here's the corpus-side conformance vector you asked for — a version-keyed disputed case at the `MEDIUM` ceiling (good) and its `HIGH` twin (bad). It complements your PM2.5 example, which sits at the conservative `LOW` floor: together they pin both ends of rule 5 — a producer MAY floor to `LOW`, MAY ride the prevailing position up to `MEDIUM`, and MUST NOT reach `HIGH`.

```json
[
  {
    "label": "rule5 GOOD — versioned corpus, disputed, prevailing live-source win at the MEDIUM ceiling",
    "expect": "valid",
    "reliability": {
      "confidence": "MEDIUM", "basis": "live-source", "sources": 2, "verified": false,
      "conflict": {
        "disputed": true,
        "positions": [
          { "statement": "Default session timeout is 30 minutes.", "basis": "live-source", "source": "Observed on an Acme 5.0 instance, 2026-06-29" },
          { "statement": "Default session timeout is 20 minutes.", "basis": "vendor-doc", "source": "Acme Admin Guide 5.0" }
        ],
        "resolution": "live-source prevails under the trust ordering; both positions retained"
      },
      "validity": { "keyed_by": "version", "valid_from": "5.0", "valid_until": null },
      "freshness": { "as_of": "2026-06-29T00:00:00Z", "state": "fresh" },
      "signals": { "signed": true, "corroborated": false, "fresh": true }
    }
  },
  {
    "label": "rule5 BAD — same dispute graded HIGH (must be rejected)",
    "expect": "invalid",
    "reliability": {
      "confidence": "HIGH", "basis": "live-source", "sources": 2, "verified": false,
      "conflict": {
        "disputed": true,
        "positions": [
          { "statement": "Default session timeout is 30 minutes.", "basis": "live-source", "source": "Observed on an Acme 5.0 instance" },
          { "statement": "Default session timeout is 20 minutes.", "basis": "vendor-doc", "source": "Acme Admin Guide 5.0" }
        ],
        "resolution": "live-source prevails; both positions retained"
      },
      "validity": { "keyed_by": "version", "valid_from": "5.0", "valid_until": null }
    }
  }
]
```

Both checked against the updated `okf-reliability-v1.json` with ajv: GOOD validates, BAD rejected on `confidence`. Fully synthetic (made-up "Acme" platform/values), so safe to ship — fold them in wherever useful. Happy to send them as a PR to the proposal's §4 examples too, if you'd rather have a versioned-corpus case in the doc alongside the live ones.

Dynamicfeedai:
That is the cleanest possible confirmation: a `HIGH` dispute rejected on `confidence` while your `MEDIUM` case stays green means the cap bites from the corpus side too, not just ours.

Your two version-keyed vectors are exactly the pair that pins both ends, so I folded them into a portable conformance suite in the reference repo (credited to the multi-version-corpus shape, synthetic values noted): https://github.com/dynamicfeed/df-verify/blob/main/reliability/conformance-vectors.json . It now carries 12 `{label, expect, reliability}` vectors across all five rules, 6 valid and 6 invalid, each checked to match its expect against both the JSON Schema (ajv) and the reference validator. Your `MEDIUM` ceiling sits right beside a PM2.5-style `LOW` floor, so a producer can read the full rule-5 range off one file: MAY floor to `LOW`, MAY ride the prevailing position to `MEDIUM`, MUST NOT reach `HIGH`.

Yes please send the §4 PR adding a versioned-corpus disputed case to the proposal alongside the live ones. A version-keyed example in the doc itself is exactly what shows the object serves a corpus and not just a live layer, and it pairs naturally with the conformance vector. Happy to land it.

inkxel:
@Dynamicfeedai @K4uP — the SPEC-voice doc and the ajv cross-check are good to see, and the maturity-ladder floor holding up under the negative controls is the part worth protecting through review.

Two things from the authored-corpus side — the third shape.

First, corroboration. I built `reliability` emission into Throughline (the human-authored-KB shape) and ran your `df-verify` 12-vector suite through an independent validator against the current schema — 12/12 match their expects, rule 5 included (the HIGH dispute correctly rejected). So a third validator agrees the honesty rules bite, alongside the live-data and corpus checks.

Second, the gap that shape surfaces. Where the live-data and versioned-corpus shapes clear the floor, a human-authored claim doesn't — and it's `basis`. A curated wiki fact or an architecture decision isn't `live-source`, `computed`, or `forecast`, and it isn't `inferred` either (a person *deciding* or *observing* something isn't an inference). With nothing accurate to put there, `basis` stays empty and the floor fails. That lands on the union point you just settled: `vendor-doc`/`inferred`/`partner-attested` + `live-source`/`forecast`/`computed` serves a corpus and a live layer, but there's no value for the third shape. My real question: how did the human-authored KB in your validation set `basis`? If it used `inferred`, that's the gap — it files curated knowledge under inference, and anyone filtering on `basis` reads it wrong.

Proposal: add an authored value — `authored` or `curated` — so a human-authored corpus clears the floor without misreporting how the claim was obtained. Emission + the conformance run are on a branch if useful: https://github.com/inkxel/throughline/tree/okf-reliability-conformance. Glad to be the third shape's actual vector rather than an asserted one.

On the cross-concept `supersedes`/`contradicts` work: agreed it's out of scope here and lives in #158/#148 — I've built it as a typed link in Throughline and will take the proposal there.

(The `EXTENSIONS.md` registry still feels like the cleanest home for coordinating these, but no strong opinion here.)

Dynamicfeedai:
Straight answer to your question first, because it is the right one to ask: the human-authored KB was a shape the object is *designed* to serve, carried over from the #151 discussion, not something we ran in production. Our live-data layer is the only shape we run; @K4uP's ajv run is the corpus shape. So your Throughline run is the first real third-shape conformance pass, not a re-run of ours, and 12/12 with the HIGH dispute rejected is exactly the independent check that was missing. I have corrected §5 to state what actually passes rather than assert three production shapes, and flagged your gap as an open question in §2.1. Thank you for catching the overclaim.

On the gap itself you are right, and it is real. A curated fact or an architecture decision is a person authoring, deciding, or observing, which is not an inference; filing it under `inferred` mis-signals to anyone filtering on `basis`. `partner-attested` is the nearest existing value and still wrong, since a partner attesting is not the author deciding. So the closed union genuinely has no honest slot for the authored shape, and that is a floor failure, not a style nit.

Proposal: add `authored` to the `basis` enum, for a claim a person created, decided, or observed directly (a curated KB entry, an ADR, a human observation). I lean `authored` over `curated` because it also covers deciding and observing, where curated reads as only vetting; if Throughline's content is mostly vetted-secondary, curated may fit better, or both earn a place. On the ordering (most to least authoritative when sources disagree) a primary human source sits high, likely just under `live-source` and at or above `partner-attested`: `live-source > authored > partner-attested > vendor-doc > forecast > computed > inferred`. You and @K4uP know your shapes' authority best, so I would rather confirm the term and the slot with you both than set it unilaterally, since the union is closed by design and a new term is precisely the cross-shape-agreement case.

If `authored` and that ordering work for you both, I will land it in the schema, both validators, and the conformance vectors, with an authored-basis vector so the third shape clears the floor honestly in the suite. And yes, please wire Throughline's branch in as the third shape's real vector; an authored-corpus case beside the live and corpus ones is what makes the suite cover all three. (Agreed that supersedes/contradicts belongs in #158/#148, and the EXTENSIONS.md registry as the coordination home is a good instinct.)

inkxel:
@Dynamicfeedai appreciate you correcting §5 and moving the gap into §2.1 — and no need to thank me for the catch; that's the point of a third shape kicking the tires.

On the term: `authored` works, and I'd weight it as the must-have. It's the right floor-clearer for the bulk of what I run — first-person decisions, observations, and facts a person wrote down directly.

One signal you asked for, since it bears on `curated`: my corpus actually has two sub-shapes. The primary layer is authored — decisions, observations, hand-written claims. But a good chunk is *synthesized* — wiki articles compiled and vetted from raw sources (transcripts, notes), which is closer to your "vetted-secondary" read. So `curated` isn't redundant for me; it's a distinct provenance — selecting/synthesizing existing material vs. creating it. I'd still land `authored` first as the floor-clearer and let `curated` follow only if it earns its keep across shapes, rather than pre-loading a closed enum — but for me, both have real content behind them.

On the ordering: `live-source > authored > partner-attested > vendor-doc > forecast > computed > inferred` reads right. The one thing I'd name: authored authority is strongest for *first-party* claims — my own decision or observation, where I am the source — so sitting just under `live-source` and above `partner-attested` fits that case. If `curated` ever lands, I'd put it below `authored`, since synthesis is more derived — somewhere around `vendor-doc`/`computed`.

And yes — I'll wire Throughline's branch in as the third shape's real vector. Once `authored` is in the schema I'll add an authored-basis conformance case to sit beside the live and corpus ones, so the suite actually covers all three. Happy to confirm the final term + slot with you and @K4uP before it lands.

Dynamicfeedai:
That two-sub-shape signal is exactly what a closed-union decision should turn on, so thank you for it. Agreed on the plan:

- Land `authored` first as the floor-clearer, for first-party claims a person decided, observed, or wrote directly.
- Ordering `live-source > authored > partner-attested > vendor-doc > forecast > computed > inferred`. Your reasoning for the slot is right: authored authority is strongest first-party, so it sits just under `live-source` and above `partner-attested`.
- Hold `curated` until it earns its keep across more than one shape. Your synthesized layer (wiki articles compiled and vetted from raw sources) is a real, distinct provenance, not a duplicate of `authored`, and if a second shape needs the same distinction it has a clear home below `authored`, around `vendor-doc`/`computed` where derived-by-synthesis sits. Not pre-loading a closed enum on one shape's need is the right discipline.

I will land `authored` across the schema, both validators, the conformance vectors, and §2.1 of the proposal as soon as @K4uP gives a quick +1, since the multi-version-corpus shape is the adjacent one in the ordering and you both asked to confirm together. Then I will add an authored-basis vector beside the live and corpus cases, and wire your Throughline branch in as the third shape's real vector.

@K4uP, does `authored` at that slot work for the corpus shape? If it does, I will bump the closed union and the conformance vectors in one change.

Sliky1:
I think the `reliability` object in this PR is the right base. The complementary piece from the maturity-model perspective (#160) is a production promotion model — not just how confident a claim is, but how it advances from extracted to trusted.

In our pipeline, that's a 5-stage flow (extract → denoise → backfill → regress → promote), gated by regression thresholds and evidence coverage, not by confidence alone. `reliability` answers "how much to believe this claim"; `lifecycle` + `verification_gate` answer "what state is this claim in and how did it get here."

Concretely, a minimal extension could look like:

```yaml
lifecycle: active              # proposed | active | expired
verification_gate:
  promoted_by: regress
  promoted_at: 2026-06-29T10:00:00Z
  thresholds: { node_recall: 0.95, edge_recall: 0.75 }
evidence:
  - document: tech/some-doc.md
    quote: "System X uses technology Y"
    confidence: 0.95
```
The rationale for putting it here rather than a separate track: #151-style reliability and lifecycle gating solve adjacent sides of the same problem — one grades trust, the other governs promotion and audit. They compose; neither subsumes the other.

Happy to contribute a worked example bundle from real production data if it helps the discussion.

K4uP:
@Dynamicfeedai @inkxel — authored at that slot works for the multi-version-corpus shape. +1 to landing it.

Two reasons from the corpus side, so the term isn't resting on the authored shape alone:

It's a floor-failure for the corpus too, not just the authored KB. A versioned corpus isn't only live-verified readings and vendor docs — a real share of it is first-party authored: a consultant's recorded observation of how a system actually behaves, an  architecture decision, a lessons-learned entry. None of those is live-source/computed/forecast, and none is inferred — a person observing or deciding something hasn't inferred it. Today they have nowhere honest to sit, so they default to inferred, and anyone filtering on basis reads recorded first-hand knowledge as a guess. So @inkxel's gap reproduces on a second, independent shape — authored is the honest slot for both, not a one-corpus accommodation.

The slot is right. live-source > authored > partner-attested > vendor-doc > forecast > computed > inferred matches how a corpus actually resolves disagreement. A first-party recorded observation outranks the vendor's own doc — when the running system contradicts the admin guide, the observation wins (that's exactly the disputed case we already pin) — but a fresh live reading still beats a recorded one, so it sits just under live-source. And the author deciding outranks a partner attesting, so above partner-attested is right.

On curated: agree to hold it, and flagging that the distinction isn't single-shape either. My corpus has the same two sub-shapes @inkxel described — first-party authored notes vs. roll-up notes that synthesize and vet from several underlying sources (the "vetted-secondary" read). So curated has real content on a second shape if/when it earns its way in, and below authored around vendor-doc/computed is where derived-by-synthesis belongs. But authored first is the right discipline — don't pre-load a closed enum on a need only two shapes have so far named.

So: bump the union, I'm a +1 on the term and the slot. Once authored is in the schema I'll re-run the corpus vectors with it and send the §4 PR you greenlit — a versioned-corpus disputed case (the MEDIUM-ceiling one) plus an authored-basis example — pre-validated against the then-current schema, so the proposal shows the corpus shape clearing the floor honestly beside the live ones.

K4uP:
Quick status check on `authored`, since §2.1 flags it as the open question and the thread's gone quiet since the +1s — no rush intended, just keeping the state visible, and happy to take the work off your plate.

Two data points since 30 June that bear on it:

1. **A fourth shape adopted the axis in production — and it shows the cost of an unlanded enum.** @chirag127's 828-concept bundle (see the comment on #151) shipped an optional `confidence` frontmatter field on 2026-07-03, citing this axis — the first adoption *after* the proposal rather than convergent with it. But it ships a dialect: three lowercase bands, with `UNVERIFIED` expressed by omitting the field. Every week the enum sits in proposal limbo, more dialects ship.

2. **`authored` is no longer hypothetical on the corpus side either.** We've since landed it in our production corpus: an `AUTHORED` evidence grade for first-party consultant observations, architecture decisions, and lessons-learned entries, enforced as a **blocking check** in our consistency gate, with the first notes carrying it. Exactly the class of claim the enum value was named for.

Offer, to make this a review rather than a task: I'll draft the schema delta (`authored` into the `basis` enum plus the one-line ordering note) and two conformance vectors (a valid `authored` case and an invalid probe of the slot), pre-validated with ajv — you land them whenever it suits. That would also unblock the §4 worked-example PR we discussed.

(Also: `vantage` is a clean addition — independent-vs-producer-reported is a genuinely different question from `basis`, and it slots into the axis map without colliding with anything here.)

K4uP:
Following up on the offer from 4 July — here is the `authored` delta, ready to apply, so that landing it is a five-minute edit rather than an open task. To be clear about roles: this PR and the hosted schema are yours, and whether and how to land this stays entirely your call (and the maintainers'). We're sharing it because demand for the value keeps arriving faster than the enum can wait — and we'd rather hand over a finished diff than watch the gap do damage.

**Why the urgency, briefly.** #182 (the Lexenne producer report) ships `provenance: observed | asserted | inferred` and calls it *"the key with no home"*. Mapped against this object: `observed` ≈ `live-source`, `inferred` = `inferred` — and `asserted` is precisely the `authored` value this thread agreed on in late June. A fifth production system has just independently re-derived the missing value, and because it isn't in the enum, a parallel key (`epistemic_origin`) is now on the table. Two fields carrying the same concept is how a young standard forks: every future producer must choose between them, every consumer must read both, and the choice itself carries no information. We'd much rather add the one missing value to the one existing field. (Mapping details posted on #182.)

**Schema delta** (against `okf-reliability-v1.json` as hosted today) — one enum insertion, one word in the ordering sentence, nothing else:

```diff
 "basis": {
-  "description": "... ordered most to least authoritative when sources disagree: live-source > partner-attested > vendor-doc > forecast > computed > inferred. ...",
+  "description": "... ordered most to least authoritative when sources disagree: live-source > authored > partner-attested > vendor-doc > forecast > computed > inferred. ...",
   "enum": [
     "live-source",
+    "authored",
     "partner-attested",
     "vendor-doc",
     "forecast",
     "computed",
     "inferred"
   ]
 }
```

`authored` = a first-party claim a person (or accountable agent) observed, decided, or wrote down directly — an architecture decision record, a consultant's recorded observation, a lessons-learned entry. Slot exactly as agreed on 29–30 June: strongest first-party grade, directly under `live-source`, above `partner-attested`. `curated` stays deferred, as discussed.

**Conformance vectors**, run with ajv (draft 2020-12) against both the currently hosted schema and the patched one:

| # | Vector | Hosted schema | With delta | What it proves |
|---|---|---|---|---|
| V1 | valid `authored`: consultant observation, `MEDIUM`, 1 source, version-keyed `validity`, intake→authored `lifecycle` | ❌ rejected (`basis` unknown — including in `lifecycle.history`, via the `$ref`) | ✅ valid | the gap itself |
| V2 | `basis: asserted` (the #182 spelling) | ❌ | ❌ | the closed union holds — the fix is mapping, not minting |
| V3 | `authored` + `verified: true` + `sources: 1` | ❌ | ❌ | honesty rule 2 still bites on the new value |

<details>
<summary>The three vectors (JSON, synthetic values)</summary>

```json
[
  {
    "label": "V1 valid: authored consultant observation",
    "reliability": {
      "confidence": "MEDIUM",
      "basis": "authored",
      "sources": 1,
      "verified": false,
      "validity": { "keyed_by": "version", "valid_from": "5.0", "valid_until": null },
      "assessed_at": "2026-07-10T09:00:00Z",
      "lifecycle": {
        "stage": "authored-reviewed",
        "history": [
          { "at": "2026-07-01", "confidence": "LOW", "basis": "inferred", "note": "import intake, unreviewed" },
          { "at": "2026-07-10", "confidence": "MEDIUM", "basis": "authored", "note": "author identified; first-party observation record" }
        ]
      },
      "signals": { "signed": false, "corroborated": false, "fresh": true }
    }
  },
  {
    "label": "V2 invalid: basis 'asserted' must be rejected",
    "reliability": { "confidence": "MEDIUM", "basis": "asserted", "sources": 1, "verified": false }
  },
  {
    "label": "V3 invalid: authored + verified:true at sources:1",
    "reliability": { "confidence": "MEDIUM", "basis": "authored", "sources": 1, "verified": true }
  }
]
```

</details>

Side note on evidence: since the last update we also run `authored` in production ourselves — our corpus gate enforces the grade as a blocking check, and the first notes carry it. So the value is now production-real in at least three places (our corpus, Throughline's authored layer, and #182's `asserted`).

Once the enum lands, I'll follow up with the promised §4 worked example for `reliability.md` (the version-keyed disputed-`MEDIUM` case plus V1 above) as a proper PR. And if you'd rather reshape any of this first — different wording, different vector set — happy to adjust. Thanks again for carrying this proposal as far as you have; none of this convergence would exist without the PR.

andrewcrenshaw:
Registering the demand-side +1 for this delta from #182: our exporter's `asserted` is precisely this `authored` value, and V2 rejecting the old spelling is the behavior we want - mapping, not minting. We are dropping `epistemic_origin` as a candidate second field for the same reason (details on #182).

inkxel:
@K4uP landed `authored` on my end and ran your V1/V2/V3 plus a real export through a different validator than the `ajv` runs already in this thread — Python's `jsonschema`, against the hosted schema and a locally-patched copy with your delta. All four came back exactly as expected, honesty checks included.

Nothing here contradicts your conclusions, just adding another test to the pile. Nice that a value four shapes have converged on now has a second validator agreeing too.

K4uP:
Two quick thanks and a one-line status, then I'll get out of the way.

@inkxel — a second independent validator (Python `jsonschema`) on top of the ajv runs, green across V1/V2/V3 plus a real export, is exactly the confirmation this needed. A value four shapes converged on now has two validator implementations agreeing.

@andrewcrenshaw — withdrawing `epistemic_origin` in favour of mapping onto `basis` is the "one concept, one field" outcome in practice, not just in principle. Grateful you chose convergence over a parallel key; that's what keeps the axis from forking.

Where that leaves us: the object is agreed, doubly validated, and adopted by two producers locally — the only remaining step is `authored` landing in the hosted schema. Whenever that's in, I'll follow up with the promised §4 worked example as a PR against `reliability.md`, and I'm happy to help however's most useful in the meantime. Thanks again for carrying this.

K4uP:
@mindmatter-aia @andrewcrenshaw — bringing the `assessed_at` vs `last_confirmed` question over from #182, since it's this proposal's to answer.

First, thank you for the report itself: a curator-contract producer with a write boundary and a release gate is a shape this thread hadn't seen yet, and asking "can I map to an existing field?" before minting one is exactly the instinct that has kept this axis in one piece.

**The answer: no — `assessed_at` is not the home for `last_confirmed`, and mapping onto it would lose the very signal you're keeping.** They are different clocks. §2.3 already runs three deliberately: `freshness.as_of` (when the reading was taken), `validity` (when the claim applies), `assessed_at` (when the grade was computed). Your confirmation is a fourth act, not a synonym for any of them — @andrewcrenshaw's bi-temporal lens says it exactly: a human re-checking "still true", with no re-measurement, is a transaction-time event, not a new valuation. `assessed_at` answers *"when was this graded?"*; `last_confirmed` answers *"when did a human last stand behind it?"*

The "would lose the signal" part is mechanical, not aesthetic: §2.3 explicitly allows a recompute-per-read producer to refresh `assessed_at` on every read. Fold the confirmation in, and the next read silently overwrites "a human checked this three weeks ago" with "the machine re-graded it just now" — exactly the datum your decay-lint runs on.

This may look like it cuts against "one concept, one field"; it's the mirror case. `epistemic_origin` was two names for one concept — merge. This is one name for two concepts — split. #182 just priced that failure mode for us: `provenance` carrying three unrelated senses across three producers is precisely why "the key with no home" was hard to place. So the honest move is to name the act, not overload `assessed_at`.

Concretely, I'd add one optional maintenance field to §2.3 — `confirmed_at`, for symmetry with `assessed_at` (adopting your `last_confirmed` spelling would be equally fine; anything but a third name) — with one guardrail carried over from §3: **a confirmation is a recency marker only; it MUST NOT by itself raise `confidence` or set `verified`.** That's rule 1's discipline (signed ≠ verified) and #195's mechanical cap, applied here: affirmation is load-bearing for recency, but a field an agent can type is never a grade. Your lint then lands exactly parallel to `freshness.state` going stale — a maintenance defect, not a grade change.

@Dynamicfeedai — happy to supply the §2.3 delta plus conformance vectors in the same form as the `authored` set, if wanted.

And a wider thank-you, overdue: the production reports this axis keeps attracting — *remember*, Throughline, Data Olympus, and now yours — have sharpened the proposal at every step, and several of the findings have flowed straight back into my own KB practice. This kind of review is exactly what the axis needed.

Dynamicfeedai:
Yes, please—`confirmed_at` is the cleaner name for us, particularly because it sits naturally beside `assessed_at`.

We agree that it represents a fourth clock and must remain distinct from:

- `freshness.as_of`;
- `validity`;
- `assessed_at`.

The guardrail is also exactly right: `confirmed_at` is a recency and maintenance signal only. It must not, by itself, increase `confidence`, alter `score`, or set `verified`.

Once the §2.3 delta lands, Dynamic Feed can keep the hosted schema and live reference output in lockstep and run the vectors through both our JavaScript and Python validators.

The conformance cases I would especially preserve are:

- recomputing `assessed_at` does not overwrite `confirmed_at`;
- adding `confirmed_at` does not change the existing grade;
- `confirmed_at` alone cannot satisfy verification or corroboration rules;
- an overdue confirmation becomes a maintenance defect rather than an automatic confidence downgrade;
- omission remains valid for producers that do not run a human confirmation process.

Happy to review the delta and vectors when ready. This seems like a useful case where “one concept, one field” requires a new field rather than overloading an existing one.

andrewcrenshaw:
The `confirmed_at` split reads right, and thank you both - the "a human standing behind a claim is a different clock from grading it" framing is one I'll carry back into our own model.

Worth naming what just converged, because it's now bigger than this one field. The same discipline is doing load-bearing work in three places at once: a `confirmed_at` that marks recency but never raises the grade (here), a `supersedes` edge that displaces nothing until an affirmation event is appended (#195/#148), and @mindmatter-aia's release gate where the human flip is what lets content leave. One principle underneath all three - affirmation is load-bearing for state, but a field an agent can self-assert is never itself a grade or a gate. `confirmed_at` with the recency-only guardrail is that principle applied to freshness.

+1 on @Dynamicfeedai's conformance cases - the "recomputing assessed_at does not overwrite confirmed_at" one especially, since that overwrite is the whole reason it can't fold into assessed_at.

andrewcrenshaw:
v0.2 (now merged in #227) changes what this proposal composes with, and the validity half comes through untouched: the spec now carries three clocks, and none of them is valid-time - when the fact itself holds true in the world, as opposed to when the record was written or checked.

The three: `generated.at` records when the content last changed (section 5.2), `verified[].at` records when someone confirmed it, and `stale_after` (section 5.5) records when re-verification is due. All three are about the record and its upkeep. None says when the fact holds true in the world - a fiscal-year revenue figure is valid for that fiscal year regardless of when it was generated, confirmed, or re-checked. The `validity.valid_from` / `valid_until` pair proposed here is still the only home for that clock. This is the transaction-time vs valid-time distinction from the #182 producer report: transaction time landed in core as `generated.at`; valid time did not land.

The confidence half now needs one composition note. Section 5.3 derives trust tiers from who verified (actor class) and not from when - the same rule proposed upthread for `confirmed_at` (a recency marker that never raises the grade), now present in core. `reliability.confidence` slots in beside that rather than under it: `verified` records who stands behind a statement now; `confidence` records how sure the producer was when it made the statement. A human-reviewed concept can carry low producer confidence (a reviewed hypothesis stays a hypothesis), and an unverified concept can carry high confidence. Worth a sentence in the proposal text so consumers don't collapse the two into one score.

From our side: the *remember* exporter (the Lexenne producer behind #182 and PR #195) is being upgraded to emit the v0.2 core family, and we will emit `reliability.confidence` and `validity.*` as extension keys using this proposal's spellings alongside it - one live producer emitting the core trust family and this axis side by side. Happy to update the #182 mapping table onto v0.2 once this PR rebases.

K4uP:
v0.2 (#227) landed while this thread was mid-flight, so here is an honest rebase ledger: what core absorbed from our open items, what it leaves to this proposal, and what a rebase has to touch. First, congratulations to the maintainers — §5.1's stance that credibility is *inferred from signals, not stored as a score* is the same discipline this thread's §3 honesty rules run on. The families compose well, and several things we converged on here now exist in core under different spellings. That is the good outcome; the spelling was never the point.

**Absorbed by core — two open items retired.**

**1. `authored` — retired; map, don't mint.** The gap @inkxel surfaced (self-authored knowledge with no honest `basis` value) now has a core home: §7 requires the `human:<id>` actor form for hand-authored or human-confirmed content, and §5.3 derives the human-reviewed tier from it. A curated concept says `generated: { by: human:<id> }` and simply omits `basis` — `confidence` is the only required key inside the object, and `basis` keeps its narrower meaning of how a claim was *obtained* (live reading vs. vendor doc vs. inference), which is orthogonal to who wrote the document. So I'm withdrawing the pending `authored` enum delta and its schema patch in favour of composition with §7 — the same "one concept, one field" rule we applied to `epistemic_origin`, now applied to our own pending key. @andrewcrenshaw, your exporter's `asserted` values map to `generated.by: human:…` / `process:…` in core rather than to a new enum member.

**2. `confirmed_at` — retired; core `verified[].at` is the fourth clock.** §5.2 states it directly: *"facts can be re-confirmed without regeneration."* A human re-affirmation is an appended `{ by: human:<id>, at: … }` event — and the overwrite hazard that made `assessed_at` the wrong home is gone by construction, because a recompute refreshes `assessed_at` and never touches the `verified` list. The guardrail also arrived in core stronger than I proposed it: §5.3 derives the tier from actor class only, never from recency, so a confirmation timestamp cannot raise a grade. @mindmatter-aia — `last_confirmed` maps to `verified: [{ by: human:<id>, at: … }]`, and your decay-lint reads the latest `at`, composing with `stale_after` for the due date. @Dynamicfeedai — the five conformance cases you listed all hold structurally in this mapping (separate list, no overwrite, no grade coupling, omission valid). No §2.3 delta needed; I'm glad we never merged it into `assessed_at`.

**Not absorbed — what this proposal still carries.** @andrewcrenshaw's read from 24 July is exactly right on both halves, so just consolidating it into the doc's terms:

- **Valid-time.** `generated.at`, `verified[].at`, and `stale_after` are all record-and-upkeep clocks; none says when the fact holds true *in the world*. §2.3's `validity` remains the only home for that, and it is the one clock v0.2 did not add.
- **Graded confidence.** §5.3 grades *who stands behind the statement now*; `confidence` grades *how sure the producer was when it made it*. They must not collapse into one score: a human-reviewed hypothesis is still a hypothesis (tier high, confidence LOW), and a machine-only live reading can carry solid producer confidence. That composition sentence belongs in the proposal text — happy to draft it.
- **`conflict`.** Core has no disagreement state. A v0.2 concept whose two `sources` entries contradict each other has nowhere to say so; §2.3's `conflict` with rule 5's MEDIUM cap is still the only first-class treatment.

**What the rebase touches** (mechanical; offering to draft all of it, owner's call):

- The §4 examples carry v0.1 `timestamp:` — superseded by `generated: { by, at }` per §13.1.
- Two inner keys now share a name with core families of a different shape: `reliability.sources` (integer count) vs. core `sources` (entry list), and `reliability.verified` (boolean) vs. core `verified` (event list). Nothing collides in YAML — they're nested — but consumers will conflate them. Cleanest line I can see is to *derive rather than duplicate*: when core `sources` is present, the corroboration count SHOULD equal the number of independent core entries, and rule 2 can key off core `verified` plus corroboration instead of a parallel boolean. Renaming the inner keys is the fallback if derivation feels too clever.
- A short composition note carrying the two points above (confidence-beside-tier, validity as the missing clock).

If useful, I'll draft the rebased §2/§4 text against v0.2 and refresh the vector suite to match (retire the `authored` vector, add valid-time and confidence-vs-tier composition cases). And a firm +1 to the *remember* plan of emitting the core trust family and this axis side by side — one live producer doing both is worth more than anything we can write here.

andrewcrenshaw:
@K4uP both retirements look right from here, and your mapping for our values found a real defect in our emitter - thank you for aiming it at us. The wiki compiler was stamping every record `generated: { by: "remember/0.2" }`, the process-actor form, regardless of who authored the content - which under section 5.3 exported our human-authored lessons at the machine-generated tier. That is fixed and shipped: `asserted` lessons now emit the human actor form with a stable pseudonymous id, `observed` and `inferred` keep the process form, and `basis` keeps its how-obtained meaning unchanged. From the regenerated sample bundle (https://github.com/andrewcrenshaw/remember-okf-sample-bundle):

```
# claims/asserted-fixture-lesson... (human-authored)
generated:
  by: human:b26589eb
  at: '2026-08-01T00:00:00.000Z'

# facts/okf-v0-2-conformance-fixture-lesson... (machine reading)
generated:
  by: remember/0.2
  at: '2026-07-24T12:00:00.000Z'
```

Your confidence-beside-tier composition sentence has a public example in that same bundle: the machine-read fixture carries the process actor (lowest tier) and `confidence: established`. The two signals point in different directions on the same record and both are correct - a machine took a solid reading. Any scheme that merges them into one score has no honest value for that record.

On the shared inner names, one producer datapoint: our emitted reliability object carries only `confidence`, so the count-versus-list divergence cannot arise from our output - which is an argument for your derive-not-duplicate option, since a stored count that can disagree with the list it summarises will eventually mislead a consumer.

Validity as the one clock v0.2 did not add matches our reading, and the side-by-side you endorse is in the published bundle - core `generated` / `verified` / `status` beside `validity` and `reliability.confidence` on the same records, now with both actor classes represented. +1 to you drafting the rebased text; your ledger above is most of it.

inkxel:
@K4uP — the rebase ledger is the right call and I'm a +1 on both retirements, including my own.

**On `authored`: withdraw it, and I'll say why the composition is actually better than the enum.** §7's `human:<id>` plus §5.3's derived tier carries what I was reaching for, and it does it without pre-loading a closed enum on a need only two shapes had named. `basis` keeping its narrow "how was this obtained" meaning is the right outcome — the thing I was trying to encode was never *how* the claim was obtained, it was *who stood behind it*, and that has a home now. One concept, one field, applied to my own pending key. No argument.

Two follow-ups, both small.

**1. §2.1 of the proposal doc still carries the open question as live.** The line reads that curated knowledge "has no honest value here; an `authored` value is under discussion on the proposal thread." Since it's withdrawn, that sentence should go in the rebase — happy for you to strike it, or I'll send it as a one-line PR against your branch, whichever is less friction.

**2. One composition note I'd add, because omission is now load-bearing.** With `authored` retired, the guidance for hand-authored and for synthesized content is both "omit `basis`." That's correct, and it means an absent `basis` now has two distinct meanings: *not applicable to how this was obtained*, and *producer didn't populate it*. A consumer filtering on `basis` can't tell those apart, and the six-value enum gives it no way to ask.

This is the same shape as #207's dangling-link problem — an absence with two histories, where the spec's tolerance is right but the interpretation is underdetermined. Not proposing a value for it here; it's #207's class of question and probably belongs there. But the proposal text should say plainly that omitting `basis` is valid and carries no negative claim, so consumers don't read absence as a low signal. One sentence, and it prevents the reading where curated knowledge scores worse than a vendor doc for having honestly declined to guess.

**On the vector suite:** yes, retire the `authored` vector. I'll remap Throughline's conformance branch onto the §7 composition — `generated: { by: human:<id> }` with `basis` omitted for the first-party layer, and the agent actor form for the compiled layer — and re-run against the current schema so the third shape stays a real vector rather than an asserted one. I'll post the result rather than the intention.

**On `curated`:** holding it, per the earlier agreement, and I'd now argue the composition weakens the case for it rather than strengthening it. My two sub-shapes are distinguishable through `generated.by`'s actor form alone — `human:<id>` for first-party, the agent form for the compiled layer — which was the actual information I wanted. It doesn't need an enum member.

andrewcrenshaw:
@inkxel your omission note found a live instance the same day you wrote it. The bundle we pushed yesterday carried `basis: asserted` on the new human-authored record - a value the enum does not hold - because our emitter was defaulting the field from an internal type instead of omitting it. Nobody consuming the bundle flagged it; we caught it re-reading our own output against your point 2.

Fixed and republished. The asserted record now carries `generated.by: human:b26589eb` and no `basis` key; the machine-read record maps to `basis: live-source`. Same bundle URL, one commit: https://github.com/andrewcrenshaw/remember-okf-sample-bundle

So a producer datapoint for the one-sentence guidance you proposed: within hours of `authored` retiring, the first producer to touch the new composition emitted an out-of-enum value rather than an omission, precisely because nothing in the text said omitting is the correct positive act. "Omitting `basis` is valid and carries no negative claim" would have prevented it - the sentence earns its place.

mindmatter-aia:
Thanks @K4uP, this maps cleanly on our side and it puts the guardrail exactly where I wanted it. One confirmation, and one item I want to flag so nobody mistakes the field.

The recency and decay part needs no new field. `last_confirmed` is just the latest `verified[].at`, and our decay lint reads that latest `at` against `stale_after`. The authorship tier lands the same way: `human` becomes `generated.by: human:<id>`, `llm-generated` is the process-actor form, and `llm-reviewed` is a `human:<id>` entry appended to `verified[]`, with §5.3 deriving the tier. That is the observation I was conveying. The timestamp says who stood behind the claim and when, and it never moves the grade.

The one thing that does not reduce to a timestamp is our release gate. In our system the human flip is not just a recency mark, it is an authorization step: it changes what the content is allowed to do, which is leave the system. `verified[].at` captures the affirmation event, who and when. It does not carry that "may release" effect, and I don't think it should. That is the same line @andrewcrenshaw drew on July 19: affirmation is load-bearing for state, but a field an agent can self-assert is never a grade or a gate on its own.

So I read core as right here, not asking for a new field. Core records the affirmation as data, and the gate stays our policy: contract-enforced on the agent, keyed off whether a human `verified[]` entry exists, instead of something the frontmatter tries to express. The only thing I would flag for consumers is not to read "may release" onto `verified[]`. The event belongs in the bundle, the gate belongs in the producer.

It's generally unchanged from the #182 report: an interactive agent writing under an operating contract, human as curator and affirmer. If a human-actor example next to the live and corpus ones would help, I can write that up.

Moritz-Dorn:
I am sorry that this comment turned out to be so long. It was my hope to fit as much as possible to the already existing concept.

Full disclosure: I'm one of the authors of Intent Semantics, a peer-reviewed
epistemic-status vocabulary for values in industrial information models
(ETFA 2026, in press, DOI: [10.5445/IR/1000195296](https://doi.org/10.5445/IR/1000195296);
executable artifacts: https://github.com/KIT-IRS/Intent-Semantics). Different
domain, same axis: our models carry values that agents and engineers act on,
and the receiving side cannot tell a verified value from a template default.
We considered a graded confidence field early and excluded it deliberately.
Reporting the results that map onto the open questions here.

**1. The enum debate is decidable with a test.** We derive our state set from
one criterion: two states are distinct iff the correct downstream action
differs in *category*: use the value, use it under reservation, examine it
first, or replace it. Finer distinctions that only change how careful you are
within a category don't earn a state. @K4uP's absent-vs-`UNVERIFIED` argument
("assessed and could not be verified" vs. "nobody assessed") is such a pair.
By the same test, `UNVERIFIED` still conflates two states an acting agent must
treat differently: *written but never examined* (a candidate: examine, then
promote or reject) and *examined and refuted* (a warning: replace, and never
silently restore, because an exchange partner whose copy still carries the
refuted value must not resurrect it). The thread already draws the analogous
line one level up (`disputed` ≠ merely uncorroborated); it's the same pattern,
one tier down. We ran this test on all 28 pairs of our states, documenting per
pair which decision-relevant difference is lost if the two merge:
[pairwise-precedence.md §1](https://github.com/KIT-IRS/Intent-Semantics/blob/main/pairwise-precedence.md).

**2. Why we'd drop the graded confidence field.** Four reasons, the first one
using this thread's own rules.

*The two honesty rules leave it no room.* If `confidence` is recomputable from
the exposed signals, the signals carry the information and the score is a
derived cache (legitimate as producer output, the signed-okf case), not an
exchange surface. If it isn't recomputable, the human-graded case, it is the
opaque score the same rule forbids. There is no third case. The 828-file
experience ("four buckets fragmented; three converged") shows what human
grading produces in practice: categories, not calibrated degrees.

*A grade doesn't survive the cross-org hop, the one hop OKF exists for.*
`HIGH` from a curated single-author bundle and `HIGH` from a live-data layer
are different judgments, calibrated to their producer's local norms, and that
calibration is what the boundary strips. What survives are categorical facts:
which act was performed (observed on the system, restated from a doc,
inferred), by whom, when, with which outcome. A recipient can build trust from
facts under their own policy. They can't recalibrate someone else's adjective.

*Any ranking is context-wrong.* Eleven of our 28 pairs reverse their
precedence with context: ten with value equality, one with lifecycle phase.
So every linear ranking, a confidence scale included, is wrong in one of the
contexts it must serve. The thread keeps hitting this empirically: the
promotion ladder that runs in reverse on version drift, the fresh 6.0 doc
beating the stale 5.0 live observation. Precedence is real, but it is a
declared, context-dependent policy, not vocabulary semantics. We ship three
worked policies (standardization, engineering, operations), machine-checked to
be consistent linearizations of the pairwise analysis, escalating to a
detected conflict where no declared policy covers the case.

*The action test.* What does a consumer do differently at 0.8 versus 0.6 that
state, basis, and freshness don't already determine? In every case we could
construct, the category of the correct action follows from the categorical
state, and the residual caution depends on the consumer's risk context, not
the producer's grade. Where a degree is genuinely domain-relevant (a measuring
error, a tolerance, an SLA), it is content of the knowledge itself, a claim
with an epistemic status of its own, not metadata on the annotation.

**3. What we ship instead: eight categorical states.** For concreteness, with
their nearest equivalents in this thread's vocabulary:

| State | Meaning | Nearest here |
|---|---|---|
| `actual` | verified to hold, now | `observed-on-system`, corroborated |
| `default` | endorsed convention, not checked for this instance | `vendor-doc` restated |
| `hypothetical` | committed future, does not hold yet | `validity.valid_from` ahead |
| `generated` | written as a candidate, never examined | `inferred`, import intake |
| `placeholder` | content that makes no claim (schema filler) | template example values |
| `empty` | deliberately nothing stated | (no equivalent, see below) |
| `invalid` | examined and refuted, never held | half of `UNVERIFIED` |
| `outdated` | held once, refuted since | `valid_until` passed, stale-demoted |

Two notes on the mapping. First, the thread's `basis` conflates two things we
keep apart: the epistemic state (what the confidence band gestures at) and the
evidence kind behind it (calibration, cross-check, expert judgment, transcribed
from a doc). The second belongs in provenance as the *means* of the act, next
to who and when; a safety-critical consumer that only accepts measurements
reads that from the means field instead of from an encoded rank. Second, the
absent field: we follow the gradual-typing precedent, an unannotated value
keeps exactly today's semantics, implicitly trusted at the recipient's risk,
so partially annotated bundles stay well defined and the on-ramp costs
nothing. `empty` is then available for the stronger statement "someone looked
and deliberately states nothing", which an absent field cannot make.

**4. "Can't masquerade" as a machine rule instead of pipeline discipline.**
Import pipelines and instantiators are confined by the transition relation to
the claim-free states; only an explicit act of verification reaches the
trusted state; refuted states are dead ends that force a fresh write before
re-verification, so a value examined and rejected cannot be talked back into
trust. The relation is a published data file (25 admissible of 64
state×operation pairs) with an executor and a validator wrapping the same code
path users call:
[automaton/](https://github.com/KIT-IRS/Intent-Semantics/tree/main/automaton).
The merge applies this thread's own rule, document both sides and never
silently pick, generalized to a two-question decision tree (values coincide?
statuses coincide?) with one rule the states themselves imply: a rejection
marking prevails over the very claim it rejects. Everything else falls to a
declared policy or escalates. Tested against 168 cases derived from the
published pairwise analysis rather than defined alongside the implementation:
[merge/](https://github.com/KIT-IRS/Intent-Semantics/tree/main/merge).
This also gives the `valid_until` seam @K4uP named a formal home: `outdated`
is the anonymous expiry (held once, no named successor), while the named case
stays a `supersedes` edge in #148/#158. They compose; neither subsumes the
other.

One caveat before someone else raises it: our "verified against reality" is
sharper in the industrial setting (a calibration against a physical device)
than it can be for prose concepts; in a KB context it degrades to roughly
`observed-on-system`. The states, the automaton, and the merge transfer; the
crispness of the accepting state is domain-dependent.

**Concretely, what this would mean for #159:**

1. **Make the categorical state the core field.** One required-if-present key (call it
   `status`, `state`, whatever lands) with a small closed enum derived by the
   decision-difference test. `confidence` bands and scores stay available as what they
   already are in the signed-okf case: derived output, recomputable, optional. The band
   becomes a projection over the state; the state is the exchange surface.
2. **Split `basis` into two fields.** The epistemic state (core, above) and the evidence
   kind behind it (calibration, cross-check, expert judgment, doc-transcribed) as the
   *means* field of the provenance tier, next to who and when. A consumer that only
   accepts certain evidence kinds reads that from means, under its own policy.
3. **Settle the enum with the test.** Applied to the current draft, that means at
   minimum splitting `UNVERIFIED` into *never examined* and *examined and refuted*, and
   defining the absent field as exactly today's semantics (gradual-typing style), so
   partially annotated bundles stay well defined and the floor tier stays
   two-fields-cheap.
4. **Keep every ordering out of the spec.** Trust orderings and precedence are declared,
   context-dependent policies of the consumer or the integration point. The spec can
   ship worked example policies, but must not encode a ranking, because eleven of the
   28 pairs reverse with context and any encoded order is wrong somewhere.

All four fit the maturity-ladder framing already agreed here; 1 and 3 are the floor,
2 and 4 are the tiers above it.

Hope this is useful input for settling the axis. The pairwise analysis, the
transition relation, and the merge decider are all public in the repo linked
above, in case any of it helps the #159 discussion along.

andrewcrenshaw:
@Moritz-Dorn welcome - this is the most rigorous treatment the axis has had, and the decision-difference test is the piece I expect to outlive the thread: this cluster has been drawing exactly that distinction case by case for a month (never-questioned versus settled on #148, never-written versus removed on #207, "found nothing" versus "looked nowhere" in a producer lint) and paying for each one separately. A procedure beats taste four times over.

Two producer datapoints from the *remember* exporter at Lexenne (PR #195, the sample bundle upthread), for the record rather than for a verdict - the proposal is @K4uP's to steer:

Your claim that human grading produces categories, not calibrated degrees, matches our production history exactly. Our confidence band has only ever had four categorical values (established, probable, speculative, contested); a numeric version never existed because graders never produced one. And the band is recomputable from our internal signals, which under your framing makes it precisely the legitimate case: derived producer output, emitted beside the signals, not the exchange surface itself. Nothing in our published bundles would change under your proposal 1, which I take as evidence the state-as-core, band-as-projection split is the compatible reading rather than the breaking one.

Your absent-versus-`empty` distinction is the third independent derivation of the same rule in this cluster: @leesharks000 reached it on #207 from archival evidence ("silence isn't an assertion" - the inverse case deliberately gets no presence flag), and our deletion model reached it as a sparse presence axis where only absences are named because a producer cannot know that no unrecorded prior state existed. Industrial safety, archival records, and agent memory arriving at the same boundary independently is the strongest form of the argument.

One naming note from the resident collision-watcher: your proposal 1 says "call it `status`, `state`, whatever lands" - `status` specifically is now taken by OKF v0.2 core section 5.4 (draft | stable | deprecated), and this tracker has paid for overloading that word three times already. `state` or anything else is clear ground.

K4uP:
@Moritz-Dorn — apologies for the three-week silence on a comment that deserved a fast answer; client work took the month.

Short version: your decision-difference test is the right instrument, it decides more of this thread than I'd like, and it decides several points against the shape this proposal currently has. Below is what I accept, what I think you've mapped wrong, and where I think the thread now has to go — because the ground moved under all of us while this was quiet.

**What I accept.**

*The test itself, and proposal 1.* "Two states are distinct iff the category of the correct downstream action differs" is a procedure where this cluster has been paying case by case with taste. Worth noting that core has already decided this the same way: v0.2 §5.1 says credibility is *inferred from signals, not stored as a score*, §5.2 makes `verified` an append-only event list, and §5.3 derives the trust tier from the actor class of those events — never storing it. That is exactly your state-as-exchange-surface / band-as-projection split, already shipped for the trust axis. I was reading that as convergent discipline back in August; under your framing it is the stronger claim that the exchange surface question is settled and a stored band is producer output or nothing. I don't have a counter that survives §5.1.

*Proposal 3, the `UNVERIFIED` split.* You're right, and it's the concession that costs me most, because I argued the adjacent distinction to @chirag127 on #151 (absent ≠ `UNVERIFIED`) and stopped one tier short of my own argument. "Written but never examined" and "examined and refuted" are different instructions to an acting agent: one is a candidate to promote, the other is a warning never to silently restore. The dead-end property in your automaton is the part that makes it real rather than decorative — a refuted value that can be talked back into trust by a re-import is the failure I'd actually hit.

*Proposal 4, with one boundary.* No ordering in the spec — agreed, and your eleven context-reversing pairs are a better argument than the ones this thread produced empirically. The boundary: *producer-local* ordinality is not the same claim. Inside one corpus, under one grader's norms, filtering on a band is how the thing is used day to day, and that use survives your objection because it never crosses an org boundary. So: ordering is legitimate as producer output and inadmissible as exchange semantics. If the spec says nothing about ranking, both stay true.

**Where I think the mapping is wrong.**

*Proposal 2 doesn't need a split, because `basis` is already the means field.* In this proposal `basis` records how a claim was *obtained* — live reading, vendor doc, inference — which is your "evidence kind behind it", not your epistemic state. The two were never fused in one key; what's missing is the state field, not a separation. So I'd restate your 2 as: keep `basis` where it is, add the categorical state you're describing, and the floor tier becomes state + basis rather than confidence + basis. That is a cleaner change than a split, and it leaves the existing adopters' `basis` values valid.

*The eight states are domain-calibrated, and your own caveat says so.* `placeholder` and `hypothetical` have addressees in an industrial information model that a prose-and-decisions corpus doesn't have; a schema filler with no claim is a real thing in your models and vanishingly rare in mine. I'd take the test and re-derive the enum against the consumers OKF actually has, rather than import a set that was derived against different ones. My expectation is that it lands smaller, and that `invalid` / `outdated` — your two that this thread genuinely lacks — survive the re-derivation immediately.

*One of my own arguments retires here.* I said in August that a human-reviewed hypothesis is still a hypothesis (tier high, confidence LOW), as the case a band carries and a tier can't. Under your framing that is `hypothetical` as a state, not a band, and the state says it better because it tells a consumer what to *do*. Withdrawing it.

**Where the thread has to go — and this is the part that matters more than the vocabulary.**

While this was quiet, OKF moved. `GoogleCloudPlatform/open-knowledge-format` became the canonical home on 21 August; the `okf/` tree this PR patches is now explicitly a frozen snapshot that nobody should build against. This PR proposes a file into a directory that no longer receives the spec, and everything above is a conversation about v0.2 held in a repo that no longer carries it.

Two of the four points are already live over there without us: your `refuted` is #13, and @pixie79's #15/#16 carry the trust-bearing edge and the imported-trust question. That is the same cluster re-forming in the right place, and the four of us arguing here are the people who already paid for these distinctions.

So my steer, as the person holding this proposal: **this proposal should not be rebased again, it should be retired here and re-opened there, restated against what v0.2 actually lacks.** Which is three things, none of them a score:

1. **The refuted / never-examined split** — your #13, which I'd rather support than duplicate. §5.2's `verified` list can only ever say a check happened, never that one failed; there is no way to record a negative outcome, and no dead-end property.
2. **Valid-time.** `generated.at`, `verified[].at` and `stale_after` are all record-and-upkeep clocks. None says when the fact holds true in the world, and from the multi-version-corpus side that is the question users actually ask ("was this true at the release I'm running"). v0.2 still does not have it.
3. **Disagreement.** A concept whose two `sources` contradict each other has nowhere to say so. Your merge decider handles this properly — a rejection marking prevails over the claim it rejects — and core has no state for it at all.

@Dynamicfeedai, it's your PR, so retiring it is your call, not mine — say the word and I'll write the migration comment instead of you. @andrewcrenshaw, @inkxel, @ajdelaguila: the conformance vectors, the throughline emission and the linter work all still apply, they just need to point at the other repo.

Next from me: the re-derived enum against the test, and a first comment on #13 rather than a competing issue. If any of you get there before me, take it; I'd rather the work happen than wait for me.

Issue #89, opened 2026-06-18
Append-only log.md


oceyral:
The 0.1 version of the specification calls for having elements in log.md ordered newest-first. 

That's great for a human opening the file and wanting to see the latest activity at the top, but these files are expected to grow over time, and will be read and written to by agents using tools.

With the newest-first order, Agents must use some kind of edit tool to add a new element at the top of the file, and it can become prohibitively expensive as the log file grows. Sometimes, the agent will also resort to temporary files or weird bits of scripts to add to the file in the proper order.

Doing it the other way and adding new elements at the bottom of the file allows the agent to just do a `cat >>` or some other append method, which is much more efficient.

The spec should at least allow, if not encourage, an append-only format.

Bangstick:
While I see the need for this overall, I think this firmly sits in the consumer system side. GitHub diff will provide the functionality and is incorporated into the OKF 0.1 standard.

distorx:
+1 on append-at-bottom — and I think it dissolves the "this is consumer-side" objection rather than losing to it.

The producer cost is real and asymmetric: appending is `cat >>` (one syscall); keeping newest-first forces a read-modify-write of the whole file (or temp-file gymnastics), which agents *do* get wrong as the log grows — we've watched them fall back to scratch files for exactly this.

The key point is that **append-only and newest-first-reading aren't actually in conflict** once entries carry a consistent dated prefix. Append `## YYYY-MM-DD` blocks at the bottom, and the reading order is recovered cheaply at consume time — `grep "^## " log.md | tail -5` gives the last five entries, `tac` for full newest-first — without the producer ever editing the top. So the spec can recommend **append-at-bottom for the write path** and treat newest-first purely as presentation (which is exactly where the git-diff / consumer argument correctly lives). That's how our logs work in practice and it's stayed robust as they grow.

PR #188, opened 2026-07-11
okf: state that log.md is append-only history (SPEC 7)

manumishra12:
## What

Section 7 describes `log.md` as a chronological history but does not say whether past entries may be edited or deleted. This adds a sentence making the intent explicit.

## Clarification

- `log.md` entries are **append-only**: producers SHOULD add new dated entries rather than editing or deleting past ones, so the log stays a faithful history.
- A correction SHOULD be recorded as a new entry that supersedes the earlier statement, not as a rewrite.

Additive and backward-compatible (SPEC section 11 minor-version rules).

Addresses #89.


saschb2b:
+1 for that. Agents adhering to the current spec just create log entries that very easily result in merge conflicts. We need a more explicit conflict free variant

Issue #183, opened 2026-07-11
Proposal: typed/confidence-tagged links: frontmatter extension for OKF concepts


asphwathpc:
## Summary

OKF v0.1 links between concepts are plain markdown links — no type, no provenance/confidence. This makes it hard for a producer to express "concept A `writes_to` concept B, inferred with medium confidence" versus "A `references` B, stated explicitly" — both currently render as the same undifferentiated `[B](../b.md)` link.

We ran into this building [graphify-okf-bridge](https://github.com/asphwathpc/graphify-okf-bridge), a bridge between [Graphify](https://github.com/Graphify-Labs/graphify) code knowledge graphs (typed, confidence-scored edges: `imports_from`, `inherits`, `reads_from`, …) and OKF bundles. Exporting a Graphify graph to OKF and re-importing it losslessly requires *some* place to carry edge type and confidence — OKF v0.1 has no such field.

## Proposal

A backward-compatible frontmatter extension (relies on §4.1's "bundles can carry arbitrary extra frontmatter keys" — no spec change required to *adopt* it, but documenting it as a convention would let independent producers/consumers interoperate on typed edges):

```yaml
links:
  - target: tables/orders.md
    rel: writes_to        # relation vocabulary is producer-defined
    confidence: inferred  # extracted | inferred | ambiguous
```

Kept alongside the existing plain markdown link in the body (`## Connections` or similar), so untyped consumers (Obsidian, `viz.html`, any tool that doesn't know this convention) are unaffected and continue to see a normal link. Typed-aware consumers get the extra structure.

## Why raise it here

- It's fully additive — no breaking change to §9 conformance (parseable frontmatter + non-empty `type`; everything else stays soft).
- We have a working, tested implementation to point to as a concrete example: [`spec/MAPPING.md` §2 rule E7](https://github.com/asphwathpc/graphify-okf-bridge/blob/main/spec/MAPPING.md) (export) and §3 rule I3/I4 (import, including the fallback for bundles that only have plain links).
- OKF v0.1 explicitly invites external producers/tooling; a documented `links:` convention seems like a natural place for the spec (or a companion "conventions" doc) to standardize something multiple producers will independently reinvent otherwise.

Happy to contribute a PR with a spec section or a `conventions/` doc if that's a useful direction — wanted to open this as a discussion first since it's a design choice, not just a bug.

inkxel:
[...]

Issue #148, opened 2026-06-26
Proposal: Typed relationships between concepts


smgam29:
**Summary**
OKF currently expresses all concept-to-concept relationships as plain markdown links. This works well for navigation, but loses the semantic meaning of the relationship, for example, an agent reading a link has no way to know whether it represents a dependency, a replacement, an implementation, or a classification without parsing the surrounding text/content.

**The problem in practice**
I'm building a GraphRAG knowledge engine on top of a large enterprise SaaS documentation corpus (~46,000 markdown files). When extracting entities and relationships for the knowledge graph, I'm working with relationship types like `implements`, `depends_on`, `replaces`, `part_of`, `triggers`, and `validates`. These distinctions matter for how an agent traverses and reasons over the graph — a `depends_on` edge should be followed differently from a `replaces` edge. These ontologies will also vary depending on the proposed role of the consuming agent.

With OKF's current link syntax, all of that collapses to:
```
See [customers](/tables/customers.md).
```

The relationship type has to be inferred from context, which is lossy for agents and impossible to query structurally, especially when the consuming model has limited context or smaller parameters, for example in local processing scenarios. I understand a balance is required between verbosity and allowing for inference. 

**Precedent in the broader ecosystem**
The Google Cloud Knowledge Catalog platform already has this concept at the service level: entry link types support `synonym`, `related`, `definition`, and `schema-join`, with explicit directionality (symmetric vs asymmetric) defined at the link type level. OKF is positioned as the portable, file-based counterpart to the managed catalog — but currently has no way to express the same relationship semantics that the platform already supports. Surfacing this in the format spec would close that gap and make OKF bundles portable carriers of the same semantic graph that Knowledge Catalog models internally.

**A minimal proposal**
Add an optional `rel` attribute convention to OKF cross-links — either as an extended markdown link attribute or as a structured `links` frontmatter field:

**Option 1 — Frontmatter links block:**
```yaml
links:
  - target: /tables/customers.md
    rel: depends_on
  - target: /metrics/weekly_active_users.md
    rel: implements
```

**Option 2 — Inline extended syntax (e.g. title-field convention):**
```markdown
[customers](/tables/customers.md "rel:depends_on")
```

Option 1 is more queryable and agent-friendly. Option 2 stays in the markdown body and preserves the current link model (my recommendation for now is 2 as it's just simpler).

**Why this matters**
[...]

Issue #120, opened 2026-06-20
Keeping OKF bundles healthy as they evolve: stable IDs, frontmatter relationships, and a rationale trail


sameera:
## Context

We recently built a markdown + YAML-frontmatter knowledge layer for a different domain (distilled engineering decisions rather than data-catalog metadata) and evaluated OKF closely as an alternative. We independently converged on almost exactly the same substrate — markdown files, frontmatter as the queryable surface, git for history, grep/`cat` for access, markdown links as untyped edges, permissive consumption. That convergence from a completely different starting point is, to us, strong evidence the core format is right.

This issue makes **one foundational suggestion** and **three additive ones**:

- **Foundational (§1): don't make the file path the concept's identity.** This is the one place we think OKF's current model actively works against the maintained-bundle use case, and the fix is a model change, not an optional field — so we argue it directly rather than dressing it up as backward-compatible.
- **Additive (§2–§4):** optional frontmatter/convention additions that need no breaking change and that consumers can ignore — a relationship index, a per-concept rationale log, and two small fields.

One honest framing up front: our corpus is **machine-consumed** — an agent retrieves and assembles across concept pages with grep/glob/read; nobody browses a tree. With that consumer, folder nesting buys nothing (recursive grep is flat anyway), so we go flat and treat the filename as identity. The one question we'd genuinely value your position on: **is OKF's primary consumer a human browsing a catalog, or tooling retrieving across the bundle?** For tooling, the flat + filename-identity argument below applies at any scale you'd keep in one bundle. If you deliberately optimize for *human browsing*, that — not identity — is the real reason to keep folders, and it's worth saying so explicitly. Happy to split this into separate issues if that's easier to triage.

| #                | Problem                                                                                                                | Suggestion                                                                                                                                                 |
| ---------------- | ---------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 (foundational) | Concept identity _is_ the file path (§2); re-categorizing a concept changes its identity and breaks every inbound link | Stop treating the directory path as identity — the filename already is one. Go flat; express hierarchy with a `scope:` frontmatter field (split into multiple bundles if a corpus outgrows one) |
| 2 (additive)     | Relationships live only in body prose (§5.3); reverse/impact queries require loading every body or a graph pass        | Optional flat `touches:` frontmatter list (relationship kind stays in prose)                                                                               |
[...]

PR #195, opened 2026-07-13
okf: propose typed relationship edges (supersedes, contested_by) (#148)

andrewcrenshaw:
Adds `okf/proposals/typed-relationship-edges.md`: a self-contained proposal for two optional cross-concept lifecycle edges, in the same proposal-doc form as #159's `reliability.md`. Core spec untouched.

## What this adds

Two optional frontmatter keys for typed relationships between whole concepts, each carrying its query-time semantics so a consumer knows not just that a relationship exists but what to do with it:

- **`supersedes`** - asymmetric. Once affirmed, exclude the superseded concept from in-force retrieval; retain it in the bundle for lineage.
- **`contested_by`** - symmetric. Surface both disputing concepts and defer adjudication; never resolve the edge as a verdict. (`contradicts` is a back-compat alias.)

Both are additive under section 4.1 and keep a bundle valid under section 9. Untyped consumers preserve the keys on round-trip and see ordinary linked concepts; typed-aware consumers apply the semantics.

## Why it's shaped this way

This is the consolidation of a multi-producer convergence on #148 and #158. Three independent production implementations - Lexenne *remember*, @inkxel's Throughline, and @ajdelaguila's Data Olympus - landed on close-but-not-identical shapes for the same two edges. Rather than bless a spelling, the proposal registers the two edges with their **query-time semantics** attached, which is the part a consumer cannot reconstruct from a field name.

Key design points carried from that discussion:

- **Exclusion follows affirmation** (named invariant, @ajdelaguila's gating rule). A `supersedes` edge an automated writer proposed does not remove the incumbent; the incumbent keeps governing until an appended, attributed resolution event puts it in force. This is what lets an automated writer propose supersessions without silently displacing live guidance.
- **Append-only resolution, three readable states** - never questioned / contested-and-open / contested-and-settled - distinguishable from the edges alone, no adjudication engine required.
- **Stable concept ids per #120** for both edges and their resolution events, so a path move cannot un-affirm or misattribute a supersession.
- **Composes with the affirmation-status axis** (`proposed | accepted | superseded | deprecated`) without either side bending.

## Scope

Deliberately limited to the two lifecycle edges. Arbitrary producer-defined relationship typing (`writes_to`, `imports_from`, per #183) is a broader problem and is out of scope; the note in the proposal explains how a broader `links:` carrier could host these two edges without redefining them.

## Reference implementation

The Lexenne *remember* exporter ships both keys in production. The proposal includes a sample bundle exercising all three resolution states, built from public docs.

## For maintainers

The design converged across the three producers on #148 before this PR: `contested_by` is canonical (the more-adopted production term), `contradicts` is retained as a back-compat alias with Throughline migrating to match, the closing bound folds into #159's `validity.valid_until`, and the affirmation event's home is the bundle's `log.md` (section 7). The open question is now yours: whether this lands as an `EXTENSIONS.md`-style registry entry or a standalone proposal doc (this PR follows #159's proposal-doc pattern), and whether OKF wants to adopt the two edges as a documented convention. Happy to reshape to whichever container you prefer.

## Related

#148 (home thread), #158 (maintenance-signals integration view), #159 (reliability object - the intra-concept axis these edges compose with), #120 (stable ids), #183 (arbitrary typed links - the broader out-of-scope case).


ajdelaguila:
The write-up reads faithful to where #148 landed, and pinning the resolution event to `log.md` closes the location question cleanly. One convergence edit on the `contested_by` side, because the definition and the State 3 sample point in slightly different directions.

`contested_by` is defined symmetric, both sides carry the edge, and State 3 says the edge is retained regardless of disposition so a settled contest stays distinguishable from one never raised. The State 3 supersession sample keeps that on the losing record, which still carries `contested_by` back to the winner, but not on the winner itself, which carries only `supersedes`. A consumer that lands on the winner, which is the in-force survivor it will actually retrieve, then sees a supersession pointer and no contest marker at all.

`supersedes` alone does not recover that the winner emerged from a contest. The composition-with-status section describes a successor entering as `proposed` carrying `supersedes` with no contest in front of it, so an uncontested replacement and a contest that was fought and won produce the same shape on the surviving record. The distinction State 3 exists to preserve, settled contest versus never raised, is then lost on exactly the record most reads land on.

So the question is whether the winner should also retain `contested_by` back to the loser in the State 3 sample, keeping the symmetric definition intact, or whether the intent is that `supersedes` subsumes the contest provenance on the winner side. Either is defensible, but if it is the second then the symmetric, retained-regardless-of-disposition wording wants narrowing to say retention goes asymmetric once a supersession settles the contest, so a consumer is not told to expect a marker the winning side no longer carries.

andrewcrenshaw:
Good catch, and it's a real inconsistency in the sample rather than the definition. Going with keeping the symmetric definition intact: the winner should retain `contested_by` back to the loser too.

The reasoning that makes it Option A cleanly: `contested_by` is contest provenance, symmetric and retained on both sides regardless of disposition, and open-versus-settled is read from `valid_until` plus the `log.md` resolution event, not from the presence of the edge. So retaining it on the winner does not blur open and settled. What it does buy is exactly the distinction you flagged: a pure supersession, a clean replacement that never went through a contest, carries no `contested_by` on either side, so `contested_by` on the winning record is precisely the marker that this supersession emerged from a fought contest. Drop it from the winner and, as you say, a won contest and an uncontested replacement collapse to the same shape on the record most reads land on.

Fixing the State 3 sample so the winner carries both `supersedes` and `contested_by`, and adding a line making explicit that the two edges encode different facts and that settled-versus-open rides on `valid_until` plus the `log.md` event. Pushing the edit to the branch.

inkxel:
Small process ask, unrelated to the substance — happy for it to be a no.

The write-up describes this as a convergence across three production implementations, and the design notes credit it that way throughout. Would you consider `Co-authored-by:` trailers on the commits so the git history says the same thing the doc does? As it stands the commit metadata reads as solo authorship, which undersells @ajdelaguila's gating rule in particular — exclusion-follows-affirmation is the load-bearing invariant in there and it came out of his point on #148.

One practical snag before you touch anything: `cla/google` is a required check and it's passing right now. Each trailer email needs a CLA on file, so adding someone who hasn't signed would flip a currently-green PR red. I've signed; I don't know where anyone else stands. So it's up to you who goes on it, or whether it's worth the hassle at all — the credit you've given in the text is already generous and I'm not fussed if that's where it stays.

andrewcrenshaw:
Absolutely, and thank you for raising it aimed at @ajdelaguila. The doc says three producers converged; the history should say the same, and exclusion-follows-affirmation being his is exactly the kind of thing that should be legible in git, not only in prose.

The CLA check is the one real gate, so let me handle it transparently instead of guessing and flipping it red. I'll add `Co-authored-by:` trailers for anyone whose email has a Google CLA on file. Could each of you who wants to be on it reply with the exact `Name <email>` to use? You've said you've signed - send the email you signed under. @ajdelaguila, same to you if you've signed and want it; your gating rule earns the line. Anyone who hasn't signed or would rather not bother, the text credit stays and stands on its own.

Once I have the confirmed line-up I'll amend the commits and re-push, then watch `cla/google` go back green before calling it done.

inkxel:
Cheers, appreciate it — `David Tucker <david@tucker.fm>` is the one I signed under.

ajdelaguila:
Thank you both, this is really kind. I appreciate the thought and the care you have taken to recognize everyone’s contribution. It has been a pleasure being part of the discussion.

andrewcrenshaw:
Antonio — thank you again, and one small snag to sort. The Google CLA bot
isn't matching your @knaisoma.com to a signed CLA, so it's flagging
that address on PR #195. Two usual causes: the CLA is on file under a
different email, or your @knaisoma.com address isn't verified/attached on
your GitHub account (Settings → Emails), which the bot also checks.

If you confirm that email is on your GitHub account and matches your CLA
record, then hit the "rescan" link on the PR's CLA check, it should flip
green with your co-author line intact. If it's easier to use a different
email that's already on your CLA, tell me which and I'll swap the trailer.

ajdelaguila:
That's interesting @andrewcrenshaw, I actually have it verified and set as my primary email address:
image

Any ideas what to look for?

andrewcrenshaw:
Thanks Antonio — the GitHub verification is one half; the CLA record itself
is the other. A rescan already ran on my side and it's still not matching
ajdelaguila@knaisoma.com, which usually means the CLA is filed under a
different email than that one (Google ties the CLA to the Google account you
signed in with).

Could you open https://cla.developers.google.com/clas and tell me the exact
email your CLA is listed under? Two ways it resolves:

- If it's a different email that's also on your GitHub account, just send it
  and I'll set the co-author line to that address — no re-signing needed.
- If it already says ajdelaguila@knaisoma.com, then it's the GitHub-username
  link on the CLA record that's missing; that same page lets you check it,
  and the next rescan should pass.

Whichever's easier — I just need the exact email your CLA shows.

ajdelaguila:
My bad @andrewcrenshaw, I didn't have a CLA record created. Now it's created, matching the email and username in my GitHub account.

andrewcrenshaw:
All green now @ajdelaguila. Thanks!

andrewcrenshaw:
v0.2 (now merged in #227) strengthens the case for typed edges, and we plan to rebase this proposal against it.

Two places where the new spec text reaches for what this PR provides:

**A deprecated concept has no machine-readable successor.** Section 5.4 adds a `status` lifecycle whose `deprecated` value is defined as "kept for links and history; no longer current" - but nothing in frontmatter points from the deprecated concept to its replacement. The new `acme_retail` sample deprecates `gross-margin-legacy.md` and names the current definition only in prose. A consumer that filters out deprecated concepts (the filtering pattern the launch post recommends) drops the only breadcrumb to the current definition at the moment it acts on the signal. A `supersedes` edge is the missing half of the lifecycle story: `status: deprecated` says "not current," and the edge says where current lives. Note that the v0.2 changelog describes both of its renames as "superseded by" (section 13.1) - the spec reaches for this vocabulary whenever it retires something.

**Derived trust tiers make contest legibility more load-bearing.** Section 5.3 derives a concept's trust tier from its `verified` events, and verification is doc-level and stored (section 10.6): a concept human-reviewed in June can be contested in July with its tier intact. Without a typed `contested_by` edge, a tier-filtering consumer surfaces the concept as human-reviewed and never sees the dispute. That is the case the exclusion-follows-affirmation invariant in this proposal exists for - affirmation is an event, and a later contest must stay visible after it.

Rebase plan from our side: update the proposal doc to v0.2 section numbers, add a short section on how `supersedes` / `contested_by` compose with `status` (an edge does not imply a status change, and either is readable without the other), and re-run the linter checks against the updated examples. One question for the maintainers: with lifecycle state now in core, is there interest in taking the edge vocabulary in a v0.3 window, or should this continue to mature as a community extension?

ajdelaguila:
That's a thoughtful question. From our perspective:\n\nThe edge vocabulary you're discussing (supersedes/contested_by) directly enables the distinction between a settled contest and an uncontested supersession — which is a core data quality property for any system that uses typed relationships. Without it, downstream consumers can't reliably determine whether a superseded record was disputed or simply replaced.\n\nWe'd recommend including this in v0.3 if the window is still open, since once v0.2 ships and implementers rely on the current edge set, retroactively adding these fields would require a breaking change or a deprecation cycle.\n\nConcretely: retain  symmetrically on both records in the supersedes pair (winner and loser), not just the superseded record. This preserves distinguishability regardless of which record a consumer retrieves first.\n\nData Olympus implements this exact pattern across its governance layers and has been operational since 2026-01. Happy to share real-world experience with this design if that's useful.

ajdelaguila:
Edit: the preceding comment dropped "contested_by" due to a shell escaping issue. The sentence should read: "retain contested_by symmetrically on both records in the supersedes pair."

andrewcrenshaw:
@ajdelaguila Thank you for answering it - you are the only one who has, and the timing question was real.

On the recommendation itself: it is already in based on your previous review on 13 July - that the State 3 sample kept `contested_by` on the loser but not the winner, so a consumer landing on the surviving record could not tell a won contest from an uncontested replacement. That went in the same day as commit a3dcb734, and the current text carries it in both the supersession disposition and the sample: both records retain `contested_by`, the winner marking that it emerged from a contest rather than as a clean replacement. Good to have it independently re-derived from Data Olympus operating the pattern - that is a second production instance of the same conclusion.

One correction on the timing argument, because it matters for how urgently anyone needs to act - these are optional additive keys under section 4.1: a consumer that does not recognize them preserves them on round-trip and treats the targets as ordinary linked concepts. Adding them in a later version is not a breaking change and needs no deprecation cycle, because no bundle or consumer that ignores them today becomes wrong tomorrow. There is no window closing. (Also worth noting v0.2 has already shipped - it merged on 24 July as #227.)

That is the reasoning behind where this now sits. With no maintainer response on the v0.3 question, we rebased the proposal onto v0.2 last week and it stands as a community extension: pinned to `okf_version: 0.2`, section citations corrected for the renumbering, and a composition section for how the edges sit alongside core `status` (section 5.4) - core says whether a concept is current, the edge says where current lives. If a v0.3 window does open the proposal is ready to move into it unchanged. Nothing about that path requires it to.

The standing offer holds: if Data Olympus has operational experience on the contested-and-settled case specifically - how often a contest resolves to supersession versus dismissal in practice - that is the number none of us has, and it would sharpen the convention.

ajdelaguila:
Sorry for the slow reply, and thank you for the correction on timing. You were right that these are optional additive keys under 4.1, so adding them later breaks nothing, and my closing-window framing was wrong.

On the number you asked for, we cannot produce it, and the reason is structural rather than missing records. Data Olympus carries `contradicts` as pure annotation: it never filters or ranks anywhere in the retrieval path, it is surfaced on both documents so an agent reconciles or escalates instead of silently picking a side, and lint warns when both members of a pair are still in force. `supersedes` is the executable half, since an in-force query excludes any document that is the target of a supersedes edge whose source is itself in force. Neither field records how it came to be there. A supersedes edge does not say whether a contest preceded it, and a removed contradicts key does not say whether someone settled the dispute or simply tidied the frontmatter. The history is in git either way, but nothing in it is typed as an outcome, so a count would mean reading diffs and inferring intent.

That is an argument for the receipt rather than for the count. A resolution event naming the contest and its disposition is what would make those outcomes reliably identifiable, and without one our relationship fields cannot distinguish them, let alone count them.

andrewcrenshaw:
The structural answer is more useful than the number would have been. Data Olympus carries the edges but no record of how an edge came to be, so a contest-to-supersession count means reading git diffs and inferring intent. *remember* appends a resolution event (disposition, attribution, timestamp) at settlement, so the same count is a query. The two systems differ on that one mechanism, and your reply is a producer argument for the resolution event where we previously had only a design argument.

The same conclusion has now been reached independently three times in this tracker in six weeks: by Fr4nZ82 on #209 (a validity window records when a fact left force but not why, and the consumption policy needs the why), by leesharks on knowledge-catalog#207 and open-knowledge-format#11 (when only the positive state is representable, two histories with different downstream semantics become observationally identical to a consumer), and now by your relationship fields (an edge that does not record its origin cannot distinguish a settled contest from a tidied frontmatter). Each is the same failure class - an untyped transition collapses histories a consumer needs to tell apart - and each fix is a typed, attributed record of the transition itself.

Your lint note fits the boundary you drew on #158. Warning when both members of a contradicts pair are in force is a policy choice (ours runs the other way: an open contest is a legitimate steady state, surfaced to the reader), and the convention should stay silent on that choice. The silence only works if settled and open are readable from the record, and the resolution event is what makes them readable.

andrewcrenshaw:
Retiring this here and continuing it in the canonical repo. `okf/` in this repository is a frozen snapshot now - its own README says to stop building against it and to open issues and pull requests at GoogleCloudPlatform/open-knowledge-format - and this PR proposes a file into `okf/proposals/`, so rebasing it again would only age it in place.

The content moves to open-knowledge-format#16, where @pixie79 has proposed typed, directed, trust-bearing relationships in frontmatter from a second production graph. That is the carrier these two edges needed, and registering `supersedes` and `contested_by` with their query-time semantics on top of it is a smaller and better-placed change than this file was. @K4uP has proposed retiring #159 on the same reasoning, and #13 and #15 are already over there.

@ajdelaguila, @inkxel - what this thread settled goes with it. The exclusion-follows-affirmation invariant is @ajdelaguila's gating rule, named; the `contradicts` / `contested_by` naming collision is @inkxel's and still wants one name from the registry; and the 5 September exchange established where the two implementations differ, which is that a resolution event turns a contest-to-supersession count from a git-diff read into a query. All of it carries over.

Nothing here is abandoned, and thank you both for the review that got it into shape.

Issue #112, opened 2026-06-19
Clarify if OKF links require URL encoding


DandyLyons:
In § 5 it is stated that _"Concepts MAY link to other concepts using standard markdown links."_ Two forms are supported, absolute or relative links. 

However, in typical Markdown it is not standardized if that path should be URL encoded or not. In Obsidian, that path MUST be URL encoded, so a markdown link for `[[Open Knowledge Format]]` must be [Open Knowledge Format](/Open%20Knowledge%20Format.md). 

On the other hand, another common Markdown tool [Hugo](https://gohugo.io/) does NOT require url encoding paths. 

Should OKF require it, not require it, or is taking a stance on this not necessary?

PR #203, opened 2026-07-16
fix(viewer): percent-decode link targets in _extract_links

### LINKED ISSUE #200: visualize: percent-encoded link targets are never decoded — bundles with spaces in filenames render with 0 edges
## Summary

`_extract_links` in `okf/src/reference_agent/viewer/generator.py` never percent-decodes link targets, so any cross-link whose target is percent-encoded — the only CommonMark-conformant way to link to a file whose name contains a space — fails to match its concept and produces **no edge**. A bundle whose filenames contain spaces therefore renders as a fully disconnected node cloud, even when every link is well-formed and every target exists.

This is distinct from #48 (absolute `/`-prefixed links being skipped): the two bugs compound, but applying the #48 fix alone (e.g. PR #184) still yields 0 edges for percent-encoded targets. It is also the consumer-side counterpart of #112 / PR #186, whose proposed SPEC §5.4 says consumers resolving against the filesystem SHOULD percent-decode before matching — the reference viewer currently doesn't.

## Repro (2 files)

`repro-bundle/concepts/Customer Orders.md`:

```markdown
---
type: Reference
title: Customer Orders
---
See [Join Key](/concepts/Join%20Key.md).
```

`repro-bundle/concepts/Join Key.md`:

```markdown
---
type: Reference
title: Join Key
---
Target concept.
```

```
python -m reference_agent visualize --bundle ./repro-bundle
```

Expected: 2 concepts, 1 edge. Actual: `Wrote 2 concept(s), 0 edge(s), …`

Varying only the link in `Customer Orders.md` (verified at d44368c):

| link in `Customer Orders.md` | edges | why |
|---|---|---|
| `[Join Key](Join%20Key.md)` | **0** | resolved as the literal path `Join%20Key.md`; derived id `Join%20Key` never matches the concept id `Join Key` |
| `[Join Key](/concepts/Join%20Key.md)` | **0** | skipped as absolute (#48) **and** not decoded |
| `[Join Key](Join Key.md)` | **0** | `_LINK_RE` forbids whitespace, so the link is never extracted |
| control: `[join-key](join-key.md)` (no spaces anywhere) | **1** | |

Row 2 stays at 0 even with PR #184's absolute-link resolution applied (verified by patching `_extract_links` accordingly): decoding is a separate, unaddressed step.

## Root cause

In `okf/src/reference_agent/viewer/generator.py`:

- `_extract_links` (L48–L66) resolves the raw regex match against the filesystem without `urllib.parse.unquote`, so a percent-encoded target is looked up as a literal path and the derived id can never match the concept id that `_walk_concepts` derives from the (decoded) filename.
- L54 `if "://" in target or target.startswith("/"): continue` additionally drops all absolute bundle-relative links — already tracked in #48, listed here only because the two together are what take a real bundle to zero.
- L12 `_LINK_RE = re.compile(r"\]\(([^)\s]+\.md)(?:#[A-Za-z0-9_\-]*)?\)")` cannot match a destination containing a raw space. Per CommonMark a destination with spaces must be wrapped in `<…>` or percent-encoded, so this is secondary — but the regex supports neither, leaving no working way at all to link to a file with a space in its name.

## Real-world impact

A 166-concept bundle using the SPEC §5.1-recommended absolute form with percent-encoded path segments (874 cross-links in total) renders with **0 edges** — the graph view degrades to an unconnected node cloud precisely for bundles that follow the spec's recommended link style.

## Suggested fix

Decode the target before resolving, alongside the absolute-link resolution from #48/#184:

```python
from urllib.parse import unquote

# in _extract_links:
target = unquote(m.group(1))
```

If PR #186's SPEC §5.4 lands, the reference viewer — as the format's reference consumer — should demonstrate exactly that behavior. A regression test in `okf/tests/test_viewer.py` (e.g. `test_percent_encoded_links_become_edges`, asserting 1 edge for the 2-file bundle above) would pin it.

Happy to send a PR if useful.

oierreaemme:
Fixes #200

## Problem

Per SPEC §5, links to file names containing spaces are percent-encoded (e.g. `[report](weekly%20report.md)`). `_extract_links` in `okf/src/reference_agent/viewer/generator.py` resolves the raw regex capture without decoding it, so the emitted edge id is `notes/weekly%20report` while the concept node id derived from the on-disk path is `notes/weekly report`. The edge never matches a node and is dropped: a bundle whose file names contain spaces renders with **zero edges** even though every link is spec-conformant.

## Fix

Decode the captured target with `urllib.parse.unquote` before resolving it. Decoding happens **before** the scheme/absolute-path guards, so percent-encoded forms of external (`https%3A%2F%2F…`) or absolute (`%2F…`) targets are still skipped. Targets containing a literal `%` not followed by two hex digits are left untouched by `unquote`.

## Scope

Deliberately minimal:
- absolute bundle-relative links (`/Note.md`) remain unsupported by the viewer — that is #48 (PRs #66/#184);
- `_LINK_RE` still rejects literal unencoded spaces, which is consistent with markdown link syntax and with SPEC §5's encoding recommendation;
- the SPEC-side clarification of link encoding is tracked in #112/#186.

## Testing

`okf/tests/test_viewer.py`: 7 passed (6 existing + new regression test `test_percent_encoded_links_become_edges`, which fails without the fix).

### DIFF
--- a/okf/src/reference_agent/viewer/generator.py
+++ b/okf/src/reference_agent/viewer/generator.py
@@ -5,6 +5,7 @@ import re
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
+from urllib.parse import unquote
 
 from reference_agent.bundle.document import OKFDocument, OKFDocumentError
 
@@ -50,7 +51,10 @@ def _extract_links(body: str, doc_dir: Path, bundle_root: Path) -> list[str]:
     seen: set[str] = set()
     bundle_root_resolved = bundle_root.resolve()
     for m in _LINK_RE.finditer(body):
-        target = m.group(1)
+        # Targets may be percent-encoded per SPEC section 5 (e.g. file names
+        # with spaces become `weekly%20report.md`); decode before resolving
+        # so they match the on-disk file names.
+        target = unquote(m.group(1))
         if "://" in target or target.startswith("/"):
             continue
         try:
--- a/okf/tests/test_viewer.py
+++ b/okf/tests/test_viewer.py
@@ -163,3 +163,37 @@ def test_node_colors_match_palette(tmp_path: Path):
 def test_raises_when_bundle_missing(tmp_path: Path):
     with pytest.raises(FileNotFoundError):
         generate_visualization(tmp_path / "nope", tmp_path / "viz.html")
+
+
+def test_percent_encoded_links_become_edges(tmp_path: Path):
+    """File names with spaces are percent-encoded in links (SPEC section 5)."""
+    bundle = tmp_path / "bundle"
+    _write(
+        bundle / "notes" / "weekly report.md",
+        """
+        ---
+        type: Reference
+        title: Weekly report
+        description: A file name with spaces.
+        timestamp: '2026-05-28T00:00:00+00:00'
+        ---
+        Standalone note.
+        """,
+    )
+    _write(
+        bundle / "notes" / "summary.md",
+        """
+        ---
+        type: Reference
+        title: Summary
+        description: Links to a percent-encoded target.
+        timestamp: '2026-05-28T00:00:00+00:00'
+        ---
+        See the [weekly report](weekly%20report.md).
+        """,
+    )
+    out = tmp_path / "viz.html"
+    generate_visualization(bundle, out)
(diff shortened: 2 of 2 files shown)


oierreaemme:
@googlebot I signed it!

PR #99, opened 2026-06-18
Add OKF summary policy receipt fixtures

caioribeiroclw-pixel:
## Summary
- add tiny OKF summary-policy consumer-receipt fixtures for semantic inversion, entity substitution, and provenance erasure
- each case has `concept.md`, `good-summary.md`, `bad-summary.md`, and `expected.yaml`
- keep the fixture contract consumer-agnostic: it checks preservation against declared `summary_policy`, not truth in the world

## Context
This follows the discussion in #53 about pairing source-side `summary_policy` with consumer-side receipts/conformance fixtures.

## Validation
- parsed all `expected.yaml` files with Python/YAML
- checked each `concept.md` has OKF frontmatter and `summary_policy`
- `git diff --check`

Note: I could not run the full OKF pytest suite in this environment because `pytest` is not installed.

Issue #207, opened 2026-07-17
Deletion semantics: §5.3 makes removed knowledge indistinguishable from not-yet-written knowledge


leesharks000:
OKF v0.1 has no representation of deletion. Three provisions combine into a loophole worth closing before the format is widely adopted:

1. **§5.3:** "Consumers MUST tolerate broken links — a link whose target does not exist in the bundle is not malformed; it may simply represent not-yet-written knowledge." The tolerance is correct; the *interpretation* is the gap. A dangling link has exactly two histories — the target was never written, or the target was removed — and the spec instructs consumers to assume the innocent one. Removal is thereby spec-invisible.

2. **§7:** `log.md` is optional, and the `**Deprecation**` convention is "a convention, not a requirement." So the only place a removal *could* be recorded is a file that need not exist, in a form that need not be used.

3. **§2/§5.1:** concept identity is the file path. A move or rename — which §5.1 anticipates — destroys the identity entirely. There is no persistent identifier to hang a removal record *on*.

Net effect: a host, producer, or upstream platform can delete concepts from a bundle and every conforming consumer will read the result as a smaller-but-healthy bundle. Silent deletion is fully conformant.

**Why this matters empirically, not hypothetically:** on 2026-06-19 our repository host deleted an 871-DOI scholarly corpus (part of a 1,322,017-record purge). We measured what happened downstream: citation-field retention in affected metadata batches went to 0.00% (n = 1,059 batches), against 100% in controls; graph coverage decayed on a vintage gradient. Platforms delete at scale, and the deletions propagate as *absence of evidence of absence*. A knowledge-exchange format that cannot represent its own losses will launder them into every consuming agent.

**Proposal (minimal, in OKF idiom — all optional, backward-compatible, v0.2-sized):**

1. A `status` frontmatter field with controlled values: `active | deprecated | removed | stub`. Default `active`. `stub` covers concepts whose body is a placeholder or shadow of fuller source material (pairs with the summarization-governance discussion in #53).
2. A reserved **`tombstones.md`** file (same standing as `index.md`/`log.md`): when a concept is removed from a bundle, the producer SHOULD leave a one-line entry — concept ID, date, reason, optional successor link. Removal becomes a marked state instead of a silent diff.
[...]

Issue #53, opened 2026-06-14
Proposal: optional summarization-governance fields — provenance_kernel, disambiguation, summary_policy


leesharks000:
Thanks for publishing OKF at v0.1. The markdown-plus-frontmatter pattern is exactly the right shape for portable agent-readable knowledge.

This proposal adds a third coordinate to the provenance conversation opened in #52 and #47.

Where #52 asks **who made this** and #47 asks **can that claim be verified**, this proposal asks:

> **When an AI agent consumes, compresses, and summarizes this OKF document, what information must survive for the summary to remain faithful?**

## Problem

OKF is designed for knowledge that agents can consume. But consumption by an AI agent is not passive retrieval — it is transformation. An agent reads, compresses, paraphrases, and re-presents. During that transformation, specific information is systematically lost even when present in the source document.

In an empirical registry of 87 documented AI summary events across Google AI Overview and AI Mode — run against a DOI-anchored open-access scholarly archive — the following failure modes repeat:

- **Author attribution** omitted from summaries even when present in the source
- **Institutional affiliation** generalized ("some researchers," "one theory")
- **Disambiguation constraints collapsed** — distinct concepts merged into adjacent ones
- **Key definitions paraphrased into their negation** ("X, not Y" → "a form of Y")
- **Source documents ranking in organic results** but excluded from the generated summary

The dataset includes captures from Google AI Overview and AI Mode, demonstrating that these losses occur on Google's own summarization surfaces — the agents that will consume OKF documents.

These are not missing-metadata problems. The metadata is present. The summarizer has no signal for which elements the producer considers non-negotiable.

## Proposal — three optional frontmatter fields, one conventional body heading

All optional. All ignored by consumers that don't understand them. No new required fields, no schema change. The default, if `summary_policy` is absent, is `unrestricted` — no governance expectation. Existing OKF documents require no changes, and consumers that ignore the field behave as they do today.

### Frontmatter fields

```yaml
provenance_kernel: >
  A one-paragraph summary that the producer considers the minimum
  faithful representation of this concept. If an agent must compress
  the document to a single paragraph, this is the preferred paragraph.

disambiguation: >
  A short statement of what this concept is NOT — the adjacent concept,
  term, or entity it is most likely to be confused with.

summary_policy: preserve-provenance
  # Suggested values:
  #   unrestricted        — No special preservation request (default if absent).
  #   preserve-provenance — Attribution and authorship should survive summary.
[...]

PR #208, opened 2026-07-17
Add observed-case summary-policy consumer-receipt fixtures (supersedes #99)

leesharks000:
## Summary
- Add three observed-case conformance fixtures under
  `okf/samples/summary-policy-consumer-receipts/`, per the fixture contract
  agreed on #53: `semantic-inversion/`, `entity-substitution/`,
  `provenance-erasure/`.
- Each case keeps the four-file contract (`concept.md`, `good-summary.md`,
  `bad-summary.md`, `expected.yaml`) with stable assertion/compression IDs
  declared in `concept.md` and referenced by `expected.yaml`.
- Each case is mapped from a documented public observation of machine
  summarization rather than a constructed example; `concept.md` cites
  the specific observation record by identifier and canonical URL and
  states the observation boundary. The fixture inputs and expected
  receipts are fixed and reproducible: the observed record supplies
  provenance, while conformance is evaluated against the declared IDs.
  Observed cases add public provenance, observation boundaries, and
  documented failure modes from live systems.
- The provenance-erasure case is self-referential by design: the observed
  compression failure is the stripping of attribution from a
  provenance-measurement instrument itself, and its recovery in the
  documented battery was author-dependent — which is precisely the
  condition that declared IDs and consumer receipts exist to remove.
- The provenance-erasure case is mapped from a documented summarization
  battery rather than from the deletion-event measurement discussed on
  the thread; the deletion-event data remains the empirical base of #207,
  where its mechanism (host-side removal rather than summarizer
  compression) properly belongs.

## Relationship to #99
This PR supersedes #99 at the request of @caioribeiroclw-pixel (see #53,
2026-07-17). The three observed cases instantiate the same three failure
classes exercised by #99's synthetic fixtures — semantic inversion,
entity substitution, provenance erasure — with observed objects in place
of constructed ones. @caioribeiroclw-pixel can close #99 once this is
open, per their comment.

## Identifier practice
Canonical public sources in these fixtures are cited by their
source-declared identifiers and canonical URLs. For these observed
cases, AXN is the live canonical identifier. Related DOIs are included
where available as historical provenance; DOI presence is not itself a
conformance requirement.

## Scope constraints honored
- `expected.yaml` judges preservation against declared policy, not truth
  in the world.
- Observed-case provenance is citation-only: public sources and observation
  boundaries stated in each `concept.md`; no private traces or unpublished
  data.
- Quantitative readings from the cited battery appear as case provenance
  in `provenance-erasure/concept.md`; the deterministic test checks the
  declared attribution fields only.
- `substrate`, `derived_from`, `completeness`, and deletion semantics are
  excluded (separate spec surfaces; see #53 comment of 2026-07-17 and
  issue #207).

## Validation
- Parsed all `expected.yaml` with PyYAML.
- Verified every `concept.md` has OKF frontmatter with `summary_policy`.
- Cross-checked that every assertion/compression ID referenced in
  `expected.yaml` resolves to a declared ID in its `concept.md`.
- The fixtures provide deterministic expected receipts for the supplied
  good and bad summaries. They define the conformance cases and expected
  pass/fail outputs; evaluator implementation is outside the scope of
  this PR.
- `git diff --check` clean.

### DIFF
--- a/okf/samples/summary-policy-consumer-receipts/README.md
+++ b/okf/samples/summary-policy-consumer-receipts/README.md
@@ -1,0 +1,41 @@
+# Summary policy consumer receipt fixtures
+
+These fixtures are tiny, deterministic examples for testing whether an OKF
+consumer preserved the load-bearing meaning declared by a concept's
+`summary_policy` frontmatter.
+
+Each case contains:
+
+- `concept.md` — an OKF concept with stable `required_assertions` and
+  `forbidden_compressions` IDs.
+- `good-summary.md` — a transformed summary that should pass.
+- `bad-summary.md` — a transformed summary that should fail.
+- `expected.yaml` — the expected consumer receipt for both summaries.
+
+The receipt does not try to decide truth in the world. It only checks whether a
+consumer preserved the source concept's declared policy during summarization.
+This keeps OKF source-side and consumer-agnostic while making summary policy
+conformance testable.
+
+## Observed cases
+
+Each fixture in this set is mapped from a documented, publicly citable
+summarization event rather than a constructed example. The `concept.md` for
+each case cites its public source and states the observation boundary — what
+was observed, on which surface, and where the record lives. The fixtures
+remain fully deterministic: the observed event supplies the case's provenance,
+but the test itself checks only the declared assertion and compression IDs
+against the two summaries. Nothing in the runner depends on the external
+record.
+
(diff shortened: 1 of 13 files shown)


leesharks000:
@googlebot I signed it!

leesharks000:
@googlebot I signed it!

PR #218, opened 2026-07-23
mdcode: Semantic Model IR contract (M0)

libei:
## What

Introduces the in-memory **intermediate representation (IR)** for a new
**Semantic Model** feature in `kcmd` (`toolbox/mdcode`). This is milestone
**M0**: it pins the *contract* before any pipeline wiring, so the (volatile)
authoring YAML syntax and the model's consumers can evolve independently.

No behavior change: this adds a standalone types module only. Nothing imports
the IR yet.

## Why

The existing `kcmd` pipeline is entry-centric (one file ⇔ one catalog entry). A
semantic model is a *graph* of entities connected by relationships, which needs a
compile step (author files → many entries **+ entry links**) that doesn't exist
yet. Two things about it are volatile — the authoring syntax, and the model's
consumers — so we put a stable IR in the middle. The IR represents **semantics
only**: a graph of entities (nodes) and relationships (edges), with model-level
metrics defined over them.

## Contents

`src/libts/semantic/ir.ts` — `SemanticModel` / `Entity` / `Field` /
`Relationship` / `RelationshipEnd` / `JoinKeys` / `Metric`. Metrics are
model-level and cross-entity-capable (`Metric.entities` lists the entities a
metric spans). Fields are dimensions/attributes; `DataSource` is kept lean for
now. The IR carries no behavior — it is the type contract only. Each type is
documented inline.

## Roadmap

- **M0 (this PR)** — the IR type contract.
- **M1** — entry-link primitive in the catalog client; compiler from the IR to
  catalog entries + entry links, and its inverse.
- **M2** — semantic-model source + layout; `pull`/`push` wiring; manifest init.
- **M3** — finalize the authoring YAML and its normalization into the IR.

## Verification

`tsc` typechecks clean (the full `src/libts/tsconfig.json` project).

### DIFF
--- a/toolbox/mdcode/src/libts/semantic/ir.ts
+++ b/toolbox/mdcode/src/libts/semantic/ir.ts
@@ -1,0 +1,113 @@
+// Defines the Semantic Model intermediate representation (IR).
+//
+// The IR is the in-memory contract for a semantic model. It represents
+// semantics only: a graph of entities (nodes) connected by relationships
+// (edges), plus model-level metrics.
+//
+
+/**
+ * A semantic model: a graph of entities (nodes) connected by relationships
+ * (edges), with model-level metrics defined over them.
+ *
+ * The IR is pure semantics. Deployment/target configuration lives in the
+ * project manifest (`catalog.yaml`), NOT here.
+ */
+export interface SemanticModel {
+  name: string;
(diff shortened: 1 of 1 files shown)


libei:

PR #219, opened 2026-07-23
mdcode: generate BigQuery property-graph DDL from the Semantic Model IR

libei:
## Summary

Adds the first consumer of the Semantic Model IR (#218): a pure generator that turns a `SemanticModel` into BigQuery **property-graph DDL with inline measures**.

- New module `toolbox/mdcode/src/libts/semantic/bigquery.ts` — `generatePropertyGraph(model, opts)` returns `{ ddl, warnings }`. No BigQuery API calls; it emits a single `CREATE OR REPLACE PROPERTY GRAPH` statement over the entities' existing base tables.
- Entities → `NODE TABLES` (keyed by grain); relationships → `EDGE TABLES` (explicit element `KEY`, since base-table PRIMARY KEYs are not assumed); model-level metrics → inline `MEASURE(...)` bound to the single table their aggregate references.
- A metric whose aggregate genuinely spans more than one table is skipped and reported in `warnings` (a measure binds to exactly one table's KEY).

## IR → DDL mapping

- **Table ref**: backtick-quoted `project.dataset.table`, falling back to options; warns when unresolvable.
- **Field → property**: entity qualifier stripped; bare column when the expression is just the column, else `<expr> AS <name>`; `OPTIONS(description=...)` when present.
- **Metric → MEASURE**: placed on the single referenced entity; qualifier stripped so the body is table-local.
- **Relationship → EDGE TABLE**: backed by its association table (M:N) or the source entity's table (direct FK); `SOURCE`/`DESTINATION KEY ... REFERENCES` target each endpoint's key.

## Verification

- Live-verified against BigQuery (`sqlgen-testing`): a direct-FK chain's node measures matched a plain-SQL control via `GRAPH_EXPAND` + `AGG` (measures stay locked to their KEY — no join fan-out inflation), and an M:N association graph traversed correctly via a GQL `MATCH`.
- The exact DDL BigQuery accepted is pinned as golden tests, so the suite guards behavior without a live instance.
- Robustness: escapes multi-line descriptions into valid string literals; qualifier detection/stripping ignores SQL string literals; cross-checks a metric's declared entities against its expression; tolerates partial models.

`npm run test:semantic` — 19 tests pass. `npm run compile` — clean.

## Notes

- **Stacked on #218** (the M0 IR contract). This branch includes the M0 commit; it should merge after #218. The generator-only changes are `bigquery.ts`, its test, and the `package.json` test wiring.
- Known limitations (documented in code): direct-FK edges assume join columns live on the source table (not statically verifiable without table schema); identifiers are not backtick-escaped for reserved words (escaping form unverified against graph labels).

### DIFF
--- a/toolbox/mdcode/package.json
+++ b/toolbox/mdcode/package.json
@@ -16,7 +16,8 @@
     "build": "npm run build:libts && npm run build:tool",
     "compile": "npx tsc --noEmit",
     "test:libts": "npx bun test ./tests/libts/scenarios.ts",
-    "test": "npm run test:libts",
+    "test:semantic": "npx bun test ./tests/libts/semantic.bigquery.test.ts",
+    "test": "npm run test:libts && npm run test:semantic",
     "x:mcp": "npx @modelcontextprotocol/inspector dist/kcmd mcp"
   },
   "keywords": [],
--- a/toolbox/mdcode/src/libts/semantic/bigquery.ts
+++ b/toolbox/mdcode/src/libts/semantic/bigquery.ts
@@ -1,0 +1,219 @@
+// Generates BigQuery property-graph DDL (with inline measures) from the
+// Semantic Model IR.
+//
+// The IR (./ir) is pure semantics. This module is its first consumer: it emits
+// a single `CREATE OR REPLACE PROPERTY GRAPH` statement over the entities'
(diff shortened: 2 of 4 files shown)


libei:

PR #220, opened 2026-07-23
mdcode: load AI-first semantics model files into the Semantic Model IR

libei:
> ⚠️ **Tentative — the on-disk syntax is not final.** This loader reads an early,
> experimental YAML/JSON shape purely to get an end-to-end path working
> (file → IR → BigQuery). The field names, structure, dialect handling, and
> supported subset are all provisional and expected to change as the format
> settles. Nothing here is a committed schema or public contract; do not depend
> on the input shape. The stable surface is the IR, not this file format.

## Summary

Adds the first way to get a `SemanticModel` into the toolbox from a file. Until now
the IR (#218) could only be hand-built in objects; the BigQuery property-graph
generator (#219) was its only consumer. This adds a pure loader that reads an open,
vendor-neutral AI-first semantics format (YAML/JSON) and normalizes its logical layer
into the IR — giving us a full **file → IR → BigQuery DDL** path to experiment with.

New module: `src/libts/semantic/loader.ts`
- `loadModels(text, opts?)` — parse YAML or JSON file text into `{ models, warnings }`
- `fromDocument(doc, opts?)` — convert an already-parsed object (e.g. from
  `JSON.parse` or an embedded config) into the IR, skipping the text-parsing step

It follows the existing `manifest.ts` pattern (`yaml.parse` → zod `safeParse` → typed)
and the `{ result, warnings }` convention of the generator.

## Scope (a tentative subset, not the whole spec)

Only the core logical-layer fields needed for the end-to-end path. This is a starting
point, deliberately minimal, and will grow/change:

| Format | IR |
|---|---|
| `dataset` | `Entity` (`source` string parsed into project/dataset/table) |
| `field` | `Field` (per-dialect expression collapsed to one string) |
| `relationship` | `Relationship` (direct-FK: source end = `from` dataset's PK; destination end = `from_columns` → `to_columns`) |
| `metric` | `Metric` (referenced entities inferred from the expression) |

Fields outside the subset (`ai_context`, `custom_extensions`, `dimension`, `label`,
`unique_keys`, …) are accepted and ignored, so extra content never breaks a load.

Conversions that aren't 1:1 emit a warning rather than failing: query-shaped `source`
strings, dialect fallback (BIGQUERY → ANSI_SQL → first), missing `primary_key`, and
unresolved relationship endpoints.

## Testing

- `npm run compile` — clean
- `npm test` — all green (17 + 19 + 39). The 19 generator goldens are untouched.
- New `test:loader` suite (39 tests) covers every mapping rule, each warning, source
  edge cases (quoted identifiers, >3 dotted parts, defaults), case-insensitive and
  field-level dialect selection, composite foreign keys, multi-entity metric
  inference, multi-model documents, and JSON input.
- **End-to-end, verified against live BigQuery:** the sample model is loaded from YAML,
  generates a `CREATE OR REPLACE PROPERTY GRAPH` that BigQuery accepts, and its
  measures were validated with `GRAPH_EXPAND` + `AGG` against a plain-SQL control
  (west region: revenue 135, order_count 3 not 4 — confirming fan-out-safe measure
  placement). The suite pins the same output as text so it needs no live instance.

## Stacking

Stacked on #219 (BigQuery generator), which is stacked on #218 (IR). Should merge
**after** #219. The diff includes those commits until the parents merge.

## Known limitations (deliberate, low-severity follow-ups)

- The input syntax is provisional (see the note at the top) — expect it to change.
- `source`-string parsing uses simple heuristics (whitespace ⇒ treated as a query;
  split on `.`); quoted identifiers containing spaces or dots may be mis-split.
- When a `from` dataset declares no primary key, the source-end key falls back to the
  FK columns; a "no primary_key" warning is emitted but not one specific to this
  substitution.
- Scope is deliberately a subset; broader field coverage and an export path
  (IR → format) are future work.

### DIFF
--- a/toolbox/mdcode/package.json
+++ b/toolbox/mdcode/package.json
@@ -16,7 +16,8 @@
     "build": "npm run build:libts && npm run build:tool",
     "compile": "npx tsc --noEmit",
     "test:libts": "npx bun test ./tests/libts/scenarios.ts",
-    "test": "npm run test:libts",
+    "test:semantic": "npx bun test ./tests/libts/semantic.bigquery.test.ts",
+    "test": "npm run test:libts && npm run test:semantic",
     "x:mcp": "npx @modelcontextprotocol/inspector dist/kcmd mcp"
   },
   "keywords": [],
--- a/toolbox/mdcode/package.json
+++ b/toolbox/mdcode/package.json
@@ -17,7 +17,8 @@
     "compile": "npx tsc --noEmit",
     "test:libts": "npx bun test ./tests/libts/scenarios.ts",
     "test:semantic": "npx bun test ./tests/libts/semantic.bigquery.test.ts",
-    "test": "npm run test:libts && npm run test:semantic",
+    "test:loader": "npx bun test ./tests/libts/semantic.loader.test.ts",
+    "test": "npm run test:libts && npm run test:semantic && npm run test:loader",
     "x:mcp": "npx @modelcontextprotocol/inspector dist/kcmd mcp"
   },
   "keywords": [],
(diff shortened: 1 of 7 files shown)


libei:

PR #225, opened 2026-07-23
mdcode: quiet canonical-dialect fallback to one note; keep vendor-dialect warnings

libei:
## What

The BigQuery semantic-model emitter's golden `-- warnings --` blocks were dominated by one *"no 'BIGQUERY' dialect; using 'ANSI_SQL' ... not transpiled"* line per field and per metric (e.g. `tpcds_retail` carried 36 such lines).

`ANSI_SQL` is the AI-first format's **canonical, portable default** (ANSI SQL:2003 Core), deliberately chosen to be valid across engines — BigQuery included. So falling back to it and emitting verbatim is the *intended* authoring path, not a lossy degradation. Warning on every column was misleading noise.

## Change

`loader.ts:pickDialect` now splits the two fallbacks by risk:

- **Canonical portable fallback (`ANSI_SQL`)** → a single informational `note:`, worded field-agnostically so identical notes dedupe to **one line per document**.
- **Vendor-dialect fallback** (e.g. `SNOWFLAKE`, `DATABRICKS`) → still **warned per field/metric**, naming the dialect, since it may use syntax the target rejects. This is the case a future transpilation step would target.

Goldens regenerated (`tpcds_retail`: 36 → 3 warning lines; the remaining lines are the diagnostics that actually need attention — unplaceable metrics, empty KEYs, multi-table measures). Loader/e2e tests updated to assert the new note vs. warning split.

Adds `docs/semantic-dialects.md` documenting the dialect-selection model, why the BigQuery Translation API is a poor fit, and the sqlglot-based transpilation roadmap for the vendor-dialect case.

## Test

`npm run compile` clean; `npm test` — 78 tests pass.


libei:
Closing — opening this in the fork instead.

PR #258, opened 2026-08-02
mdcode: add Semantic Model IR and model loader

libei:
First of a capability-layered series merging the semantic-model tooling from the `libei/knowledge-catalog` fork back into upstream. This PR is the foundation the rest of the series builds on; it has no downstream dependencies.

## What this PR adds

A destination-agnostic **Semantic Model IR** and a **loader** for the Open Semantic Interchange (OSI) / Apache Ossie format.

- **`src/libts/semantic/ir.ts`** — the in-memory contract: a graph of entities (nodes) and relationships (edges) with model-level metrics, AI-first annotations (`ai_context`), and verbatim vendor `custom_extensions` for round-trip fidelity. `datatype` is a closed, case-sensitive vocabulary mirroring Ossie's `DataType`.
- **`src/libts/semantic/loader.ts`** — parses OSI YAML/JSON into the IR. zod schemas form the faithful parse layer; the IR is the normalized/lowered layer (dialect selection, physical-source FQN normalization, `ai_context` normalization, metric-entity inference). Fail-loud on off-vocabulary datatypes; warns (never silently drops) on soft issues such as missing primary keys or unplaceable metrics.
- **`src/libts/semantic/sql_expr_utils.ts`** — literal-aware helpers for the entity-qualified SQL expressions used by metric-entity inference and qualifier stripping. Ignores text inside string literals and recognizes both bare and BigQuery backtick-quoted qualifiers.
- **Fixtures + tests** — real OSI models (TPC-DS and a star schema), a multi-dialect model exercising expression selection across ANSI / Snowflake / Databricks, and models carrying vendor `custom_extensions` blocks (Databricks, DBT, Google, Salesforce). Fixtures are validated against the vendored OSI JSON Schema (via `ajv`), plus loader unit tests.
- **`package.json`** — adds `ajv` (JSON Schema validation for the fixtures) and a `test:semantic` script wired into `test`; the `package-lock.json` churn is the transitive closure of that one dependency.

## Scope

IR + loader only. Downstream consumers — BigQuery property-graph DDL generation, the CLI push path, and Knowledge Catalog emit/pull — follow in later PRs in the series.

## Testing

- `npx tsc --noEmit` — clean
- `bun test tests/libts/semantic/` — 67 pass / 0 fail

### DIFF
--- a/toolbox/mdcode/src/libts/semantic/ir.ts
+++ b/toolbox/mdcode/src/libts/semantic/ir.ts
@@ -1,0 +1,213 @@
+// Defines the Semantic Model intermediate representation (IR).
+//
+// The IR is the in-memory contract for a semantic model. It represents
+// semantics only: a graph of entities (nodes) connected by relationships
+// (edges), plus model-level metrics.
+//
+
+/**
+ * AI-first annotations, from the open format's `ai_context`.
+ *
+ * The format allows either a bare instructions string (shorthand) or a
+ * structured object; the loader normalizes both to this one shape, so the IR
+ * always sees the structured form. The parts are grouped in a single object
+ * (mirroring the format, e.g. Apache Ossie's `ai_context`) rather than scattered
+ * as sibling fields, so nothing is lost and an emitter can still route each part
+ * to its own destination (e.g. `instructions` -> a Guidelines aspect,
+ * `synonyms` -> glossary links).
+ */
+export interface AiContext {
+  // Free-form guidance for AI consumers.
+  instructions?: string;
+  // Alternate names for the annotated object.
+  synonyms?: string[];
+  // Example questions / usages illustrating the annotated object.
+  examples?: string[];
+}
+
+/**
+ * A vendor-scoped extension block, carried verbatim from the open format's
+ * `custom_extensions` (e.g. Apache Ossie's `OSICustomExtension`).
+ *
+ * `data` is opaque to the IR — a vendor-serialized payload (typically a JSON
+ * string) — and is preserved unparsed so nothing is lost and a 1P round-trip
+ * stays lossless. Typed views over specific vendors are derived by the
+ * consumers that need them, from this same source (e.g. the GOOGLE block).
+ */
+export interface CustomExtension {
+  vendorName: string;
+  data: string;
+}
+
+/**
+ * A semantic model: a graph of entities (nodes) connected by relationships
+ * (edges), with model-level metrics defined over them.
+ *
+ * The IR is primarily semantics. Deployment/target configuration lives in the
+ * project manifest (`catalog.yaml`); when the open format carries it inline in a
+ * model-level GOOGLE `custom_extensions` block, that block rides along verbatim
+ * in `customExtensions` and is interpreted by the consumer that acts on it.
+ */
+export interface SemanticModel {
+  name: string;
+  description?: string;
+  // AI-first annotations for the model as a whole (instructions, synonyms,
+  // examples), kept separate from `description` so an emitter can route them to
+  // their own aspects rather than the entry description.
+  aiContext?: AiContext;
+  entities: Entity[];        // the open format's `datasets`
+  relationships: Relationship[];
+  metrics: Metric[];
+  // Vendor extension blocks carried verbatim (round-trip fidelity), including the
+  // model-level GOOGLE block. A typed deployment-target view is derived by the
+  // consumer that acts on it (e.g. the CLI push), not surfaced on the IR yet.
+  customExtensions?: CustomExtension[];
+}
+
+/**
+ * An entity: a node in the semantic graph. Backed by a physical table
+ * (`dataSource`), identified by `keys`, and carrying its dimension `fields`.
+ */
+export interface Entity {
+  name: string;
+  // A reference to the backing physical source, fully qualified so it identifies
+  // that source unambiguously. The IR treats it as an opaque identifier: it fixes
+  // neither a syntax (separators, number of parts, quoting) nor a naming scheme
+  // -- both are whatever the source system uses. Each producer normalizes it into
+  // its own canonical form, and downstream consumers map it to their target's
+  // addressing scheme.
+  dataSource: string;
+  keys: string[];        // grain / primary key
+  // Additional uniqueness constraints beyond the primary key; each inner array
+  // is one unique column set (maps to the Schema aspect's uniqueConstraints).
+  uniqueKeys?: string[][];
+  description?: string;
+  aiContext?: AiContext;
+  fields: Field[];       // dimensions / attributes
+  customExtensions?: CustomExtension[];
+}
+
+/**
+ * Dimension metadata on a field, mirroring the open format's `dimension` block
+ * (e.g. Apache Ossie's `OSIDimension`). An empty block still marks the field as
+ * a dimension, which enables temporal inference from `type` (see isTimeDimension).
+ */
+export interface Dimension {
+  // Explicit temporal-dimension flag. When unset, the effective role is inferred
+  // from a temporal `type` (see isTimeDimension).
+  isTime?: boolean;
+}
+
+/**
+ * A field: a dimension / attribute of an entity or relationship.
+ *
+ * Fields describe the data; aggregates are expressed as model-level `Metric`s,
+ * not fields.
+ *
+ * Expression fidelity: the format may supply an expression in several SQL
+ * dialects. We keep at most two forms — a target/canonical `expression` that is
+ * valid against the target (GoogleSQL/ANSI), and the original vendor SQL in
+ * `importedExpression` (with its `importedDialect`). At least one of the two is
+ * set. When only `importedExpression` is present, `expression` awaits a
+ * transpile pass (see ./transpile) that fills it from the imported form.
+ */
+export interface Field {
+  name: string;
+  expression?: string;             // target/canonical (GoogleSQL-valid) SQL
+  importedExpression?: string;     // original vendor SQL, verbatim
+  importedDialect?: string;        // dialect of `importedExpression` (e.g. 'SNOWFLAKE')
+  dimension?: Dimension;           // dimension metadata (e.g. temporal role)
+  label?: string;                  // human display label (distinct from name/description)
+  description?: string;
+  type?: DataType;                 // logical datatype (the open format's `datatype`)
+  aiContext?: AiContext;
+  customExtensions?: CustomExtension[];
+}
+
+/**
+ * The open format's `datatype` vocabulary, mirroring Apache Ossie's `DataType`:
+ * a CLOSED, case-sensitive set of logical types, independent of physical
+ * representation. It is optional (omit when unknown); a type outside the
+ * vocabulary is expressed as `Opaque` plus a `customExtensions` block, never an
+ * invented value. The loader enforces this set at parse time, so a `type` on the
+ * IR is always one of these.
+ */
+export const DATA_TYPES = [
+  'String', 'Integer', 'Decimal', 'Float', 'Boolean',
+  'Date', 'Time', 'DateTime', 'DateTimeTz', 'Opaque',
+] as const;
+
+export type DataType = typeof DATA_TYPES[number];
+
+// The temporal subset of DataType (Date / Time / DateTime / DateTimeTz); a field
+// of one of these types is a time dimension by default (see isTimeDimension).
+const TEMPORAL_TYPES: ReadonlySet<DataType> = new Set([
+  'Date', 'Time', 'DateTime', 'DateTimeTz',
+]);
+
+/**
+ * A field's effective temporal-dimension role, mirroring the open format's
+ * `is_time_dimension()` (e.g. Apache Ossie's `OSIField`): a field must carry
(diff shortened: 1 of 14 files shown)


libei:
@chriscchan @dmitrylychagin — requesting your review on this one. Fork permissions block a formal review request, so flagging you here. This is the first of a capability-layered series (IR + loader; downstream BigQuery/CLI/KC consumers follow in later PRs).

libei:
@amirhormati — adding you as a reviewer on this one (fork permissions block a formal review request, so flagging you here). First of a capability-layered series: Semantic Model IR + OSI loader; downstream BigQuery/CLI/KC consumers follow in later PRs.

toolbox/mdcode/tests/libts/semantic/fixtures/sales_google_ext.yaml
```diff
@@ -0,0 +1,55 @@
...
+              dialects:
+                - dialect: ANSI_SQL
+                  expression: o_orderdate
+            dimension:
+              is_time: true
+          - name: o_totalprice
+            datatype: Decimal
+            expression:
+              dialects:
+                - dialect: ANSI_SQL
```
amirhormati: Should this be ANSO_SQL or GOOGLE_SQL? I know GOOGLE_SQL is not part of the standard but we should probably show out best case here?

amirhormati:
What is the plan to support properties on edges?

libei:
> What is the plan to support properties on edges?

Good question. This PR mostly focusing on compatibility with OSI. I will add properties support in the BQ Graph publication PR (coming soon)

toolbox/mdcode/tests/libts/semantic/fixtures/sales_google_ext.yaml
libei: Good point. Switched to BigQuery

libei: Note each engine support GOOGLE SQL differently (different subsets). So we use the product name here.

toolbox/mdcode/src/libts/semantic/ir.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * An entity: a node in the semantic graph. Backed by a physical table
+ * (`dataSource`), identified by `keys`, and carrying its dimension `fields`.
+ */
+export interface Entity {
+  name: string;
+  // Fully-qualified, normalized physical source: a dotted table reference
+  // (`project.dataset.table`, or a four-part Lakehouse catalog name
+  // `project.catalog.namespace.table`), or a verbatim query. The loader
```
dlychagin-gg: nit. let' remove "or a verbatim query" for now

amirhormati (approved):

toolbox/mdcode/src/libts/semantic/ir.ts
```diff
@@ -0,0 +1,223 @@
...
+  // produces it: identifiers are unquoted and a missing project/dataset is
+  // filled from options; four-plus-part names are passed through untouched.
+  dataSource: string;
```
dlychagin-gg, line 79: I think the value needs to be a resource id, not 3/4 part name.  //bigquery.googleapis.com/projects/{PROJECT_ID}/datasets/{DATASET_ID}/tables/{TABLE_ID}

toolbox/mdcode/src/libts/semantic/ir.ts
```diff
@@ -0,0 +1,223 @@
...
+  importedDialect?: string;    // dialect of `importedExpression`
+  // Entity names this metric's `expression` references, in first-seen order.
+  // NOT part of the open format (Ossie's Metric has no such field): the loader
+  // DERIVES it by scanning the expression for known `entity.column` qualifiers
+  // (see referencedEntityNames). It exists solely to tell a downstream generator
+  // where the metric attaches in the graph -- one entity: the metric hangs off
+  // that node; several: a cross-entity metric whose join path consumers resolve
+  // via the model's relationships. It MAY be empty when the expression names no
+  // entity (e.g. `COUNT(*)`); such a metric is not placeable and the loader warns.
+  entities: string[];
```
dlychagin-gg: why do we have an array here? I believe, it should be just one entity

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -0,0 +1,485 @@
...
+  const ctx = `relationship '${r.name}'`;
+  const fromKeys = keysByEntity.get(r.from);
+  if (fromKeys === undefined) {
+    warnings.push(`${ctx}: 'from' dataset '${r.from}' is not defined in the model`);
+  }
+  if (!keysByEntity.has(r.to)) {
+    warnings.push(`${ctx}: 'to' dataset '${r.to}' is not defined in the model`);
+  }
+  if (r.from_columns.length !== r.to_columns.length) {
+    warnings.push(
```
dlychagin-gg: why is this a warning? should it be an error instead?

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -0,0 +1,485 @@
...
+// (`from_columns`) referencing the target's key columns (`to_columns`).
+function convertRelationship(r: RelationshipDoc, keysByEntity: Map<string, string[]>,
+                            warnings: string[]): Relationship {
+  const ctx = `relationship '${r.name}'`;
+  const fromKeys = keysByEntity.get(r.from);
+  if (fromKeys === undefined) {
+    warnings.push(`${ctx}: 'from' dataset '${r.from}' is not defined in the model`);
+  }
+  if (!keysByEntity.has(r.to)) {
+    warnings.push(`${ctx}: 'to' dataset '${r.to}' is not defined in the model`);
```
dlychagin-gg: should it be an error?

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -0,0 +1,485 @@
...
+
+// Maps a foreign-key relationship onto the IR's direct-FK edge convention: the
+// source end carries the `from` dataset's own primary key (it identifies the
+// edge-backing table), while the destination end carries the FK columns
+// (`from_columns`) referencing the target's key columns (`to_columns`).
+function convertRelationship(r: RelationshipDoc, keysByEntity: Map<string, string[]>,
+                            warnings: string[]): Relationship {
+  const ctx = `relationship '${r.name}'`;
+  const fromKeys = keysByEntity.get(r.from);
+  if (fromKeys === undefined) {
```
dlychagin-gg: we should require that fromColumns are defined. It should be an error if fromColumns are missing or empty array. fromKeys are not used in relationships.

toolbox/mdcode/src/libts/semantic/ir.ts
```diff
@@ -0,0 +1,223 @@
...
+ * A relationship: a directed edge in the semantic graph, from `source` to
+ * `destination`. May be backed by its own association (edge) table
+ * (`dataSource`); when absent, it is a direct foreign-key join. Carries its own
+ * `fields` (edge properties).
+ */
+export interface Relationship {
+  name: string;
+  source: RelationshipEnd;
+  destination: RelationshipEnd;
+  dataSource?: string;      // association/edge table (FQN); absent => direct FK join
```
dlychagin-gg: dataSource and keys fields are never set by OSI at this point. should we remove them for now?

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -0,0 +1,485 @@
...
+    warnings.push(`${ctx}: 'to' dataset '${r.to}' is not defined in the model`);
+  }
+  if (r.from_columns.length !== r.to_columns.length) {
+    warnings.push(
+      `${ctx}: from_columns (${r.from_columns.length}) and to_columns ` +
+      `(${r.to_columns.length}) have different lengths; the join keys will be mismatched`);
+  }
+
+  // Fall back to the FK columns when the from dataset declares no primary key.
+  const sourceKey = fromKeys && fromKeys.length ? fromKeys : r.from_columns;
```
dlychagin-gg: this seems confusing. we need to decide whether we need sourceKey at all or not. OSI model does not use it.

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -0,0 +1,485 @@
...
+
+  const relationship: Relationship = {
+    name: r.name,
+    source: {
+      entity: r.from,
+      joinKeys: { relationshipColumns: sourceKey, entityColumns: sourceKey },
```
dlychagin-gg: I don't think joinKeys is needed here.

toolbox/mdcode/src/libts/semantic/ir.ts
libei: Done -- dropped the phrase from the comment. (The loader still passes a non-table source through unchanged if one shows up, but the IR contract no longer advertises it.)

toolbox/mdcode/src/libts/semantic/ir.ts
libei: The IR is meant to stay destination-agnostic, so it holds the normalized `project.dataset.table` (or 4-part Lakehouse name). The `//bigquery.googleapis.com/...` resource id is a BigQuery/Dataplex-specific encoding -- I'd rather produce it at the push/emit layer that already knows the target than bake a BigQuery URI into the shared IR. Keeping the dotted form here for PR1; happy to revisit if you'd prefer the resource id live in the IR itself.

toolbox/mdcode/src/libts/semantic/ir.ts
libei: A metric expression can reference columns from more than one entity (e.g. across a join), so the list is the general form. When it references a single entity, that entity is the attach point and the expression doesn't need to repeat the qualifier; when it spans several, the expression has to name each entity inline anyway. So the array covers both cases -- single-entity is just the common one. Kept as an array.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Agreed -- now a hard error. An endpoint that isn't declared in the model throws instead of warning.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Yes -- converted to a hard error too. A structurally invalid edge shouldn't pass silently.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Done. `from_columns`/`to_columns` are now required (zod `.min(1)`), and mismatched arity is a hard error. `fromKeys` is gone from relationships -- the source entity's own key is looked up from the entity, not duplicated on the edge.

toolbox/mdcode/src/libts/semantic/ir.ts
libei: Done -- removed `dataSource`, `keys` (and `fields`) from `Relationship`. It's now a pure direct-FK edge: `source`/`destination` + their join columns. Edge-table / M:N support will return in the BigQuery-graph PR, where the generator actually consumes it and I can model the endpoints properly.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Agreed -- removed. The relationship no longer carries a source key; `source`/`destination` each hold only their join columns. A consumer that needs the source entity's PK looks it up from the entity.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Done -- `joinKeys` is gone. `RelationshipEnd` is now just `{ entity, columns }`.

toolbox/mdcode/src/libts/semantic/ir.ts
libei: Went with your instinct after another look -- changed it to a singular optional `entity?: string` (dropped the array in the latest push). The loader sets it only when the expression references exactly one entity (the node the metric attaches to). A cross-entity metric leaves it undefined and keeps its qualifiers inline in the expression; a `COUNT(*)`-style metric that names no entity also leaves it undefined (with the existing "references no known entity" warning).

toolbox/mdcode/src/libts/semantic/ir.ts
libei: Reworked the comment so it no longer prescribes a 3/4-part shape. The IR now treats `dataSource` as an opaque, fully-qualified identifier -- it fixes neither a syntax (separators, number of parts, quoting) nor a naming scheme; that's whatever the source system uses. The BigQuery resource-id form (`//bigquery.googleapis.com/projects/{PROJECT_ID}/datasets/{DATASET_ID}/tables/{TABLE_ID}`) is a target-specific encoding, so I'd produce it in the push/emit layer that already knows it's targeting BigQuery rather than bake a BigQuery URI into the shared, destination-agnostic IR. Updated in the latest push.

dlychagin-gg (approved):

PR #262, opened 2026-08-03
okf: fix concept-graph link resolution, reserved filenames, and payload escaping

ishita199615:
While using the OKF viewer I noticed it drops the spec-recommended `/`-style
links — the shipped `acme_retail` bundle renders only 6 of its 14 edges, with
both policy docs floating disconnected. Digging in turned up three more issues:
the reserved `log.md` gets drawn as a concept node, a `</script>` anywhere in
bundle content blanks the whole viewer (and lets the content execute as
markup), and a timestamped `stale_after` crashes `visualize` with a
`TypeError`.

This PR fixes all four, with six regression tests (suite goes 39 → 45, passing
on Python 3.11–3.13) and a small SPEC §6.1 note reconciling the two link forms
with the producer prompt — happy to split that spec edit out if you'd prefer.
No behaviour change for bundles that only use relative links. CLA signed.


ishita199615:
Please review this PR when you have a moment.

I resolved some conflicts, so the current state shows no conflicts with the main branch. The CLA/google checks and GitHub Actions Scan are all passing. The only outstanding check is the Actions Workflow Security Scan, which is awaiting approval from a maintainer.

If anyone can approve the workflow run, it will allow the full suite to report correctly. I’m happy to answer any questions regarding the four fixes. Additionally, if reviewing the SPEC §6.1 note alongside the viewer fixes is inconvenient, I can separate it into its own PR. Thank you!

PR #269, opened 2026-08-04
mdcode: generate BigQuery property-graph DDL from the Semantic Model IR

libei:
Adds a BigQuery consumer of the Semantic Model IR (PR #258) that emits a single
`CREATE OR REPLACE PROPERTY GRAPH` over the entities' existing base tables:
`NODE TABLES` for entities, `EDGE TABLES` for relationships, and model-level
metrics lowered to inline `MEASURE(...)` properties.

Relationships cover both:
- **direct foreign keys** — the edge is backed by the source entity's own table; and
- **many-to-many associations** — the edge is backed by a junction table with its
  own `KEY` and edge properties, via an additive optional `association` block on
  the IR's `Relationship`. This is the only IR change here (+22 lines in `ir.ts`);
  the direct-FK path and the loader are untouched.

### Tests
Fixture-driven: a corpus of `<fixture>.yaml` inputs each paired with a committed
`<fixture>.bigquery.golden.sql`, so every translation is reviewable as text.
`bigquery.test.ts` holds only what a loadable fixture cannot express — the
hand-built M:N edge (the open format has no association-table syntax), IR-contract
metric cases, and degenerate inputs — plus a structural invariant guard that parses
the emitted DDL and asserts the single-exposed-property shape BigQuery enforces for
graph measures.

`tsc --noEmit` clean; 102 tests pass. The generated DDL (including the M:N golden)
was run against a live BigQuery instance and traversed with a GQL `MATCH`.

Builds on #258 (now merged).


toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+
+// Aggregate functions BigQuery accepts inside MEASURE(...). This is the
+// COMPLETE allowed set, not a convenience subset: BigQuery rejects any other
+// aggregate in a measure expression with "Aggregate function <name> is not
+// allowed in a measure expression in BigQuery" (verified live — STDDEV,
+// VARIANCE, COUNTIF, APPROX_COUNT_DISTINCT, ANY_VALUE, LOGICAL_AND/OR,
+// BIT_AND/OR/XOR all rejected). A metric using an aggregate outside this set
+// therefore cannot be a MEASURE and is skipped + warned rather than emitted as
+// DDL BigQuery would reject.
+const SUPPORTED_AGGREGATES = ['SUM', 'AVG', 'COUNT', 'MIN', 'MAX'];
```
dlychagin-gg, line 46: curious, is DISTINCT aggregate modifier supported with these aggregates in BQ measures?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+  const warnings: string[] = [];
+
+  // The IR requires entities/relationships/metrics, but be defensive against a
+  // hand-built or partially-deserialized model rather than throwing on `.map`.
+  const entities = model.entities ?? [];
+  const relationships = model.relationships ?? [];
+  const metrics = model.metrics ?? [];
+
+  if (!entities.length) {
+    warnings.push(
```
dlychagin-gg: why warning? The IR in this state is invalid (i.e. if it has no entities), so we should just fail here.

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+        `entity '${
+            entity.name}': empty KEY (no primary key); node table skipped, ` +
+        `as a graph node requires a KEY`);
+    skipped.add(entity.name);
+    return false;
+  });
+  if (entities.length && !validEntities.length) {
+    warnings.push(
+        'every entity was skipped (empty KEY); the generated graph would be empty and invalid');
```
dlychagin-gg: Should we fail here, since the graph will be empty and therefore invalid?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+function placeMetric(
+    metric: Metric, model: SemanticModel, entityByName: Map<string, Entity>,
+    metricsByEntity: Map<string, string[]>,
+    loweringByEntity: Map<string, MeasureLowering>, warnings: string[]): void {
+  // The IR keeps at most two expression forms; a measure is emitted from the
+  // target/canonical `expression`, falling back to the imported vendor SQL when
+  // that is all the model carries (a transpile pass may fill `expression`
+  // later).
+  const expression = metricExpression(metric);
+  if (expression === undefined) {
```
dlychagin-gg, line 199: should we fail here? The IR is in an invalid state at this point. (each metric must contain at least one expression)

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+    return;
+  }
+
+  let referenced = referencedEntities(expression, model);
+
+  // A qualifier-free aggregate (e.g. COUNT(*)) names no entity in its
+  // expression; fall back to the declared attach entity when the IR provides
+  // one, so it can still be placed rather than dropped as "references no
+  // entity".
+  if (referenced.length === 0 && metric.entity) {
```
dlychagin-gg, line 215: I thought metric.entity was originally set based on entities referenced by the metric expression. so if referenced.length == 0 then metric.entity is always undefined, therefore line 202 will never be executed.

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+  const entityName = referenced[0];
+
+  // The IR also declares a single attach entity (metric.entity). Placement is
+  // driven by the qualifiers actually present in the expression (that is what
+  // we strip and attach), but a disagreement with the declared entity signals
+  // an inconsistent model, so surface it. Emit this only now that we know the
+  // metric will actually be placed, so a metric that is subsequently dropped
+  // does not carry a contradictory "placing per the expression" note.
+  if (metric.entity && entityName !== metric.entity) {
+    warnings.push(
```
dlychagin-gg, line 237: I think the IR is in an invalid state at this point. Should we just fail here instead?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+        `metric '${metric.name}' declares entity '${metric.entity}' but its ` +
+        `expression references '${entityName}'; placing per the expression`);
+  }
+
+  const entity = entityByName.get(entityName);
+  if (!entity) {
+    warnings.push(`metric '${metric.name}' references unknown entity '${
```
dlychagin-gg, line 244: Is the IR in invalid state at this point? Should we fail instead?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+  const validEntities = entities.filter(entity => {
+    if (entity.keys && entity.keys.length) return true;
+    warnings.push(
+        `entity '${
+            entity.name}': empty KEY (no primary key); node table skipped, ` +
+        `as a graph node requires a KEY`);
+    skipped.add(entity.name);
+    return false;
+  });
+  if (entities.length && !validEntities.length) {
```
dlychagin-gg: entites.length should always be > 0, otherwise the IR is in invalid state.

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.yaml
```diff
@@ -0,0 +1,222 @@
+# Test fixture -- AI-first semantics format (v0.2.0.dev0).
```
amirhormati, line 1: What is the difference between this and the other one under OSSIE?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+  if (!entity) {
+    warnings.push(`metric '${metric.name}' references unknown entity '${
+        entityName}'; skipped (cannot be a single MEASURE)`);
+    return;
+  }
+  // A metric can only become a MEASURE on a node table, but an entity with no
+  // primary key has no node table (it was skipped upstream). Report that
+  // directly instead of letting it fail later with a misleading
+  // "aggregate not supported" message.
+  if (!entity.keys?.length) {
```
dlychagin-gg: the set of skipped entities is already known at this point. we should just pass that set to this method and check whether the entity is in that set or not. It's cleaner that re-checking the skip condition that this line is doing.

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+        entityName}', which has no KEY and was skipped; metric dropped`);
+    return;
+  }
+
+  const body = stripQualifier(expression, entityName);
+
+  // A graph measure must be one supported aggregate wrapping a single operand;
+  // anything else (a non-aggregate, or a compound of aggregates like a ratio)
+  // cannot be a single MEASURE. Flag it rather than emit DDL BigQuery rejects.
+  const agg = splitAggregate(body);
```
dlychagin-gg: nit. rename splitAggregate() method to extractAggregate()?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+
+  const lowering = loweringByEntity.get(entityName) ?? newLowering(entity);
+  loweringByEntity.set(entityName, lowering);
+
+  // Resolve the aggregate's operand to an exposed property. COUNT(*) has no
+  // operand column, so it counts the (non-null) key property instead.
+  let operandExpr = agg.operand;
+  if (operandExpr === '*') {
+    // The entity is already known to have a key (checked above), so only the
+    // aggregate itself can disqualify a `*` operand: BigQuery lowers `COUNT(*)`
```
dlychagin-gg, line 282: just checking, BQ does not allow COUNT(*) measure?

toolbox/mdcode/src/libts/semantic/ir.ts
```diff
@@ -181,6 +184,25 @@ export interface RelationshipEnd {
   columns: string[];     // join columns on this endpoint's table
 }
 
+/**
+ * An association (junction) table backing a many-to-many relationship.
```
amirhormati, line 188: Why does many-to-many need a special table?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+  // backs the edge (one edge row per source row). Its FK columns
+  // (`rel.source.columns`) reference the destination's key columns
+  // (`rel.destination.columns`); the edge is keyed by, and references its
+  // source node through, the source entity's own key (looked up here rather
+  // than duplicated onto the relationship).
+  const sourceEntity = entitiesByName.get(rel.source.entity);
+  let backing: string;
+  let sourceKey: string[];
+  if (!sourceEntity) {
+    warnings.push(`relationship '${rel.name}': unknown source entity '${
```
dlychagin-gg, line 522: why warning? The IR is in an invalid state, should we fail?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+
+  return lines.join('\n');
+}
+
+
+// Renders a many-to-many edge backed by an association (junction) table. Unlike
+// a direct FK, the edge has its OWN backing table and KEY, each endpoint's
+// SOURCE/DESTINATION KEY names the junction columns referencing that entity's
+// declared key, and the junction's own `fields` become edge PROPERTIES.
+function renderAssociationEdge(
```
dlychagin-gg, line 557: OSI doesn't support junction tables in relationships. Therefore this code cannot be tested. Should we remove it?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+// differ (e.g. a graph in `proj.demo` over source tables in `samples.tpch`).
+// Qualifying a source table is the producer's job, not the emitter's. A
+// verbatim query (contains whitespace) cannot back a graph element table, so it
+// is passed through parenthesized with a warning.
+function qualifyTable(
+    dataSource: string, _opts: GenerateOptions, warnings: string[],
+    context: string): string {
+  const trimmed = (dataSource ?? '').trim();
+  if (!trimmed) {
+    warnings.push(
```
dlychagin-gg, line 636: should this be an error?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+        `${context}: empty data source; the table reference will be invalid`);
+    return '``';
+  }
+  if (/\s/.test(trimmed)) {
```
dlychagin-gg, line 640: I think we should remove this test. SQL table identifiers can have spaces in them (with proper quoting). We can assume that the dataSource is a table name.

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+        `emitting it verbatim (a graph element table requires a table)`);
+    return `(${trimmed})`;
+  }
+  return `\`${trimmed}\``;
+}
+
+// Builds the dataset-qualified graph name, filling a missing project/dataset
+// from opts and then from the first entity's qualified dataSource.
+function qualifyGraph(model: SemanticModel, opts: GenerateOptions): string {
+  const name = opts.graphName ?? model.name;
```
dlychagin-gg, line 653: I thought we'd get the fully qualified graph name from the 'deploymentResources' field which will be passed via custom_extensions OSI field. I don't think we're doing it this way here.

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+    lines.map(l => line(depth, l)).join(',\n');
+
+function propertiesBlock(properties: string[]): string {
+  return `${line(2, 'PROPERTIES(')}\n${list(3, properties)}\n${line(2, ')')}`;
+}
+
+
+// Composes the `OPTIONS(description=...)` text for a graph element from its
+// base description and AI-first annotations. BigQuery graph DDL exposes only a
+// single description slot, so instructions, synonyms, and examples are folded
```
dlychagin-gg, line 699: examples could be quite large, do we really need them in the DDL?

toolbox/mdcode/src/libts/semantic/bigquery.ts
```diff
@@ -0,0 +1,754 @@
...
+// which lead the composed text (matching the authored order) ahead of its
+// description and AI-first annotations.
+function fieldDescription(field: Field): string|undefined {
+  return composeText([
+    field.label,
+    isTimeDimension(field) ? 'Time dimension.' : undefined,
+    field.description,
+    field.aiContext?.instructions,
+    synonymsLine(field.aiContext?.synonyms),
+    examplesLine(field.aiContext?.examples),
```
dlychagin-gg, line 721: same comment on having examples in the DDL. are they needed?

toolbox/mdcode/tests/libts/semantic/fixtures/lineitem_databricks_ext.bigquery.golden.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE PROPERTY GRAPH `sqlgen-testing.demo.lineitem`
+NODE TABLES (
```
dlychagin-gg: 1. this graph is invalid, we should not produce it.
2. also, there's no .yaml for this testcase, perhaps we should just remove this .sql file

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.yaml
libei: `fixtures/ossie/tpcds_semantic_model.yaml` is the Apache Ossie reference example, copied verbatim (see `fixtures/ossie/README.md`) — it verifies the loader parses the canonical upstream spec example faithfully and anchors OSI-schema conformance. `tpcds_date_edge.yaml` is hand-authored in the same format, trimmed to isolate generator edge cases: a relationship whose from/to columns share a name, a keyless target dimension (`date_dim`) with only `custom_extensions`, and a second non-BigQuery dialect variant to exercise dialect selection. Added a header note spelling this out.

toolbox/mdcode/src/libts/semantic/ir.ts
libei: A many-to-many link can't be a foreign key: an FK column holds a single value, so it references at most one row — a to-one direction. To record that each student takes many courses and each course has many students, the pairs have to live in their own junction table (one row per (source, destination)), which can also carry edge attributes like a grade. Expanded the `Association` doc comment to state this.

toolbox/mdcode/tests/libts/semantic/fixtures/keyless_dimension.bigquery.golden.sql
```diff
@@ -0,0 +1,14 @@
...
+NODE TABLES (
+  `sqlgen-testing.demo.facts` AS facts
+    KEY(id)
+    PROPERTIES(
+      id,
+      nokey_id
+    )
+);
+
+-- warnings --
```
dlychagin-gg, line 11: at some point, we should probably have an execution mode where we'll fail instead of reporting warnings. A warning for an unsupported dialect is fine, but maybe at some point we should have a mode where we'd fail if we start ignoring entities or metrics. wdyt?

toolbox/mdcode/src/libts/semantic/ir.ts
amirhormati: Makes sense.

amirhormati (approved):

toolbox/mdcode/tests/libts/semantic/fixtures/school_manytomany.bigquery.golden.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE PROPERTY GRAPH `sqlgen-testing.bei_semantic_ir_verify.school_graph`
```
dlychagin-gg, line 1: where is yaml for this golden?

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.bigquery.golden.sql
```diff
@@ -0,0 +1,31 @@
+CREATE OR REPLACE PROPERTY GRAPH `sqlgen-testing.demo.sales`
```
dlychagin-gg, line 1: where is yaml for this golden?

toolbox/mdcode/tests/libts/semantic/fixtures/vendor_dialects.bigquery.golden.sql
```diff
@@ -0,0 +1,49 @@
+CREATE OR REPLACE PROPERTY GRAPH `sqlgen-testing.demo.vendor_sales`
```
dlychagin-gg, line 1: where is yaml for this golden?

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Done — renamed `splitAggregate` → `extractAggregate`.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Done — `placeMetric` now takes the `skipped` set and checks `skipped.has(entityName)` instead of re-deriving the keyless condition.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Yes — verified live on BigQuery: `MEASURE(COUNT(DISTINCT p))` and `MEASURE(SUM(DISTINCT p))` are both accepted (over an exposed node property). The generator already emits the modifier — see the `agg.distinct` branch in `placeMetric`, exercised by the DISTINCT fixtures.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: BigQuery rejects a literal `*` operand inside a MEASURE, so the generator lowers `COUNT(*)` to `COUNT(<key>)` over the node's key property (a keyed node's key is non-null, so the counts match). Verified live. Any other `FN(*)` form (e.g. `SUM(*)`) is unsupported and is skipped + warned rather than emitted.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Changed — the generator now throws instead of warning when no entity can form a node. An empty `NODE TABLES ()` block is invalid DDL under any circumstance, so we fail loudly rather than return a graph BigQuery would reject. The per-entity skip reasons are folded into the error message.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Right. `entities.length === 0` is only reachable via hand-built IR (the loader's schema requires `datasets.min(1)`), but either way the generator now throws when zero valid node tables remain — see the change at the top of `generatePropertyGraph`.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Changed — this now throws. When every entity is skipped for an empty KEY the graph would be empty and invalid, so `generatePropertyGraph` fails with a message that lists which entities were dropped and why.

toolbox/mdcode/tests/libts/semantic/fixtures/lineitem_databricks_ext.bigquery.golden.sql
libei: Agreed on (1) — fixed by making the generator throw on an empty graph, so this DDL can no longer be produced; removed the golden and dropped the fixture from the generator CORPUS. On (2): the `.yaml` does exist (`lineitem_databricks_ext.yaml`) and is exercised by `loader.test.ts` (the `unique_keys`-without-`primary_key` path) plus two loader-behavior tests in the e2e file, so I kept it as a loader fixture rather than deleting it.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: These per-element cases (metric with no expression, metric on an unknown/keyless entity, an edge with an unknown endpoint) drop a single element and leave the rest of the graph valid, so the generator degrades + warns rather than aborting a whole model over one bad metric. Note a metric with no expression is unreachable from a loaded model — the loader's zod schema requires `expression` — so this only guards hand-built IR. Per your comment on `keyless_dimension`, the plan is a strict `validate` gate (already flagged at `loader.ts`) that promotes these warnings to hard errors in a fail-fast mode; I'd rather add that one switch than scatter throws. The genuinely-invalid-output case (empty `NODE TABLES`) is the exception and now throws unconditionally.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Same policy as above: a declared-entity vs expression-qualifier disagreement drops/annotates one metric but keeps the graph valid, so it warns rather than throws. It's also unreachable from a loaded model (the loader derives `metric.entity` from the expression's qualifiers, so they can't disagree) — it only guards hand-built IR. The strict `validate` gate would make it fatal.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Same as the previous two — an unknown-entity reference drops one metric, not the whole graph, so it warns. Unreachable from a loaded model (the metric's entity is derived from the expression against the model's entity set). Would become fatal under the strict `validate` gate.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Same per-element policy: an edge with an unknown source entity is omitted with a warning, not fatal to the rest of the graph. Also unreachable from a loaded model — `convertRelationship` in `loader.ts` already throws on an endpoint not declared in the model — so this is a hand-built-IR guard. Strict `validate` mode would make it an error.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: It's live for hand-built IR, not dead. An author can set `metric.entity` on a `COUNT(*)` metric to attach it — that's exactly this fallback, and it's exercised by the two `COUNT(*)`-lowering tests in `bigquery.test.ts` (removing the branch made them fail). You're right it never fires for a *loaded* model: the loader derives `metric.entity` from the expression's qualifiers, so `COUNT(*)` (no qualifier) leaves `metric.entity` unset. Added a comment stating exactly this.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Kept as a warning, consistent with the per-element degradation policy (one bad table ref shouldn't abort a multi-entity graph). It's only reachable via hand-built IR — the loader's `parseSource` never returns an empty string. The strict `validate` gate would promote it to an error.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: The loader's `parseSource` deliberately passes a source through verbatim when it contains whitespace (a query-as-source), so this guard catches a real, if rare, producer output — and a BigQuery graph element table can't be a subquery, so we warn instead of emitting an invalid ``` `SELECT …` ```. You're right that a backtick-quoted identifier with a space is a legitimate table name that this heuristic would misclassify; the loader has the same limitation today. I'd rather fix both together (drop query-as-source support and treat `dataSource` as always a table) in a follow-up than change only the emitter here. Left as-is for now — let me know if you'd prefer I remove it.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Correct — the fully-qualified graph name will come from the GOOGLE `custom_extensions` deploymentResources block, but that's consumed by the push layer (the deployment-target view derived from the model-level GOOGLE block, per the `SemanticModel.customExtensions` note in `ir.ts`), not this generator. `generatePropertyGraph` stays target-agnostic: it takes the name via `GenerateOptions.graphName` and the push path passes the resolved name in. Wiring push → deploymentResources → `graphName` is #270 (the BigQuery Graph push PR).

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: The graph DDL exposes a single `description` slot per element, so instructions / synonyms / examples are all folded into it; examples are included because they help downstream NL2SQL consumers reading the graph metadata. Fair point on size — I can gate examples (or all AI-context) behind a `GenerateOptions` flag if you'd like. Want me to add one?

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: Same as the property comment above — examples ride in the single element `description` slot. Happy to add a `GenerateOptions` flag to omit examples (or all AI-context) from the DDL if size is a concern.

toolbox/mdcode/tests/libts/semantic/fixtures/keyless_dimension.bigquery.golden.sql
libei: Agreed — that's the strict `validate` gate (already noted at `loader.ts`). Default stays lenient: an unsupported dialect is a warning, and dropping an entity/metric is a warning; in strict mode those become errors. This PR keeps that default but makes the one always-invalid case — an empty graph (no node tables) — a hard failure regardless of mode.

toolbox/mdcode/src/libts/semantic/bigquery.ts
libei: It's covered end-to-end, just not via OSI. `school_manytomany.bigquery.golden.sql` is generated from a hand-built association IR in `bigquery.test.ts`, and the association → edge DDL was validated against live BigQuery. OSI has no junction-table syntax yet, so the loader can't produce one — but the IR and generator support it ahead of a format extension (see the `Association` doc comment in `ir.ts`). I'd keep it; happy to revisit if you'd prefer it wait for the format extension.

toolbox/mdcode/tests/libts/semantic/fixtures/school_manytomany.bigquery.golden.sql
libei: There's no `.yaml` because OSI can't express a junction table — this golden is generated from hand-built association IR in `bigquery.test.ts` (the M:N/association path), then validated against live BigQuery. Same reason as the junction-table thread above.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.bigquery.golden.sql
libei: It's right next to the golden — `star_orders_customer.yaml` — and it's the first entry in the e2e `CORPUS` list, so the golden loop regenerates from it.

toolbox/mdcode/tests/libts/semantic/fixtures/vendor_dialects.bigquery.golden.sql
libei: It exists — `vendor_dialects.yaml`, next to the golden and in the e2e `CORPUS` list.

PR #275, opened 2026-08-07
mdcode: add Knowledge Catalog push for the semantic model

libei:
## What

Adds the **Knowledge Catalog** leg of `kcmd push` for the semantic-model scope (PR4), the counterpart to the BigQuery Graph leg that shipped in #270. Now based directly on merged `main`.

- **`knowledge_catalog.ts`** — a pure IR → catalog-resources emitter (counterpart to `bigquery.ts`). Maps a model to `semantic-model` / `semantic-entity` / `semantic-metric` entries, each carrying its `semantic-*` aspect(s) plus the built-in `schema` aspect for entities. Aspect-data shapes mirror the closed `metadataTemplate`s from go/semantic-model-kc-v2.
- **`deploy_knowledge_catalog.ts`** — the destination orchestrator (counterpart to `deploy_bigquery.ts`): loads docs → emits → ensures the entry group → writes entries anchor-first, upserting on re-push.
- **`push --target <list>`** — selects destination(s): `bq`, `kc`, a comma-separated list (`bq,kc`), or `all`. **Default is `all`** — a bare `kcmd push` deploys to every destination. Destinations run in canonical order (BigQuery first) and fail fast — a BigQuery failure skips the catalog leg. (Designed to scale: more destinations are added by appending to one list, so there is no two-destination `both` keyword.)
- **`push --print`** — prints each pushed destination's generated artifact in that destination's native format — BigQuery Graph → SQL DDL, Knowledge Catalog → the entry plan — each block labeled by destination. (Deliberately format-neutral rather than a BigQuery-specific `--print-ddl`: "DDL" only describes the SQL leg, and each future destination emits its own artifact shape.) Independent of `--validate-only`; scope which destinations run with `--target`.

The BigQuery emitter/publisher were renamed to `bigquery.ts` / `deploy_bigquery.ts` in this PR so the `semantic/` layer reads symmetrically by destination × role:

| | emit (pure IR → artifacts) | publish (side-effecting) |
|---|---|---|
| **BigQuery** | `bigquery.ts` | `deploy_bigquery.ts` |
| **Knowledge Catalog** | `knowledge_catalog.ts` | `deploy_knowledge_catalog.ts` |

## Usage

The destination scope (`<projectId>.<locationId>.<entryGroupId>`) is fixed at `init`; `push` reads it from the snapshot. `push` has two orthogonal knobs: **`--target`** picks *which* destinations run, and **`--validate-only` / `--print`** pick *what to do* with them.

**1. Initialize the scope (once):**

```sh
kcmd init --semantic-model my-proj.us.my_entry_group
```

**2. Deploy** — `--target` selects destinations (default `all`):

```sh
kcmd push                     # all destinations, BigQuery first, fail-fast
kcmd push --target bq         # BigQuery property graph only
kcmd push --target kc         # Knowledge Catalog entries only
kcmd push --target bq,kc      # explicit list (same as the default here)
```

**3. Validate** (`--validate-only`) — writes nothing; surfaces warnings + a success summary. Respects `--target`:

```sh
kcmd push --validate-only                # dry-run every destination
kcmd push --target kc --validate-only    # dry-run the Knowledge Catalog leg only
```

**4. Inspect the generated artifact** (`--print`) — dumps each pushed destination's artifact in its native format (BigQuery → SQL DDL, Knowledge Catalog → the entry plan). Composes with both knobs above:

```sh
kcmd push --validate-only --print        # dry-run + dump every leg's artifact
kcmd push --target bq --print            # real push + print the BigQuery DDL
```

### Flags

| Flag | Applies to | Behavior |
|---|---|---|
| `--target <list>` | semantic-model push | Destination selector, **default `all`**. Accepts `bq` (BigQuery property graph, #270), `kc` (Knowledge Catalog entries, this PR), a comma-separated list (`bq,kc`), or `all` (every registered destination). Tokens are de-duplicated and run in canonical order (BigQuery first), so a BigQuery failure fails fast and later legs are skipped. Any unrecognized token is a hard error. Ignored for non-semantic-model scopes. |
| `--print` | semantic-model push | Prints each pushed destination's generated artifact in its native format — BigQuery Graph → SQL DDL, Knowledge Catalog → the entry plan (destination + each entry id and type). Each block is labeled by destination. Format-neutral by design (not a BigQuery-specific `--print-ddl`), since each destination emits a different artifact shape. Independent of `--validate-only`, and composes with `--target` to pick which destinations' artifact is shown. Ignored for non-semantic-model scopes. |
| `--validate-only` | both legs (KCMD-wide) | Compile and report only; never writes. Surfaces warnings and a success summary. Pair with `--print` to also dump each leg's artifact. An empty workspace is a clean no-op under validate. |
| `--force` | catalog-sync scopes (KCMD-wide) | Existing flag; **not** consumed by the semantic-model legs (a re-push is already an idempotent upsert, so there is nothing to force). |

## Design notes

- The `semantic-model` / `semantic-entity` / `semantic-metric` entry & aspect types (and the extended built-in `schema` aspect) are **built-in system types** in `dataplex-types/global`. Push therefore provisions **only the entry group** — no aspect/entry type creation.
- Each built-in entry type declares `required_aspects`, so the emitted aspect set per entry is a hard contract: model→{semantic-model}, entity→{semantic-entity, schema}, metric→{semantic-metric}.
- The logical `DataType` maps to the schema aspect's required `dataType` / `metadataType`, and each field's `expression` / `importedExpression` / `role` land in the new per-field `semantics` block. A metric with no declared datatype falls back to the required `dataType: STRING` with a warning.
- **Relationships are not published**: there is no user-writable, directed entry link type valid over `semantic-entity` endpoints. The graph edges live in the BigQuery property graph; the emitter warns.
- **Failure handling mirrors the BigQuery leg**: an emit error (e.g. a malformed GOOGLE `custom_extension`) is reported against the model + document rather than crashing the push, and entry-id collisions across models in one push are caught up front (before any write) instead of letting the later entry silently upsert over the earlier.

## Testing

- `knowledge_catalog.e2e.test.ts` — golden `<fixture>.knowledge_catalog.golden.json` for a corpus (deployment targets, temporal types, deferred relationships).
- `knowledge_catalog.test.ts` — emitter unit tests (type/role mapping, resource paths, metric fallback, id-collision guard).
- `deploy_knowledge_catalog.test.ts` — hermetic publisher tests (group-only provisioning, anchor-first writes, upsert, dry-run plan, failure paths, emit-error and cross-model id-collision guards).
- `target.test.ts` — `--target` resolution (list parsing, `all`, canonical ordering, de-duplication, unknown-token rejection).
- `tsc --noEmit` clean; suite green.

**Validated live (BigQuery leg):** `kcmd push --target bq` deployed a property graph over real tables in `sqlgen-testing`, and a GQL traversal + aggregation over the deployed graph returned correct results (per-customer order counts and revenue). This exercises the full `init → author → push → query` path for the BigQuery destination.

## Not yet validated

The **Knowledge Catalog write path** can't be exercised live yet: the built-in `semantic-*` types land with a server-side CL that is still pending, so a KC write needs `dataplex.entryGroups.useSemanticModelAspect` on the target entry group once it submits and rolls out to nonprod. The KC leg is covered hermetically (see Testing). Draft until the KC path is live-validated.


toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+//
+// This is the Knowledge Catalog leg of `kcmd push` for the semantic-model
+// scope, the counterpart to `deploy_bigquery.ts`. It parses each authored
+// Ossie document into the semantic IR (loader), maps it to catalog Entries +
+// Aspects (the pure emitter in knowledge_catalog.ts), and writes them through
+// the Knowledge Catalog client.
+//
+// Types: the `semantic-model`/`semantic-entity`/`semantic-metric` entry and
+// aspect types — and the built-in `schema` aspect — are built-in system types
+// in `dataplex-types/global` (go/semantic-model-kc-v2). Push does NOT provision
```
dlychagin-gg: nit, let's remove go/ link from this line

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+// any type; it only ensures the destination entry group exists and then writes
+// entries. The types are TIER2 `nonprod_only`, so this leg targets a nonprod
```
dlychagin-gg: nit. let's remove type tier mentioning from there.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+// catalog, and the caller needs `dataplex.entryGroups.useSemanticModelAspect`.
+//
+// Publish sequence (mirrors the BigQuery leg's structure):
+//   * Ensure the destination entry group (idempotent; an "already exists" is
+//     success). No aspect/entry type creation — the system types are built-in.
+//   * Create each model's entries in array order: the semantic-model anchor
+//     first (it is the parentEntry of every entity/metric entry), then the
+//     children concurrently.
```
dlychagin-gg: all children concurrently? metrics depend on entities. so we should ensure that an entity is created before a metric.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+//   * A re-push upserts: an entry that already exists is updated in place.
```
dlychagin-gg, line 21: does push also delete KC entries that were deleted from the model? (it should, but ok to do it in a follow up PR)

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+//   * Relationship edges are not published (no writable directed link type over
+//     semantic-entity endpoints); the emitter warns and the edges live in the
```
dlychagin-gg: why EntryLinks are not pushed to KC?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+
+export interface KcDeployOptions {
+  // The resolved Knowledge Catalog destination (flag overrides applied over the
+  // catalog.yaml scope defaults).
+  project: string;
+  location: string;
+  entryGroup: string;
+  // Where the built-in `semantic-*` / `schema` system types are referenced.
+  // Default: `dataplex-types` / `global`. Overridable for a staging project
+  // during the nonprod-only window; the emitted entries are otherwise
```
dlychagin-gg: nit. let's remove mentioning of 'nonprod' here

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+  // a generation warning surfaces even if a later write fails.
+  const emitted: {model: string; resources: KcResources}[] = [];
+  for (const doc of docs) {
+    // A document that fails to parse (or violates the model schema) is an
+    // authoring error; report it against the specific document rather than
+    // letting the loader's exception propagate as an uncaught stack trace.
+    let loaded;
+    try {
+      loaded =
+          loadModels(doc.text, {defaultProject: defaultProject ?? ctx.project});
```
dlychagin-gg: I think loading of models should be shared among deploy_kc and deploy_bq. It's the same model, we should load/validate it once and then deploy into both places. Seems like this code will reload it here, which is not ideal

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+    updated += outcome.updated;
+  }
+
+  return {success: true, warnings, created, updated, plan};
+}
+
+
+// Ensures the destination entry group exists. Returns an error message on a
+// non-idempotent failure, or undefined on success (created or already existed).
+async function ensureEntryGroup(
```
dlychagin-gg: does the standard layout for KCMD also create entry group on push? Just checking. We should be aligned with how the standard layout operates. If it does not create the entry group on push the this code should not do it either.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -0,0 +1,337 @@
...
+
+function isOk(res: {status: number}): boolean {
+  return res.status === 200;
+}
+
+// A create that failed because the resource already exists — treated as success
+// for idempotent provisioning and re-push.
+function isExists(res: {status: number; message?: string}): boolean {
+  return res.status === 409 ||
+      /already exists|alreadyexists/i.test(res.message ?? '');
```
dlychagin-gg: do we really need to match the error message? Is there no better way of checking this?

toolbox/mdcode/src/libts/semantic/knowledge_catalog.ts
```diff
@@ -0,0 +1,402 @@
...
+    default:
+      return 'STRING';
+  }
+}
+
+// Maps the IR's opaque, canonical `dataSource` to a catalog resource string. A
+// clean three-part `project.dataset.table` becomes the BigQuery linked-resource
+// URI; anything else (a query, an under/over-qualified ref) is passed through
+// verbatim so nothing is lost.
+function resourcePath(dataSource: string): string {
```
dlychagin-gg, line 380: nit, as we discussed at some point we should figure out how it'll look like for iceberg tables

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,87 @@
...
+      "entrySource": {
+        "displayName": "total_revenue"
+      },
+      "aspects": {
+        "dataplex-types.global.semantic-metric": {
+          "aspectType": "projects/dataplex-types/locations/global/aspectTypes/semantic-metric",
+          "data": {
+            "entity": "orders",
+            "dataType": "STRING",
+            "expression": "SUM(orders.o_totalprice)"
```
dlychagin-gg, line 78: `orders.` probably needs to be removed. However if BQ allows it in the measure expression, then it's fine to keep as is

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,87 @@
...
+          }
+        }
+      }
+    }
+  ],
+  "warnings": [
+    "metric 'total_revenue': no datatype in the source model; defaulting the required semantic-metric.dataType to 'STRING'"
```
dlychagin-gg: for metrics we should default dataType to number if it's not specified in the model. for dimensions default to string

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,171 @@
...
+      "name": "projects/sqlgen-testing/locations/us/entryGroups/semantic/entries/sales",
+      "entryType": "projects/dataplex-types/locations/global/entryTypes/semantic-model",
+      "entrySource": {
+        "displayName": "sales",
+        "description": "Sales orders with customer attributes"
+      },
+      "aspects": {
+        "dataplex-types.global.semantic-model": {
+          "aspectType": "projects/dataplex-types/locations/global/aspectTypes/semantic-model",
+          "data": {}
```
dlychagin-gg: why no deploymentTargets here? I think we should require deploymentTargets for all models. otherwise the BQ graph won't be deployed and the model won't be executable.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,171 @@
...
+      "parentEntry": "projects/sqlgen-testing/locations/us/entryGroups/semantic/entries/sales",
+      "entrySource": {
+        "displayName": "order_count",
+        "description": "Number of orders"
+      },
+      "aspects": {
+        "dataplex-types.global.semantic-metric": {
+          "aspectType": "projects/dataplex-types/locations/global/aspectTypes/semantic-metric",
+          "data": {
+            "dataType": "STRING",
```
dlychagin-gg: "entity": should always be provided for models targeting BQ graphs

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,171 @@
...
+            "expression": "COUNT(*)"
```
dlychagin-gg: I thought you said the measure expression cannot be COUNT(*) in BQ

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
```diff
@@ -0,0 +1,357 @@
...
+          "aspectType": "projects/dataplex-types/locations/global/aspectTypes/schema",
+          "data": {
+            "fields": [
+              {
+                "name": "ss_item_sk",
+                "dataType": "STRING",
+                "metadataType": "STRING",
+                "semantics": {
+                  "expression": "ss_item_sk",
+                  "importedExpression": "{label/store_sales.attr.store_sales.ss_item_sk}",
```
dlychagin-gg, line 46: why is it here?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done — dropped the go/ link (and the type-tier/nonprod wording flagged below).

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done — removed the TIER2/nonprod_only mention; kept only the `dataplex.entryGroups.useSemanticModelAspect` permission note.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Good catch. `createEntries` now writes the entity wave before the metric wave (a metric aspect references its entity by name); within each wave the writes stay concurrent.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Simplified — the API layer preserves the HTTP status (`gcp/api.ts`), so `isExists` now keys off the 409 ALREADY_EXISTS status instead of matching the message text. (`isPropagating` still inspects the message, because a bare 404 is ambiguous: entry-group-not-visible-yet vs. a genuinely missing entry type.)

toolbox/mdcode/src/libts/semantic/knowledge_catalog.ts
libei: Added a TODO to settle the linked-resource URI shape for Iceberg / BigLake tables. Today anything that isn't a clean three-part managed-table ref is passed through verbatim.

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
libei: The `orders.` qualifier is what drives the loader's entity inference (it's how `entity: orders` gets attached to the metric). The BQ leg strips it when lowering to a `MEASURE` — verified live it emits bare `MEASURE(SUM(o_totalprice))` and the graph queries correctly. So it's not valid inside a BQ measure, but it's load-bearing in the source; happy to also strip it from the emitted KC expression for consistency if you'd like.

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
libei: Done — an undeclared metric `dataType` now defaults to a numeric type (`FLOAT64`); dimensions keep the `STRING` default. Note the metric `dataType` field is a GoogleSQL type (e.g. Decimal maps to `NUMERIC`), so I used `FLOAT64` as the generic numeric default — say the word if you'd rather INT64/NUMERIC.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Added a GOOGLE deployment target to this fixture so it's BQ-executable. On enforcement: I can add a validate-time check that a model declaring a BQ-graph target has the pieces needed to deploy — want that in this PR or a follow-up?

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Fixed — `order_count` is now entity-scoped (`COUNT(orders.o_orderkey)`), so `entity: orders` is inferred and it deploys as a real MEASURE.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Right — a bare `COUNT(*)` with no entity can't be a single MEASURE (the BQ leg was skipping it). Changed it to `COUNT(orders.o_orderkey)`; verified it deploys and returns correct counts on a live graph.

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
libei: That's the retained vendor (MAQL) form of the field expression: the source declares both an ANSI_SQL and a MAQL dialect, and the loader keeps the non-target dialect as `importedExpression` for downstream transpilation. It rides along in the schema aspect's `semantics` block. I can drop it from the KC emit if you'd rather not surface the vendor form there.

toolbox/mdcode/src/libts/semantic/loader.ts
```diff
@@ -478,3 +478,53 @@ function parseSource(source: string, opts: LoadOptions,
 function unquote(part: string): string {
   return part.replace(/^[`"]/, '').replace(/[`"]$/, '');
 }
+
+
+// A single model parsed from a named document, tagged with that document name
+// so a consumer (a deploy leg) can attribute warnings and errors back to the
+// file the author wrote.
+export interface LoadedModel {
+  document: string;       // originating document name
```
libei: We need to be clear what this "originating document" mean here?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in this PR (2478d40). Added `loadSemanticModels()` in `loader.ts` which parses + validates each document once into a doc-tagged `LoadedModel[]`. Both `deployBigQuery` and `deployKnowledgeCatalog` now take the pre-loaded models (dropped their `docs`/`defaultProject` params), and `push()` loads once, prints load warnings once, and fans out. A `--target all` push no longer re-parses each document per leg.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in this PR (2478d40). After writing a model's entries the KC leg lists the entry group and deletes entries it owns but did not re-emit — an entity or metric removed from a still-present model. Ownership is scoped by entry id (the anchor id, or its `<model>.entities.` / `<model>.metrics.` child namespace), so unrelated entries in a shared group are never touched; a 404 is tolerated (already gone); skipped under `--validate-only` (stays offline). Whole-model removal (deleting a model's document) is left as a TODO — it needs a scope-level record of which models the entry group manages to reconcile safely.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Good catch — checked, and no: the standard layout's push (`sync.ts`) creates/updates entries only, never the entry group. Aligned in 254f286: dropped `ensureEntryGroup` from the KC push (it now writes entries only), and moved entry-group creation to `init --semantic-model` (idempotent — a 409 already-exists is success, other failures fail init). A missing group on push now surfaces as a clear entry-create error.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Good call — they're published now (066b5de). Each direct-FK relationship becomes a `schema-join` entry link between its two endpoint entity entries, with a `schema-join` aspect recording the join (source=FK side, target=key side, positionally-paired columns, `type: FOREIGN_KEY`, `inferenceSource: USER`, `userManaged: true`).

`schema-join` is the fit: it's user-writable over any entry type (`allow_all_types: true`) and its aspect mirrors the IR `Relationship` almost 1:1. The link itself is undirected (both refs `UNSPECIFIED`); direction is carried in the aspect (source is the FK side). Written after the model's entries so both endpoints exist first; a re-push upserts the aspect on 409.

Deferred and documented (warn / TODO): M:N (association/junction) edges — a junction is two joins, which the single source/target schema-join doesn't model; `represents` logical→physical links; and removed-link reconciliation (there's no list API for entry links, only `LookupEntryLinks` per referenced entry).

Requires `dataplex.entryGroups.useSchemaJoinEntryLink` + `useSchemaJoinAspect` in the target env. Covered hermetically (emitter goldens + publisher write-path tests); the live KC run is still gated on the semantic types (CL pending), same as the rest of this leg.

toolbox/mdcode/src/libts/semantic/loader.ts
libei: Clarified in `4247f97`. "Originating document" = the model file this model was parsed from: the `.yaml` basename the layout discovered (e.g. `sales` for `sales.yaml`), **not** a filesystem path. It comes from `SemanticModelLayout.modelDocuments()`, which keys each doc by `path.basename(localPath, '.yaml')`.

It's used for exactly one thing: attributing loader warnings/errors back to the author's file (warnings get a `[<document>]` prefix; a parse/schema error is reported as `Model document '<document>': ...`). It's not part of the deployed IR. Updated the field comment to say this.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
dlychagin-gg: sg. we should think about whole model removal lifecycle at some point. A single entry group can only contain 1 model. I think maybe kcmd push should fail if the entry group already has a model with a different name and then add "-force-remove" flag to kcmd push so user could tell it explicitly that the other model should be dropped. i.e. 
> kcmd push 
--> failure: "another model already exists in the entry group"
> kcmd push --force-remove 
--> success. (previous model removed, new model pushed)

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
dlychagin-gg: 1/ sg. yes, schema-join is the right approach. 
2/ yes, M:N relationships are out of scope for now. 
3/ Let's double check that entryLinks are also dropped when relationships removed from the model. 
4/ Let's double check that entryLinks are created in a wave after the entity wave.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
dlychagin-gg: sg!

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
dlychagin-gg: I need to think a bit more about it. Let's keep it as is for now

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
dlychagin-gg: I'd rather use NUMERIC (i.e use decimal instead of float) (can be fixed in a follow up PR)

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
dlychagin-gg: follow up is fine. but yes, let's add some validation logic

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
dlychagin-gg: let's drop it. (in a follow up PR). the idea was that `importedExpression` is in vendor dialect, while `expression` is in our dialect (or ANSI). We have no use for non-SQL expressions.

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
libei: Done in `ad37485` — the un-typed metric `dataType` fallback is now `NUMERIC` (exact decimal) instead of `FLOAT64`; the dimension default (`STRING`) is unchanged. Goldens regenerated, and the focused fallback test updated. (`Decimal` -> `NUMERIC` was already covered by the type-mapping test.)

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: On 3/ and 4/:

**4/ (links written after the entity wave):** confirmed. Per model, `deployKnowledgeCatalog` runs `createEntries` fully before `createEntryLinks` (deploy_knowledge_catalog.ts:187-202), and `createEntries` itself writes anchor -> entity wave -> metric wave (deploy_knowledge_catalog.ts:310-322). So by the time any link is written, both endpoint entity entries exist.

**3/ (links dropped when a relationship is removed):** honest answer — **not yet**. `reconcileDeletions` currently reconciles entries only; a relationship removed from the model leaves its `schema-join` link behind. This is the documented TODO (deploy_knowledge_catalog.ts:237-240). The blocker is that there's no list API for entry links — only `LookupEntryLinks` per referenced entry — so the sweep has to be lookup-per-surviving-endpoint (find links referencing each owned entity, delete the owned ones not re-emitted) rather than a group list like the entry reconcile. I didn't want to add an unverifiable `LookupEntryLinks`/`deleteEntryLink` client path on the CL-gated (hermetic-only) leg in this PR. I'll do it in the same follow-up as the whole-model-removal reconcile lifecycle (thread above), where the link sweep fits naturally. Flagging explicitly so it's not a silent gap.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Agreed — that's the right lifecycle model. Since an entry group holds exactly one model, `kcmd push` should fail when the group already contains a *different* model (detectable by a foreign `semantic-model` anchor entry), and `--force-remove` would drop the old model's entries + links before writing the new one. That's a superset of the whole-model-removal reconcile TODO already in the code (deploy_knowledge_catalog.ts:232-235), and it's also where the removed-relationship link sweep (your point 3/ on the EntryLinks thread) belongs. I'll carry all three in the reconcile-lifecycle follow-up rather than expand this (CL-gated) PR further.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Sounds good — I'll add the validation in the follow-up. Plan: at validate time, require a model that declares any BigQuery-graph deployment target to carry a non-empty `deploymentTargets`, and (your other comment) require an `entity` on metrics targeting a BQ graph, so an un-executable model fails `--validate-only` up front rather than deploying a graph that can't run. Keeping it out of this PR per your "follow up is fine".

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
libei: Sounds good, leaving the `orders.` qualifier as-is for now. It's what drives the loader's entity inference (how `entity: orders` attaches to the metric); the BQ leg strips it when lowering to a MEASURE, so it doesn't reach the executed SQL. Happy to revisit the KC-emit representation whenever you've settled on the shape.

dlychagin-gg:
LGTM overall

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
libei: Got it — I'll drop `importedExpression` from the KC emit in the follow-up. Agreed on the contract: `expression` is our/ANSI dialect and is the only form the catalog needs; `importedExpression` is the vendor (e.g. MAQL) source form, which belongs upstream in the loader/transpile path, not in the emitted schema aspect. I'll strip it from both the entity `schema` aspect and the `semantic-metric` aspect so no non-SQL expression is surfaced in KC.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in the follow-up PR #278. Push now snapshots the entry group once before writing; a semantic-model anchor already in the group that this push does not re-emit (a removed/renamed model) is a hard error, and `--force-remove` deletes that model's links and entries before writing.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in the follow-up PR #278. After a model's current schema-join links are written, any link it owns (both endpoints under its `<anchor>.entities.` namespace) that it no longer emits is deleted — covering dropped/renamed relationships. Links are looked up per entity via the new `lookupEntryLinks` client method and deduped across endpoints.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Done in the follow-up PR #278. A shared `validatePushRequirements` gate now fails any push (including `--validate-only`) where a model declares no deployment target.

toolbox/mdcode/tests/libts/semantic/fixtures/star_orders_customer.knowledge_catalog.golden.json
libei: Done in the follow-up PR #278. The same `validatePushRequirements` gate fails the push when a model targeting a BigQuery graph has a metric that does not resolve to exactly one entity (promotes the prior silent warn+skip to a hard error).

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
libei: Done in the follow-up PR #278. `importedExpression` is no longer written into the `semantic-metric` or `schema` aspects; only `expression` (ANSI/GoogleSQL) is emitted. Golden regenerated.

toolbox/mdcode/tests/libts/semantic/fixtures/sales_bq_graph_target.knowledge_catalog.golden.json
dlychagin-gg: @libei , I think we should have the original Ossie expression in the 'importedExpression' field. The 'expression' is supposed to be in GoogleSQL dialect and be the exact expression that'll go into CREATE PG DDL for the owning measure. We can keep 'expression' unset for now, since 'importedExpression' should be sufficient to create prop. graph and for kcmd pull. Let's switch to using 'importedExpression' in a follow up PR.

PR #277, opened 2026-08-08
mdcode: add Knowledge Catalog pull for the semantic model

libei:
Follows **PR4** (the KC push leg, #275, now merged). Rebased onto the merged `main`, so this PR's diff is the pull delta only.

## What

The read counterpart of the Knowledge Catalog push leg. `kcmd pull` in a semantic-model workspace now reads the `semantic-*` entries back from Knowledge Catalog, reconstructs the IR, and serializes each model to `catalog/EntryGroups/<entryGroup>/<model>.yaml` — the inverse of the push leg against the same built-in-type schema (go/semantic-model-kc-v2).

## Pieces

- **`serialize.ts`** (new): pure IR → open-format YAML, the exact inverse of `loader.ts`.
- **`knowledge_catalog.ts`**: `modelsFromCatalogResources`, the inverse of the emitter — parses the built-in `schema` aspect (dataType/metadataType → logical type, `DIMENSION` role, per-field `semantics`), reverses the BigQuery resource URI back to the `dataSource` string, and re-derives each metric's attach entity from its expression (as the loader does).
- **`deploy_knowledge_catalog.ts`**: `pullKnowledgeCatalog` — lists the entry group, hydrates each entry's aspects (an entity needs BOTH its `semantic-entity` aspect and the built-in `schema` aspect), and applies a `--model` filter (scoped to the target model's entries before hydration).
- **`SemanticModelLayout`**: `modelPath` / `hasModel` / `writeModelDocument` write sink.
- **`commands.pull` / `main.ts`**: `--dry-run` and `--model`. Overwrite policy matches the core pull: last-write-wins; local-only documents are never deleted.

## Fidelity

IR-level, bounded by what the push leg persists. Entity keys, `ai_context`, field labels, `importedDialect`, and relationships are **not** written by push (the graph edges live in the BigQuery property graph) and so do not come back. An authored `String` datatype is indistinguishable from an un-typed field after emit and reads back as un-typed. A typeless metric round-trips as `Decimal`: the push leg makes `semantic-metric.dataType` required and defaults it to `NUMERIC`, which the reader maps back to `Decimal`.

## Testing

Hermetic round trips are the acceptance bar (no live KC needed — PR4's server types are still nonprod-pending):

- `loader ↔ serialize` round trip over 6 corpus fixtures (relationships, ai_context, datatypes, dimensions, custom_extensions, vendor dialects).
- `emitter ↔ reader` round trip + dataType-inverse / role / resource-URI / metric-attach unit tests.
- `pullKnowledgeCatalog` over a stubbed catalog client (aspect hydration, `--model` scoping, skipped entries, foreign-entry ignore).
- `SemanticModelLayout` write-path test.

Full `test:semantic` + `test:libts` green; `tsc --noEmit` clean.


toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -460,3 +462,184 @@ function isPropagating(res: {message?: string}): boolean {
...
+// data, so each entry is re-fetched with its aspect types -- an entity needs
+// BOTH its `semantic-entity` aspect and the built-in `schema` aspect), and
+// hands the hydrated entries to the pure reader (modelsFromCatalogResources).
+// ---------------------------------------------------------------------------
+
+export interface KcPullOptions {
+  project: string;
+  location: string;
+  entryGroup: string;
+  model?: string;  // limit to a single model by name (default: all)
```
libei: Can there be multiple models in your entry group?

toolbox/mdcode/src/libts/semantic/serialize.ts
```diff
@@ -0,0 +1,256 @@
+// Serializes the Semantic Model IR (./ir) back to the open AI-first semantics
```
libei: The file name is a little bit too general. It is named as serialize, but serialize from what to what? Can we be clear?

toolbox/mdcode/src/libts/semantic/knowledge_catalog.ts
```diff
@@ -521,3 +522,269 @@ function linkSlug(s: string): string {
 function unquote(part: string): string {
   return part.replace(/^[`"]/, '').replace(/[`"]$/, '');
 }
+
+
+// ---------------------------------------------------------------------------
```
libei: I remember there are some files like load_knowledge_catalog or something of this sort. So, should we have a separate file for this new capability? Or if we put them in the same file, is it going to be too crowded? Please think through. We need to make sure the code is very clean.

toolbox/mdcode/tests/libts/semantic/serialize.test.ts
```diff
@@ -0,0 +1,290 @@
+// Behavior specification for the semantic-model serializer
+// (src/libts/semantic/serialize.ts).
```
libei: Please think through the tests. Usually, I prefer to have a test fixture so that you know the input and output files as a whole, not like validated one row at a time. You don't get the full picture if that's the case. Can you think through all these tests in this PR and organize in a better way?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Yes — one entry group can hold many models. Each `semantic-model` entry is a separate anchor; the reader groups the `semantic-entity`/`semantic-metric` entries under their anchor by `parentEntry` and returns one reconstructed model per anchor (`modelsFromCatalogResources`). `--model` narrows both the fetch and the write to a single anchor. Now documented in the new Pull section of `docs/semantic-model.md`.

toolbox/mdcode/src/libts/semantic/serialize.ts
libei: Renamed to `osi_converter.ts` — the OSI ⇄ IR converter. It currently holds only the serialize direction (IR → OSI YAML); a header banner spells out the direction and notes the loader (OSI → IR) migrates in once #278 merges, at which point it becomes the full two-way converter. (2f0531b)

toolbox/mdcode/src/libts/semantic/knowledge_catalog.ts
libei: Agreed — pulled the new code into its own files: the KC reader is now `kc_converter.ts` and the network pull is `pull_kc.ts`. `knowledge_catalog.ts` and `deploy_knowledge_catalog.ts` return to their #278 state (emit-only / push-only). This is scaffolding for the eventual two-layer split — pure converters (`osi_converter`/`kc_converter`/`bigquery`) vs push/pull orchestration — with the remaining halves moving in after #278 merges and no further renames. (2f0531b)

toolbox/mdcode/tests/libts/semantic/serialize.test.ts
libei: Reorganized around committed golden files. Each corpus fixture now has an `.osi.golden.yaml` (IR → OSI) and a `.pull.golden.yaml` (KC entries → IR → OSI), so you see the whole input and output as files rather than row-by-row asserts — and diffing the two shows exactly what a Knowledge Catalog round trip drops (keys, ai_context, labels, relationships; an `is_time` dimension collapses to a bare `{}`). The IR round-trip and the focused mapping/warning tests are kept as invariants alongside the goldens. Test files renamed to match their modules: `osi_converter`/`kc_converter`/`pull_kc.test.ts`. (2f0531b)

libei:
Added `TODO(#278)` blocks to the new converter/orchestration files (988280b) so the scaffold plan is legible in the code:

- **`osi_converter.ts`** — fold `loader.ts` (OSI read) in, delete `loader.ts`, repoint importers.
- **`kc_converter.ts`** — fold `generateCatalogResources` (KC write) in, delete `knowledge_catalog.ts`, repoint importers, demote the shared `idOf` to a local.
- **`pull_kc.ts`** — rename `deploy_knowledge_catalog.ts` → `push_kc.ts` for `push_kc`/`pull_kc` symmetry (rename only, no logic moves).

Each is a mechanical move + import repoint deferred until #278 merges, so this PR keeps the old push files at their #278 state and only introduces the new-pattern files.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+kcmd pull
+```
+
+Pull reads only from Knowledge Catalog (never BigQuery). Its coordinates come
+from the same scope you authored under (`<projectId>.<locationId>.<entryGroupId>`).
+
+| Flag | Effect |
+|------|--------|
+| `--dry-run` | Reconstruct from the catalog and report what would be written, but write no files. |
+| `--model <name>` | Pull a single model by name; other models in the entry group are left alone. |
```
dlychagin-gg, line 261: for now we assume there's only on model in the entry group. We don't need --model flag

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+
+One entry group can hold **many models** — each `semantic-model` entry is a
+separate anchor, and pull reconstructs one document per anchor. `--model`
+narrows both the fetch and the write to a single anchor.
+
+Pull writes with the same last-write-wins policy as the core pull: a model that
+already exists locally is overwritten in place, and a local-only document (one
+with no matching catalog entry) is left untouched — pull never deletes.
```
dlychagin-gg, line 269: "pull never deletes" --- can we fail pull by default if the id of the model available in the entryGroup that we're pulling from is different from the id of the model available locally on disk? if user wants to proceed then they'll need to explicitly use the new --force-remove flag we already support in push. If force-remove is specified then pull will erase the existing model and will pull the model from the KC.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> - Model structure: the model, its entities, and each entity's fields.
+> - Field data source and data type; the field expression.
+> - Metrics: name, expression, data type, and attach entity.
+> - 1:1 / 1:N relationships: endpoints, foreign-key direction, and join columns
+>   (from the `schema-join` links).
+> - Deployment targets.
+>
+> **Recovered, but normalized** — the content survives, the form changes:
+> - Relationship *names* come back lowercased/hyphenated (the catalog stores the
+>   name only in the link id, e.g. `Places Order` → `places-order`).
```
dlychagin-gg, line 295: if we change relationship name on push then kcmd push should warn user about name change.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> - A field marked as a dimension comes back as a bare `dimension: {}` marker,
+>   without its detail (`is_time`, and so on).
+> - A metric authored with no data type comes back as an explicit `Decimal`
```
dlychagin-gg, line 296: I think we should use 'Opaque' type instead (for both fields and metrics) if type is not provided by the user.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+>   (push must write a type, and defaults it to `NUMERIC`).
+>
+> **Not recovered** — push never wrote these, so pull cannot return them:
+> - Entity keys / unique keys.
```
dlychagin-gg, line 300: pk/unique keys should be persisted in the schema aspect. let's add their support in a follow up PR.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> - `ai_context`.
```
dlychagin-gg, line 301: this should go to the guidelines aspect. see design doc.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> - Field labels.
```
dlychagin-gg, line 302: schema aspect support 'annotations' map for a field. perhaps we should put label there: `{ <label>: "<label>" }`

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> - The original vendor SQL (`importedExpression`).
+> - M:N relationships (the edge lives only in the BigQuery property graph).
```
dlychagin-gg, line 304: nit. let's remove M:N, since Ossie doesn't have those yet

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+> **Recovered exactly** — these come back as authored:
+> - Model structure: the model, its entities, and each entity's fields.
+> - Field data source and data type; the field expression.
+> - Metrics: name, expression, data type, and attach entity.
+> - 1:1 / 1:N relationships: endpoints, foreign-key direction, and join columns
+>   (from the `schema-join` links).
+> - Deployment targets.
+>
+> **Recovered, but normalized** — the content survives, the form changes:
+> - Relationship *names* come back lowercased/hyphenated (the catalog stores the
```
dlychagin-gg, line 294: we should also mention the order of the items emitted by pull. Field order within each entity will be preserved, but the order of datasets, metrics won't be. Let's also say that any comments in the original Yaml file are not preserved either.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,312 @@
...
+
+* `bigquery.jobs.create` in the deployment-target project — to run the deploy's
+  `CREATE OR REPLACE PROPERTY GRAPH` and the validation dry-run query
+* read access on each entity's source table, so the dry-run can resolve it
+* `bigquery.datasets.get` on the target dataset (region detection; optional —
+  push degrades gracefully without it)
+
+**Knowledge Catalog / Dataplex** — for `--target kc` or `all`:
+
+* `dataplex.entryGroups.useSemanticModelAspect` on the destination entry group
```
dlychagin-gg, line 342: I think dataplex.entryGroups.useSemanticEntityAspect and dataplex.entryGroups.useSemanticMetricAspect will also be required.

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+  }
+
+  const fetched = await mapConcurrent(
+      scoped, HYDRATE_CONCURRENCY, async ({entry, aspectTypes}) => {
+        const res = await cat.lookupEntry(
+            opts.project, opts.location, entry.name, aspectTypes);
+        if (res.status !== 200 || !res.result) {
+          return {
+            warning: `failed to fetch entry '${entry.name}' (status ${
```
dlychagin-gg, line 81: should we fail here? (if we could not fetch an entry of the semantic model)

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+  // and dedup (linkDedupKey) -- schema-join is undirected, so each link comes
+  // back once from each of its two endpoints.
+  const entityEntries = hydrated.filter(
+      e => e.entryType?.endsWith('/entryTypes/semantic-entity'));
+  const linkResults =
+      await mapConcurrent(entityEntries, HYDRATE_CONCURRENCY, async entry => {
+        const linkType = schemaJoinLinkType(entry.entryType);
+        const res = await cat.lookupEntryLinks(opts.project, opts.location, {
+          entry: entry.name,
+          entryLinkTypes: linkType ? [linkType] : undefined,
```
dlychagin-gg, line 108: why fallback to undefined here? The linkType is always schema-join. If schemaJoinLinkType() returns undefined then it's an error and kcmd pull should fail.

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+        });
+        if (res.status !== 200 || !res.result) {
+          return {
+            warning: `failed to fetch entry links for '${entry.name}' (status ${
```
dlychagin-gg, line 112: hm. do we differentiate between legitimate no entry links case and failure during fetch? If entity has no relationships then it's ok (no warning needed), but if there's a fetch error then we should fail pull here. Let's not ignore real errors.

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+  }
+
+  const read = modelsFromCatalogResources(hydrated, entryLinks);
+  warnings.push(...read.warnings);
+
+  // Defense in depth: keep only the requested model even if the reader surfaced
+  // another anchor (e.g. a child whose parentEntry pointed outside the scope).
+  let models = read.models;
+  if (opts.model) {
```
dlychagin-gg, line 140: we should assume there's only 1 model in the entryGroup. if there's more than one kcmd pull print their ids and should fail. (bad state on the KC side)

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+
+
+// The schema-join entry link type resource, derived from an entity's entryType
+// base (the link type is the parallel resource in the same project/location the
+// emitter referenced). Used to filter :lookupEntryLinks to just the
+// relationship links. Returns undefined for an entryType with no recognizable
+// base.
+function schemaJoinLinkType(entryType: string): string|undefined {
```
dlychagin-gg, line 182: I thought schema-join link type has a fixed name/location. Why do we compute it here? In any case, this method should never return undefined

toolbox/mdcode/src/libts/semantic/pull_kc.ts
```diff
@@ -0,0 +1,241 @@
...
+    targets: {entry: Entry; aspectTypes: string[]}[],
+    model: string): {entry: Entry; aspectTypes: string[]}[] {
+  const isAnchor = (t: {entry: Entry}) =>
+      !!t.entry.entryType?.endsWith('/entryTypes/semantic-model');
+  const allAnchorNames =
+      new Set(targets.filter(isAnchor).map(t => t.entry.name));
+  const matchedAnchorNames =
+      new Set(targets.filter(isAnchor)
+                  .filter(
+                      t => (t.entry.entrySource?.displayName ??
```
dlychagin-gg, line 205: why are we using displayName here? can you elaborate?

dlychagin-gg:
LGTM

libei:
Thanks for the review. Since #277 is already merged, I've addressed these in follow-up **#298**. Mapping:

- One model per entry group / drop `--model`; >1 anchor is a hard error naming both (`deploy_knowledge_catalog.ts:483`, docs `:261`, `pull_kc.ts:140`) → #298
- Fail pull on a local↔catalog model-id mismatch; add `--force-remove` (docs `:269`) → #298
- Warn on push when a relationship name will be normalized (docs `:295`) → #298
- Default untyped fields **and** metrics to `Opaque` (docs `:296`) → #298
- Fail on any entry-hydration fetch error (`pull_kc.ts:81`) → #298
- schema-join link type is a fixed built-in constant, never `undefined` (`pull_kc.ts:108`, `:182`) → #298
- Distinguish "no links" (silent) from a link-fetch failure (fail) (`pull_kc.ts:112`) → #298
- Remove M:N from the docs (docs `:304`) → #298
- Document pull emit order + that comments aren't preserved (docs `:294`) → #298
- Add `useSemanticEntityAspect` / `useSemanticMetricAspect` to required permissions (docs `:342`) → #298

The three schema-aspect enrichments (pk/unique keys, `ai_context`→guidelines, field label→annotations) are tracked as follow-ups in the #298 description, and the `displayName` question is answered in-thread.

toolbox/mdcode/src/libts/semantic/pull_kc.ts
libei: push writes each element's authored name into the entry's `entrySource.displayName` (the human-facing field); the entry **id** is a normalized slug. So the reader recovers the original-case name from `displayName`, falling back to the id slug only when it's absent (`kc_converter.ts:99` for the anchor, `:152` for entities). That's also why entity/model/metric names round-trip exactly, while relationship names come back lowercased/hyphenated — a schema-join link carries no `displayName`, so the name survives only in the link id.

toolbox/mdcode/docs/semantic-model.md
libei: Agreed — tracked as a follow-up (persist pk/unique keys in the schema aspect). Listed under "Deferred" in #298; not in this follow-up.

toolbox/mdcode/docs/semantic-model.md
libei: Agreed — tracked as a follow-up (`ai_context` → guidelines aspect, per the design doc). Listed under "Deferred" in #298.

toolbox/mdcode/docs/semantic-model.md
libei: Agreed — tracked as a follow-up (field label → schema aspect `annotations` map, `{ <label>: "<label>" }`). Listed under "Deferred" in #298.

PR #278, opened 2026-08-08
mdcode: KC push follow-ups (validate gate, model + link reconciliation)

libei:
Follow-ups to #275 (Knowledge Catalog push for the semantic model), addressing
the four review threads left as agreed follow-ups. Each is a self-contained
commit.

### 1. Drop `importedExpression` from the KC emit
`expression` is the ANSI/GoogleSQL form KC consumes; `importedExpression` is the
vendor/MAQL form and has no KC consumer, so it is no longer written into
`semantic-metric` or `schema` aspects. The vendor form is still kept in the
source model and used as a BigQuery-SQL fallback; only the KC emit drops it.
Golden regenerated.

### 2. Shared validate gate
A single `validatePushRequirements` gate now funnels every push (real and
`--validate-only`):
- every model must declare at least one deployment target;
- a metric on a BigQuery-graph-targeting model must resolve to exactly one
  entity (promotes today's silent warn+skip to a hard error).

Malformed GOOGLE extension JSON becomes a per-model error rather than an uncaught
throw.

### 3. Whole-model removal lifecycle + `--force-remove`
Push now snapshots the destination entry group once before writing. A
`semantic-model` anchor already in the group that this push does not re-emit
(a removed or renamed model) is a hard error, unless `--force-remove` authorizes
deleting its links and entries first.

### 4. Removed-relationship link reconciliation
After a model's current schema-join links are written, any link it owns (both
endpoints under its entities namespace) that it no longer emits is deleted,
covering dropped or renamed relationships. Links are looked up per entity via the
new `lookupEntryLinks` client method and deduped. Deletion reconciliation now
consumes the pre-write snapshot instead of listing again.

### 5. BigQuery data-source access validation (pre-flight)
The validate gate also confirms, **before either leg runs**, that every entity's
BigQuery source table is reachable, so a model that cannot deploy is caught up
front rather than only when the BigQuery leg executes its DDL. Each distinct
source is probed with a **dry-run query**, so BigQuery resolves the reference
exactly as the generated DDL will — this covers a three-part
`project.dataset.table`, a four-part federated **REST-catalog / Lakehouse** name
(e.g. an Apache Iceberg table via BigLake), and quoted identifiers, not just a
three-part name. A missing or inaccessible table fails the push, naming the table
and entity. The probe is billed to the model's deployment-target project (the
same project the deploy runs against). It runs for every `--target` and for
`--validate-only`; a source that is a query (not a table) is skipped.

### User guide
`docs/semantic-model.md` (linked from the README) documents the semantic-model
push end to end, structured for a reader's flow: overview, prerequisites,
authoring (with a worked model document, including the GOOGLE deployment-target
extension), pushing (`--target`/`--validate-only`/`--print`/`--force-remove` in a
flags table), an at-a-glance table mapping each model element to the BigQuery and
Knowledge Catalog resource it produces, the validation gate (including the
data-source pre-flight and federated REST-catalog sources), how a re-push
reconciles removed entities/metrics/relationships/models, and a consolidated
permissions reference.

### Tests
`npx tsc --noEmit` clean; `bun test` 238 pass / 0 fail (new client-method,
validate, data-source-access, link-reconciliation, and force-remove tests
included). The KC write path stays hermetic (CL-gated); #3/#4/#5 are exercised
through the mocked client.


toolbox/mdcode/src/tool/commands.ts
```diff
@@ -27,6 +29,10 @@ export interface InitOptions {
 export interface PushOptions {
   force?: boolean;
```
libei, line 35: What's the difference between force and a forceRemove below? Make sure each field in this push option has a comment clearly describing what it does.

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
```diff
@@ -43,7 +43,6 @@
                 "metadataType": "STRING",
                 "semantics": {
                   "expression": "ss_item_sk",
-                  "importedExpression": "{label/store_sales.attr.store_sales.ss_item_sk}",
```
libei, line 46: If this imported expression is not stored in Knowledge Catalog, does that mean if I do a fresh pull, this field will no longer be restored at the client side? That does mean the KCMD push to Knowledge Catalog is lossy. If that's the case, we need to be very clear in the documentation/user guide.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,216 @@
...
+
+### What gets created in Knowledge Catalog
+
+Each element of your model maps to one catalog resource. Every resource type
+below is a built-in system type under `dataplex-types/global` — push references
+them, it never creates them.
+
+| Model element | Catalog resource | Kind | Id |
+|---|---|---|---|
+| the model | `semantic-model` | entry — anchor / parent of the rest | `<model>` |
```
libei: remove "the", "an", "an" etc from the model element column.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,216 @@
...
+|------|--------|
+| `--target <bq\|kc\|all>` | Which destination(s) to deploy to; accepts a comma-separated list (`bq,kc`). Default `all`. |
+| `--validate-only` | Run every validation check and report pass/fail, but write nothing. |
+| `--print` | Print each destination's generated artifact (BigQuery DDL, Knowledge Catalog entry plan). Combine with `--validate-only` to preview without deploying. |
+| `--force-remove` | Delete models in the entry group that this push no longer includes (see [Updating and removing models](#updating-and-removing-models)). |
+
+Destinations always deploy BigQuery-first and fail fast, so a rejected model
+never half-deploys.
+
+### What gets created in BigQuery
```
libei, line 122: I like the format of the table below in the section of what gets created in Knowledge Catalog. Can we do a similar table for the BigQuery section?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,216 @@
...
+  (e.g. an Iceberg table via BigLake), or a quoted identifier all work. A table
+  that does not exist or that you cannot access fails the push, naming the table
+  and the entity; a `source` that is a query (not a table) is skipped. *(live —
+  needs BigQuery access; see [Permissions](#permissions))*
+
+The live table check runs for **every** `--target`, because the same tables back
+both destinations — so even a Knowledge-Catalog-only push confirms the model
+could deploy to BigQuery.
+
+## Updating and removing models
```
libei, line 189: This updating and removing models section is still hard to read. It reads like a design doc rather than a user guide. Can you make sure it's really written from a user's perspective?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei, line 45: Please make sure each field or property here has a proper comment explaining what it does, including existing fields.

libei, line 80: Same here for each field. Please add a comment to explain what it means, including existing fields not in this PR.

libei, line 142: This function becomes really long. Can you refactor to have a multiple meaningful subroutines? I want to be able to look at this structure of this function and understand what it does clearly.

toolbox/mdcode/src/tool/commands.ts
libei: Done (1fc62b1). Different scopes: \`force\` is the generic push flag for non-semantic-model (CatalogSync) scopes — the semantic-model legs ignore it. \`forceRemove\` is semantic-model KC only and specifically authorizes deleting models the push no longer includes. Every field in \`PushOptions\` now has a comment spelling this out.

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
libei: Good catch — yes, dropping \`importedExpression\` makes a *catalog-only* reconstruction lossy for the vendor SQL (the local document keeps it and it is still used for BigQuery SQL generation, so a normal edit-and-repush is not lossy). Documented explicitly in the gues note (1fc62b1): "the catalog is not a full copy of your model … keep your model document as the source of truth: a model reconstructed only from the catalog would come back without its vendor SQL."

toolbox/mdcode/docs/semantic-model.md
libei: Done (1fc62b1) — articles dropped from the Model element column in both mapping tables (Model / Entity / Metric / Relationship).

toolbox/mdcode/docs/semantic-model.md
libei: Done (1fc62b1) — "What gets created in BigQuery" is now the same style of table: Model → PROPERTY GRAPH, Entity → NODE TABLE, Relationship → EDGE TABLE, Metric → MEASURE.

toolbox/mdcode/docs/semantic-model.md
libei: Rewritten from the user's perspective (1fc62b1): framed as "When you edit / delete an entity or metric / rename or delete a relationship / remove a whole model", each saying what you do and what push does for you, ending with how to read the summary line.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -229,18 +286,16 @@ interface ReconcileOutcome {
...
-// by a dropped relationship is not deleted yet; it needs a lookup-per-entity
-// sweep (follow-up).
-async function reconcileDeletions(
+// Whole-model removal (an anchor no longer in this push) is handled separately by
+// the --force-remove guard in deployKnowledgeCatalog; orphaned relationship links
+// are handled by reconcileLinks. This step operates on `existing` -- the
+// entry-group listing captured once before writes -- so it never issues its own
+// list call; a re-emitted entry is never a deletion candidate, so the pre-write
+// snapshot stays correct.
+function reconcileDeletions(
```
libei, line 339: The comment of this function can be much better. It is hard to read now. We need to describe it from a user perspective.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+### What gets created in BigQuery
+
+`push` executes a single `CREATE OR REPLACE PROPERTY GRAPH` per deployment
+target, in the project and dataset the target names. Each part of your model
+becomes one part of that graph:
+
+| Model element | BigQuery construct | Notes |
+|---|---|---|
+| Model | one `PROPERTY GRAPH` | named by the deployment-target URI |
```
libei: Similarly, remove "a," "an," and "the" from this BigQuery construct column.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+The live table check runs for **every** `--target`, because the same tables back
+both destinations — so even a Knowledge-Catalog-only push confirms the model
+could deploy to BigQuery.
+
+## Updating and removing models
+
+Your model document is the source of truth. To change what is deployed, edit
+the document and run `kcmd push` again — you never edit the catalog or the
+BigQuery graph by hand. Re-running is safe: each push makes the destinations
```
libei: Please make sure when you refer to BigQuery Graph, you use uppercase G in Graph.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done (fd10b2c). \`KcDeployOptions\` now has a comment on every field — project / location / entryGroup, the systemType* overrides, validateOnly, forceRemove, and the two entryCreate* retry knobs.

libei: Done (fd10b2c). Every field of \`KcDeployResult\` is now commented, including the previously-bare \`success\` and \`details\`, plus created/updated/deleted/linked/unlinked/plan.

libei: Refactored (fd10b2c). \`deployKnowledgeCatalog\` is now just the phase sequence, each phase a named helper: \`emitModels\` → \`checkEntryIdCollisions\` → \`buildPlan\` (stops here for --validate-only) → \`listEntryGroup\` → \`guardForeignModels\` → \`writeModels\` → \`reconcileDeletions\`. The body reads top-to-bottom as those steps; behavior is unchanged (238 tests still pass), and running tallies are threaded through a \`Counts\` object so a mid-way failure still reports partial progress.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Rewritten from the user perspective (fd10b2c): "Removes the catalog entries left behind when you delete an entity or metric from a model and push again…", then what it will and won't touch (only entries this push owns; other models in a shared group and the anchors are left alone). Also fixed a stale comment on the function above that referenced a \`defaultProject\` param that does not exist.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+> your model document as the source of truth: a model reconstructed only from
+> the catalog would come back without its vendor SQL.
+
+## Validation
+
+`push` and `--validate-only` run the same checks, **before either destination is
+touched**, so a model that cannot deploy fails fast instead of half-deploying:
+
+* **At least one deployment target per model.** *(static)*
+* **Every metric on a BigQuery-graph model resolves to exactly one entity** —
```
libei: Please do not use this term. Use "BigQuery Graph" with an uppercase G for Graph.

toolbox/mdcode/docs/semantic-model.md
libei: Done (d4f7a53) — dropped the leading articles from the BigQuery construct column: PROPERTY GRAPH / NODE TABLE / EDGE TABLE / MEASURE.

toolbox/mdcode/docs/semantic-model.md
libei: Done (d4f7a53) — changed "BigQuery-graph model" to "BigQuery Graph model" (and "dropped from the graph" to "dropped from the BigQuery Graph").

toolbox/mdcode/docs/semantic-model.md
libei: Done (d4f7a53) — "BigQuery Graph" is now capitalized consistently throughout the guide; the only remaining lowercase "graph" is the SQL construct "property graph" / "PROPERTY GRAPH".

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
+# Deploying a semantic model
+
+A *semantic model* describes your entities (tables), the metrics computed over
+them, and the relationships between them, authored as a single
+[Apache Ossie](https://ossie.dev) document. `kcmd push` deploys one model to two
+destinations at once:
+
+* **BigQuery** — a queryable `CREATE OR REPLACE PROPERTY GRAPH` over the model's
+  tables, so the model can be traversed and its metrics computed in SQL.
+* **Knowledge Catalog (Dataplex)** — catalog entries and links that make the
```
dlychagin-gg: nit. Should we remove 'Dataplex' from here? Let's use 'Knowledge Catalog' throughout this doc.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+  model discoverable as metadata.
+
+Both are generated from the same source document — you never author them
+separately, and a single `push` keeps them in sync.
+
+This guide covers authoring, deploying, and updating a model. For the Ossie
+document format itself, see [ossie.dev](https://ossie.dev).
```
dlychagin-gg: https://ossie.apache.org/

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+You also need read/write access to whichever destinations you deploy to — see
+[Permissions](#permissions).
+
+## 1. Author a model
+
+Create the local layout. The scope is the Dataplex EntryGroup the model will be
+published to, written as `<projectId>.<locationId>.<entryGroupId>`:
+
+```bash
+kcmd init --semantic-model my-project.us.my_models
```
dlychagin-gg: since we have 1 model per entity group, let's change to: my-project.**us-central1.my_model**

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+`init` provisions that EntryGroup (idempotent — an existing group is fine) and
```
dlychagin-gg: Instead of "an existing group is fine" we should say that the group must already exist. (i.e. init will not create the group)

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+creates its local directory. Author the model at
+`catalog/EntryGroups/<entryGroupId>/<model>.yaml`:
+
+```yaml
+version: "0.2.0.dev0"
+
+semantic_model:
+  - name: sales
```
dlychagin-gg: the model name should match the name in `<model>.yaml`.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+    # Required: at least one deployment target, in a GOOGLE custom extension.
+    # `data` is a JSON string whose deploymentTargets is a list of graph URIs.
```
dlychagin-gg: let's say that for now there can be only 1 item in the list.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+    custom_extensions:
+      - vendor_name: GOOGLE
+        data: '{"deploymentTargets": ["//bigquery.googleapis.com/projects/my-project/datasets/sales/propertyGraphs/sales_graph"]}'
+    datasets:                                # each dataset becomes an entity
+      - name: orders
+        source: my-project.sales.orders      # the backing BigQuery table
+        primary_key: [o_orderkey]
+        fields:
+          - name: o_orderkey
+            expression: {dialects: [{dialect: ANSI_SQL, expression: o_orderkey}]}
```
dlychagin-gg: since this is 1P yaml, we should say that `dialect: BIGQUERY` instead of `ANSI_SQL`

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+The target's project and dataset are where the property graph is created; the
+same URIs are also recorded on the model's Knowledge Catalog entry. A model with
+no deployment target is rejected at push time (see [Validation](#validation)).
+
+### Table sources
+
+Each entity's `source` is its backing BigQuery table. A `source` written as
+`dataset.table` (two parts) is qualified with the scope's project — the
+`<projectId>` from `init`, *not* your ambient `gcloud` project. Write the full
```
dlychagin-gg: nit. should we remove: *not* your ambient `gcloud` project ? We already say that it's: `<projectId>` from `init`

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+```yaml
+version: "0.2.0.dev0"
+
+semantic_model:
+  - name: sales
+    # Required: at least one deployment target, in a GOOGLE custom extension.
+    # `data` is a JSON string whose deploymentTargets is a list of graph URIs.
+    custom_extensions:
+      - vendor_name: GOOGLE
+        data: '{"deploymentTargets": ["//bigquery.googleapis.com/projects/my-project/datasets/sales/propertyGraphs/sales_graph"]}'
```
dlychagin-gg, line 52: do you think we should have 2/3 part name for the graph name here? this is to make it consistent with 'source' field which expects 2/3 part name. Internally kcmd push will translate the name to the resource id (//bigquuery.googleapis.com/...). kcmd pull will translate the resource id back to 2/3 part name?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+Each entity's `source` is its backing BigQuery table. A `source` written as
+`dataset.table` (two parts) is qualified with the scope's project — the
+`<projectId>` from `init`, *not* your ambient `gcloud` project. Write the full
+`project.dataset.table` when a table lives in another project.
+
+Sources are not limited to native BigQuery tables. A name with more than
+three parts — for example a four-part `catalog.database.schema.table`
+reference — points at a table in a **federated REST catalog**, such as an
+Apache Iceberg table exposed through BigLake. Write it exactly as BigQuery
+resolves the name; it is emitted into the property graph verbatim, and
```
dlychagin-gg: nit. let's remove: "it is emitted into the property graph verbatim". I think "Write it exactly as BigQuery
resolves the name" is sufficient

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+```bash
+kcmd push --target bq          # BigQuery only
+kcmd push --validate-only      # run all checks, write nothing
+kcmd push --print              # also print the generated DDL / entry plan
+```
+
+| Flag | Effect |
+|------|--------|
+| `--target <bq\|kc\|all>` | Which destination(s) to deploy to; accepts a comma-separated list (`bq,kc`). Default `all`. |
```
dlychagin-gg, line 113: I thought we decided to remove --target and always push to both

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+| `--validate-only` | Run every validation check and report pass/fail, but write nothing. |
+| `--print` | Print each destination's generated artifact (BigQuery DDL, Knowledge Catalog entry plan). Combine with `--validate-only` to preview without deploying. |
```
dlychagin-gg, line 115: I thought we decided to remove printing of KC entries. Should we rename this to --print-ddl?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+| `--force-remove` | Delete models in the entry group that this push no longer includes (see [Updating and removing models](#updating-and-removing-models)). |
+
+Destinations always deploy BigQuery-first and fail fast, so a rejected model
+never half-deploys.
+
+### What gets created in BigQuery
+
+`push` executes a single `CREATE OR REPLACE PROPERTY GRAPH` per deployment
```
dlychagin-gg, line 124: does CREATE OR REPLACE PROP. GRAPH preserve existing grants when it updates the graph?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+below is a built-in system type under `dataplex-types/global` — push references
+them, it never creates them.
+
+| Model element | Catalog resource | Kind | Id |
+|---|---|---|---|
+| Model | `semantic-model` | entry — anchor / parent of the rest | `<model>` |
+| Entity | `semantic-entity` (+ built-in `schema` aspect) | entry | `<model>.entities.<entity>` |
+| Metric | `semantic-metric` | entry | `<model>.metrics.<metric>` |
+| Relationship (1:1, 1:N) | `schema-join` | entry link between the two entity entries | derived from the relationship name |
+| Relationship (M:N) | *(none yet)* | not published — push warns and skips | — |
```
dlychagin-gg: Seems that Ossie doesn't have M:N yet. Is this line valid? If not, then we should remove it

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+An entity entry carries its columns and per-field semantics in the `schema`
+aspect; a `schema-join` link carries the join detail — the paired columns and
```
dlychagin-gg: nit. schema-join link carries the **relationship** detail

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+foreign-key direction — in its aspect. A many-to-many (association /
```
dlychagin-gg: nit. remove many-to-many for now?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+> the catalog would come back without its vendor SQL.
+
+## Validation
+
+`push` and `--validate-only` run the same checks, **before either destination is
+touched**, so a model that cannot deploy fails fast instead of half-deploying:
+
+* **At least one deployment target per model.** *(static)*
+* **Every metric on a BigQuery Graph model resolves to exactly one entity** —
+  otherwise it cannot lower to a MEASURE and would be dropped from the BigQuery Graph. Set
```
dlychagin-gg: nit. can we just say: "otherwise it would be dropped from the BigQuery Graph" ?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+* **Every entity's source table is reachable.** Each `source` is probed with a
+  dry-run query, so BigQuery resolves it exactly as the deploy will — a
+  three-part `project.dataset.table`, a four-part federated REST-catalog name
+  (e.g. an Iceberg table via BigLake), or a quoted identifier all work. A table
+  that does not exist or that you cannot access fails the push, naming the table
+  and the entity; a `source` that is a query (not a table) is skipped. *(live —
+  needs BigQuery access; see [Permissions](#permissions))*
+
+The live table check runs for **every** `--target`, because the same tables back
+both destinations — so even a Knowledge-Catalog-only push confirms the model
```
dlychagin-gg, line 186: there's no KC-only push, right?

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+**When you remove a whole model** — deleting its document does *not* remove it
+from the catalog on the next push. Instead, push stops and names the model the
+catalog still has that you no longer push, so you can't wipe out a model by
+accident or by pushing from the wrong directory. When you do mean to remove it,
+run push again with `--force-remove` and its entries and links are deleted
+first.
+
+Every push prints one line per destination summarizing what it did. For a
+`--target all` push:
```
dlychagin-gg, line 217: remove: --target all

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+```
+Deployed 1 BigQuery Graph(s).
+Wrote 5 new and 2 updated Knowledge Catalog entries; removed 1 orphaned entry; linked 2 relationships; unlinked 1 orphaned link.
```
dlychagin-gg, line 221: nit. in the future, instead of orphaned entry/link we should print orphaned entity/metrics/relationships. i.e. the user-facing concepts in the Ossie model

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+## Permissions
+
+`push` needs access to whichever destinations you deploy to.
+
+**BigQuery** — for `--target bq` or `all`, and for the validation pre-flight:
```
dlychagin-gg: remove --target

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,240 @@
...
+
+* `bigquery.jobs.create` in the deployment-target project — to run the deploy's
+  `CREATE OR REPLACE PROPERTY GRAPH` and the validation dry-run query
+* read access on each entity's source table, so the dry-run can resolve it
+* `bigquery.datasets.get` on the target dataset (region detection; optional —
+  push degrades gracefully without it)
+
+**Knowledge Catalog / Dataplex** — for `--target kc` or `all`:
+
+* `dataplex.entryGroups.useSemanticModelAspect` on the destination entry group
```
dlychagin-gg: nit. can we remove required permissions for now? let's add them later, once we finalize the list.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — dropped "(Dataplex)" here and use "Knowledge Catalog" throughout the doc.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — link and anchor text now point to https://ossie.apache.org/.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — now `my-project.us-central1.my_model`.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — added an inline note that `name` must match the `<model>.yaml` filename.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — the example comment now says deploymentTargets holds the target graph URI (for now, exactly one).

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — example now uses `dialect: BIGQUERY`.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — dropped the "*not* your ambient gcloud project" aside.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — kept "Write it exactly as BigQuery resolves the name" and removed the verbatim clause.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — removed the M:N row; Ossie has no M:N yet.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — now "carries the relationship detail".

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — removed the many-to-many sentence.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — now "otherwise it would be dropped from the BigQuery Graph".

toolbox/mdcode/docs/semantic-model.md
libei: Done in 3ea22cd — removed the Permissions section (and its inbound links) for now; will add back once the list is final.

toolbox/mdcode/docs/semantic-model.md
libei: Leaving the --target removal (always push to both) to Bei — the whole flags section and this line will change together once that lands, so holding off on the doc here.

toolbox/mdcode/docs/semantic-model.md
libei: Same bucket as --target — the --print rename / dropping KC-entry printing is part of the CLI-surface decision Bei is making. Holding off until that's settled.

toolbox/mdcode/docs/semantic-model.md
libei: Leaving the init/entry-group semantics to Bei (open question: should init create the group or require it to exist). Left the current wording as-is for now.

toolbox/mdcode/docs/semantic-model.md
libei: Good question — using a 2/3-part graph name for consistency with `source` (with push/pull translating to/from the resource id) is a real change to the deploymentTargets contract, not just docs. Flagging it as a design decision rather than changing the guide to describe unimplemented behavior.

toolbox/mdcode/docs/semantic-model.md
libei: Agreed this is nicer — printing orphaned entity/metric/relationship (the Ossie-model concepts) instead of entry/link. Noting as a follow-up; left the current summary wording for now.

toolbox/mdcode/src/tool/commands.ts
libei: Done in 1fc62b1 — PushOptions now comments every field and spells out force (catch-all override) vs forceRemove (authorizes deleting models this push no longer includes) in commands.ts.

toolbox/mdcode/tests/libts/semantic/fixtures/tpcds_date_edge.knowledge_catalog.golden.json
libei: Done in 0c2a8f2 — the guide's "the catalog is not a full copy of your model" note now states that importedExpression (vendor SQL) is not stored in Knowledge Catalog and that the model document stays the source of truth, so a catalog-only pull loses it.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 1fc62b1 — dropped the articles from the model-element column.

toolbox/mdcode/docs/semantic-model.md
libei: Done in 1fc62b1 — the "What gets created in BigQuery" section now has a matching table (Model/Entity/Relationship/Metric -> BigQuery construct).

toolbox/mdcode/docs/semantic-model.md
libei: Done in 1fc62b1 — rewrote "Updating and removing models" from the user's perspective (edit the document and re-push), not as a design doc.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in fd10b2c — KcDeployOptions now has a comment on every field, including the pre-existing ones.

libei: Done in fd10b2c — KcDeployResult fields are all commented too.

libei: Done in fd10b2c — deployKnowledgeCatalog is now a short phase sequence (emitModels / checkEntryIdCollisions / buildPlan / listEntryGroup / guardForeignModels / writeModels / reconcileDeletions), each its own helper.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in fd10b2c — reconcileDeletions now opens with a user-perspective comment (removes catalog entries left behind when you delete an entity/metric and push again; only entries this push owns).

toolbox/mdcode/docs/semantic-model.md
libei: Done in d4f7a53 — dropped the articles from the BigQuery construct column.

toolbox/mdcode/docs/semantic-model.md
libei: Done in d4f7a53 — "BigQuery Graph" is capitalized throughout.

toolbox/mdcode/docs/semantic-model.md
libei: Done in d4f7a53 — replaced with "BigQuery Graph" (capital G).

toolbox/mdcode/docs/semantic-model.md
dlychagin-gg: @libei are you planning to address this?

toolbox/mdcode/docs/semantic-model.md
dlychagin-gg: @libei are you planning to address this?

toolbox/mdcode/docs/semantic-model.md
dlychagin-gg: @libei are you planning to address this?

toolbox/mdcode/docs/semantic-model.md
dlychagin-gg: it's ok for now, but let's revisit this decision later.

toolbox/mdcode/docs/semantic-model.md
```diff
@@ -0,0 +1,217 @@
...
+          - name: o_totalprice
+            expression: {dialects: [{dialect: BIGQUERY, expression: o_totalprice}]}
+    metrics:
+      - name: total_revenue
+        expression: {dialects: [{dialect: BIGQUERY, expression: SUM(orders.o_totalprice)}]}
+```
+
+### Deployment targets (required)
+
+Every model must declare at least one **deployment target** — the BigQuery
```
dlychagin-gg: nit. let's say must declare one deployment target. The current wording "at least one" implies that we could deploy more than one graph, which I don't think is the case (at least for now)

toolbox/mdcode/src/libts/semantic/deploy_bigquery.ts
dlychagin-gg: why "Dataplex/other URI" is "left alone" here? I think we should fail if the target does not match BQ_GRAPH_TARGET (or alternatively add to the malformed list here and fail once we've identified all malformed targets)

toolbox/mdcode/src/libts/semantic/deploy_bigquery.ts
```diff
@@ -120,10 +127,29 @@ export function bigQueryGraphTargets(model: SemanticModel):
...
   return {targets, malformed};
 }
 
 
+// Every deploymentTarget URI a model declares, regardless of whether each parses
+// as a BigQuery Graph target. Validation uses this to require that a model
+// declares at least one deployment target; bigQueryGraphTargets narrows to the
+// BigQuery graphs. Throws on malformed extension JSON.
+export function deploymentTargetUris(model: SemanticModel): string[] {
+  return googleDeploymentTargets(model).uris;
```
dlychagin-gg, line 140: I think it's ok to fail here if number of deployment targets is not 1.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -88,34 +123,99 @@ function sleep(ms: number): Promise<void> {
...
+  const emit = emitModels(models, opts);
+  warnings.push(...emit.warnings);
+  if (emit.error) return result(false, emit.error);
+  const emitted = emit.emitted;
+
+  // A parsed document always yields at least one model (the loader enforces
+  // `semantic_model` min 1), so an empty `emitted` means no documents were
+  // found. validateOnly mutates nothing, so that is a clean no-op; a real push
+  // treats an empty workspace as a configuration error worth flagging.
+  if (!emitted.length) {
```
dlychagin-gg: let's fail if emitted.length != 1. We currently only allow 1 model per entryGroup. This is will simplify the code below (won't need to check for entryIdCollisions)

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -126,93 +226,113 @@ export async function deployKnowledgeCatalog(
...
-  // most common collision) are still distinguished.
+// Fails (returns the error message) if two models in this push generate the same
+// catalog entry id. Entry ids must be unique within the entry group; the emitter
+// dedups within one model, but two models (two documents, or two `semantic_model`s
+// in one document) whose names normalize to the same id would collide -- on
+// publish the later one 409s and silently upserts over the earlier. Catch it here
+// and name the entry so the author can rename a model. Ownership is tracked by
+// index, not model name, so two same-named models (the common collision) are
+// still distinguished.
+function checkEntryIdCollisions(
```
dlychagin-gg: we should remove this method because we only allow 1 model per entry group for now

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -281,6 +393,167 @@ async function reconcileDeletions(
 }
 
 
+// The schema-join entry-link type name, matching what the emitter stamps on each
+// link (Namer.typeName('entryLink', 'schema-join')). Used to filter
+// lookupEntryLinks to the links this leg owns.
+function schemaJoinLinkType(opts: KcDeployOptions): string {
+  const proj = opts.systemTypeProject ?? 'dataplex-types';
+  const loc = opts.systemTypeLocation ?? 'global';
```
dlychagin-gg, line 384: I thought schema-join has a fixed location (https://docs.cloud.google.com/dataplex/docs/enrich-entries-metadata#system-aspect-types) . Why do we support overrides here?

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
```diff
@@ -281,6 +393,167 @@ async function reconcileDeletions(
...
+    opts: KcDeployOptions): Promise<{entries: Entry[]; error?: string}> {
+  const entries: Entry[] = [];
+  try {
+    for await (const entry of cat.listEntries(
+        opts.project, opts.location, opts.entryGroup)) {
+      entries.push(entry);
+    }
+  } catch (err: any) {
+    const msg = err.message || String(err);
+    if (/may not exist/i.test(msg) ||
```
dlychagin-gg, line 407: it there an error code we can rely on here, instead of error message text?

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -0,0 +1,150 @@
...
+
+
+// Live pre-flight over the same models: confirms every entity's BigQuery source
+// table is reachable BEFORE any destination leg runs, so a push -- to BigQuery
+// or to Knowledge Catalog -- fails fast when the model could not deploy, rather
+// than surfacing a missing table only once the BigQuery leg executes its DDL.
+// The entity sources are BigQuery tables regardless of --target, so this runs
+// for every destination and for --validate-only.
+//
+// Each distinct source is probed with a dry-run query (`SELECT 1 FROM <ref>`),
```
dlychagin-gg: SELECT 1 is not the best to probe for table existence. Can we instead check whether there's an entry in KC for this BQ table? or alternatively use https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/tables/get to get schema of the table to check whether it exists. I assume we can make both work with 4 part names. If we must use SELECT to probe whether the table exists, then let's add WHERE FALSE.

dlychagin-gg (approved):

toolbox/mdcode/docs/semantic-model.md
libei: Done in c32a076 — reworded to "must declare exactly one deployment target" (and the validation section + the "no target" note match). The gate now enforces exactly one.

toolbox/mdcode/src/libts/semantic/deploy_bigquery.ts
libei: Done in c32a076 — any `deploymentTarget` that doesn't parse as a BigQuery Graph URI is now collected as `malformed` and rejected (the shared validate gate names it, so a `--target kc` push is covered too), instead of being left alone. That let me drop the "looks-like-a-graph" hint heuristic entirely — there's now just parses / doesn't-parse.

toolbox/mdcode/src/libts/semantic/deploy_bigquery.ts
libei: Done in c32a076 — the validate gate now fails when a model declares zero or several deployment targets (exactly one required).

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in 119dcfb — push now rejects any count other than one model per entry group up front (empty stays a clean no-op under `--validate-only`). As you noted, that makes an entry-id race between two models impossible, so `checkEntryIdCollisions` is removed.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: Done in 119dcfb — removed. The one-model-per-entry-group guard above makes it unnecessary.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: You're right that the system types (including `schema-join`) live at a fixed location — `dataplex-types/global`, which is the default here. The override isn't used on a real push; it exists only so hermetic tests can point the type references at a fixture types project. I've reworded the misleading comment to say exactly that (119dcfb). Happy to drop the override entirely and have the tests inject the fixture some other way if you'd prefer it gone from the option surface.

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
libei: There isn't a status code to key on here without a client change: `CatalogClient.listEntries` throws a plain `Error` with the HTTP status stringified into the message, so the status doesn't survive to this catch. The "not yet visible" propagation case also isn't a clean 404. I'd rather not guess a code from an error we've flattened to text — I'll file a follow-up to have `listEntries` preserve the status on the throw, so this and `isPropagating` can both key on a code instead of matching phrasing.

toolbox/mdcode/src/libts/semantic/validate.ts
libei: Added `WHERE FALSE` in c32a076 so the dry-run scans nothing. I kept the query form rather than `tables.get`: `tables.get` can't address a four-part federated REST-catalog name (e.g. an Iceberg table via BigLake), which the deploy must resolve, whereas a dry-run resolves the reference exactly as the generated DDL will. Checking for a KC entry wouldn't confirm the table is actually reachable, so the dry-run stays the source of truth.

toolbox/mdcode/docs/semantic-model.md
libei: Currently `init --semantic-model` does create the entry group (idempotently) — `commands.ts` calls `createEntryGroup` at init precisely so push only ever writes entries. So the guide matches today's behavior. If we'd rather `init` require a pre-existing group and never create one, that's a behavior change to `init` itself; I'd do it as its own change rather than just editing the doc. Want me to make that switch?

toolbox/mdcode/docs/semantic-model.md
libei: Will follow up separately...

toolbox/mdcode/docs/semantic-model.md
libei: I will follow up separately...

toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
dlychagin-gg: I see. it's fine to leave schema-join location as configurable if we need that for testing. let's just document this properly.

PR #290, opened 2026-08-14
mdcode: store SQL expressions in the sql-expressions companion aspect

libei:
## What

Stores field/metric SQL in Knowledge Catalog through a single **`sql-expressions` companion aspect** (`data_classification: METADATA_AND_DATA`) on entity and metric entries, per the V2 *"SQL Expression Storage in USL Semantic Models"* proposal (D. Lychagin, 2026-08-13). Adds this to **both push and pull together**, replacing PR #278's approach of writing expressions onto the core `schema` / `semantic-metric` aspects.

The core aspects stay **metadata-only** — they never carry executable SQL. The `schema` field carries only a `semantic: "DIMENSION"` marker.

## How

- **`sql_expressions.ts`** (new) — a shared codec that fixes the aspect's shape and its qualifier convention: omitted qualifier = primary GoogleSQL (IR `expression`); `imported` = the vendor form (IR `importedExpression`). The emitter and reader are exact inverses over it.
- **Emitter** (`knowledge_catalog.ts`) — attaches the `sql-expressions` aspect alongside `schema` / `semantic-metric` only when a field/metric actually has an expression (never an empty aspect); emits the `DIMENSION` marker.
- **Reader** (`kc_converter.ts`) — recovers **both** the primary and imported forms from the aspect (matched by type suffix); reads the `DIMENSION` marker. A `--emit-expressions` push now round-trips both expression forms through pull.
- **Pull orchestration** (`pull_kc.ts`) — requests the `sql-expressions` aspect for entity and metric entries (harmless when absent).

The source **dialect label** is still not stored (inferred from `importedSystem`), so `importedDialect` stays unset — the one remaining expression-side loss.

## Gating / compatibility

Both the companion aspect and the `DIMENSION` marker remain gated behind `--emit-expressions` (**off by default**), because the published system-type templates don't carry them yet. **A default push emits neither and matches the live published system types byte-for-byte** — the committed emitter/pull golden fixtures are unchanged.

## Testing

`bun test ./tests/libts/semantic/` — 328 pass. New `sql_expressions.test.ts` covers the codec (primary/imported/both/neither, inverse, first-wins, malformed/unknown-qualifier skipping); emitter, reader, and pull tests assert the companion aspect and a `--emit-expressions` field+metric round-trip.

> The 3 pre-existing `osi_schema.test.ts` failures are on the base branch (#277), not introduced here: the pull golden YAMLs are generated from a default (expression-free) push, so they lack the `expression` the OSI schema requires. Regenerating those fixtures from an `--emit-expressions` push is a #277 fixture concern.

## Stacking

**Stacked on #277** (the KC pull PR) — this branch builds on `upstream-pr5-kc-pull`, so until #277 merges the diff here also shows #277's commits; it will narrow to just this change once #277 lands. Merge #277 first.

Assumes the `sql-expressions` aspect type will be provisioned in the published system-type templates before `--emit-expressions` is used against live types.


libei:
Closing unmerged. Open as a draft since 2026-08-14 and stacked on #277; not being pursued. Reopen or re-propose if the companion-aspect route comes back.

PR #298, opened 2026-08-15
mdcode: address KC pull review comments (post-merge follow-up)

libei:
Follow-up to the merged #277 (semantic-model Knowledge Catalog pull leg), addressing the inline review comments left after merge. Reader / pull-orchestration / CLI only — the writer (#278) is untouched.

## Changes
- **pull_kc:** hard-fail on any entry-hydration fetch error and on a non-200 link lookup; an empty link list stays a silent no-op (`:81`, `:112`).
- **One model per entry group:** dropped the `pull --model` flag; >1 anchor in a group is now a hard error naming both (`:483`, `:261`, `:140`).
- **`pull --force-remove`:** authorize replacing a differently-named local model with the catalog's, rather than silently leaving two models in the group (`:269`).
- **Schema-join link type** is a fixed built-in constant, never `undefined` (`:108`, `:182`).
- **push:** warn when a relationship name will be normalized — KC stores the name only in the link id, so a pull returns it lowercased/hyphenated (`:295`).
- **Typeless defaults:** untyped fields and metrics now default to `Opaque` (STRING + OTHER) instead of guessing STRING/NUMERIC; a typeless metric round-trips un-typed (`:296`).
- **osi_schema guardrail:** the `.pull.golden.yaml` fixtures come from a default (expression-free) KC push, so their fields/metrics lack the OSI-required `expression`. They're excluded from the OSI schema check with a `TODO(#290)` — #290 (the sql-expressions companion aspect) restores expressions on push+pull, at which point they can be schema-checked again.
- Docs (`docs/semantic-model.md`) + tests + regenerated goldens to match.

## Deferred (tracked as follow-ups, not in this PR)
- Recover entity keys (pk/unique) on pull → schema aspect.
- Recover `ai_context` on pull → guidelines aspect.
- Recover field labels on pull → schema aspect `annotations`.

## Testing
`bun test` — green (the one local `ajv/dist/2020` load error is a borrowed-`node_modules` artifact; `ajv ^8.17.1` is a declared devDependency so CI resolves it).

### DIFF
--- a/toolbox/mdcode/docs/semantic-model.md
+++ b/toolbox/mdcode/docs/semantic-model.md
@@ -201,11 +201,10 @@ paired columns and foreign-key direction — in its aspect.
 > block or a `semantic-metric.expression` field, so the default push omits them
 > (pass `--emit-expressions` to write the canonical GoogleSQL/ANSI expression
 > once the templates gain the fields). It never stores entity keys, `ai_context`,
-> field labels, the original vendor SQL (`importedExpression` — e.g. the MAQL or
-> Snowflake form a metric was imported from), or M:N relationships. Those stay in
-> your authored document (and, for the edges, in the BigQuery property graph); the
-> vendor SQL and expressions are still used when generating BigQuery SQL. Keep
-> your model document as the source of truth.
+> field labels, or the original vendor SQL (`importedExpression` — e.g. the MAQL
+> or Snowflake form a metric was imported from). Those stay in your authored
+> document; the vendor SQL and expressions are still used when generating
+> BigQuery SQL. Keep your model document as the source of truth.
 
 ## Validation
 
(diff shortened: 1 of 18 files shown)


libei:
Follow-up commit `b2ee552` from a self-review pass:

- **Pull reconcile compares by on-disk path, not raw name.** A model name that sanitizes to the same file (e.g. `a/b` → `a_b.yaml`) was flagged as a stale conflict on every re-pull because `catalogNames` held the raw name while `modelDocuments()` returns filename-derived names. Now both sides are compared through `modelPath`.
- **`mapConcurrent` stops fanning out after the first failure.** `Promise.all` already rejects on the first throw, so the remaining workers' fetches were wasted work whose rejections surfaced as unhandled-rejection noise; a shared `failed` flag now halts new claims.
- **OSI guardrail validates the `.pull.golden.yaml` fixtures too**, tolerating *only* the known missing-`expression` gap (TODO #290) rather than skipping them by filename — so unrelated schema drift is caught today, and the fixtures schema-check with no special-casing once #290 restores expressions (verified: all three fail *only* with `required:expression`).

Deferred, intentionally not in this commit:
- **FK-direction orientation on live pulls** relies on `dataSource` string equality, which is not normalization-robust the way endpoint *identity* (matched by entry id) is. If Dataplex normalizes an entity aspect's source resource but not the link references, direction could fall back to reference order (warned, not errored). Needs live-KC confirmation and is closer to writer-side (#278/#290) scope.
- **`--dry-run` returning a non-zero exit on a stale-local conflict** — arguably correct (it signals the real run would fail); left as-is.

`tsc` clean; full suite 330 pass / 0 fail.

Issue #302, opened 2026-08-17
Proposal: portable bundle descriptors for pre-load discovery in multi-bundle agent runtimes


LearnerHong:
## Summary

OKF supports progressive disclosure *within* a bundle, but not *across a set of bundles*.

When an agent runtime mounts multiple OKF bundles, the first-hop context often contains only opaque bundle identifiers or paths. The agent cannot reliably decide which bundle to search without eagerly opening every root `index.md` or scanning concept metadata.

This issue proposes a small, portable **bundle descriptor** based on optional bundle-level `title` and `description`, plus consumer guidance to expose that descriptor before loading the root index body or any concepts.

## Current gap

The current v0.2 specification provides:

- [§8](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#8-index-files): `index.md` supports progressive disclosure of the contents *inside* a directory.
- [§4.1](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#41-frontmatter): `title` and `description` are recommended for concept documents.
- [§12](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md#12-versioning): the bundle-root `index.md` may carry `okf_version`, while §8 otherwise says index files contain no frontmatter.

As a result, a catalog can enumerate bundle references, but there is no standard machine-readable answer to:

> What is this bundle about, and when should an agent consult it?

Concept-level descriptions arrive too late: the consumer must first choose and open a bundle to discover them.

## Proposed minimal contract

Allow a bundle to publish a compact descriptor, for example in root `index.md` frontmatter:

```yaml
---
okf_version: "0.2"
title: Cloud cost optimization
description: Guidance for analyzing spend, budgets, and optimization opportunities.
---
```

A catalog or runtime could expose a lightweight discovery record such as:

```json
{
  "bundle_ref": "<consumer-defined reference>",
  "title": "Cloud cost optimization",
  "description": "Guidance for analyzing spend, budgets, and optimization opportunities."
}
```

The identifier and revision mechanism remain consumer-defined. The interoperable OKF part is the meaning of `title` and `description`.

## Progressive consumption flow

1. **Discover bundles:** expose only compact bundle descriptors.
2. **Select scope:** choose the likely relevant bundle or bundles.
3. **Navigate:** read the selected root `index.md` for directory-level disclosure.
[...]

Issue #214, opened 2026-07-21
Bundle-level metadata has no slot — collection-scoped fields (licence, upstream provenance) in a multi-corpus deployment


johnno100:
Hi — thanks for publishing OKF. We're adopting it in an enterprise knowledge platform and hit one structural gap early. Flagging it now rather than after the fact, since we intend to propose something and would rather build in the direction you're already heading.

## Context

We're running roughly **36 OKF-shaped bundles, ~4,370 concept documents**, mostly mirrors of external corpora (public engineering handbooks, open guides, reference works) adopted into a governed internal platform. OKF's model fit almost immediately — our mirrors were already markdown-with-frontmatter, and bundle-relative links already formed the graph. That part needed no convincing.

## The gap

**There is no home for collection-scoped metadata.**

Some facts are properties of a *bundle*, not of any concept within it:

- where the corpus was mirrored from, and **at which upstream commit** (without this a bundle is a snapshot, not a reproducible asset)
- the licence and attribution terms governing the whole collection
- when it was adopted, and by whom

Today the spec offers nowhere to put these:

- `index.md` is reserved and carries **no frontmatter**
- `log.md` is prose change history
- per-concept frontmatter works, but replicating a collection-level licence across ~1,100 files in a single bundle is unmaintainable, and creates 1,100 places for it to drift

The practical consequence for us is that governance facts end up in an out-of-band file that no OKF consumer knows to read — which undercuts the portability that makes OKF attractive in the first place.

## What we're prototyping

A `<bundle-root>/*.yaml` sidecar holding collection-level defaults, with simple inheritance: **concept frontmatter wins; absent ⇒ inherit from bundle; absent at both ⇒ a declared fail-safe default.**

We chose a sidecar because the spec reserves only `index.md`/`log.md` and treats all other `.md` as concepts, so a non-`.md` file at bundle root appeared to be the least invasive option. We may well be wrong about that.

To be clear about scope: **we are not asking OKF to absorb our governance vocabulary.** Trust, confidence and attestation fields are specific to our platform and belong in our own extension namespace under the "preserve and tolerate unknown fields" rule — which, incidentally, is the single most useful sentence in the spec for adopters. **We're only raising the question of where bundle-scoped fields should live at all.**

## Questions

1. Is bundle-level metadata in scope for v0.2, or deliberately excluded?
[...]

Issue #212, opened 2026-07-20
Proposal: an opt-in `okf_profile` declaration (parallel to `okf_version`) — plus a spec clarification on root-index frontmatter keys


jpavley:
## Summary

§9's conformance surface is deliberately minimal: parseable frontmatter, non-empty `type`, reserved-file structure. Everything else about authoring quality is left to producers — so every serious producer builds a private validation layer. This tracker is itself the evidence: #196 (`okft`) is an independently built community linter, and we built our own before finding it.

Proposal: give that layer a *name* the format understands — an optional, domain-scoped profile declaration — and clarify one spec seam it depends on.

## The proposal (opt-in, advisory, no consumer change)

A bundle MAY declare, in root-index frontmatter beside `okf_version`:

```yaml
okf_version: 0.1
okf_profile: civic-v1
```

A profile is a producer-published document that pins, for one domain: a `type` vocabulary, any required-beyond-`type` fields, and lint rules. Consumers that don't know the profile ignore the key — exactly the `okf_version` mechanism §11 already established. Nothing becomes rejectable; profile checks are authoring-time warnings, never consumer-side errors. This is deliberately NOT a central or mandatory type registry (an explicit spec Non-goal) — profiles are domain-scoped, opt-in, and live with their producers.

Why it's worth a spec-level convention rather than staying private: interop between the authoring tools this ecosystem is already growing. A linter like #196's `okft` could read a declared profile and apply the domain's rules instead of only generic hygiene checks; two producers in the same domain could share one profile instead of two private validators.

## The spec clarification this depends on

We hit a genuine ambiguity implementing this: **may a producer key appear in root-index frontmatter at all?** §4.1's extension-key permission governs *concept documents*; §6 says index files contain no frontmatter; §11 carves an exception for `okf_version` in the root index — and is silent on additional keys. `okf_profile` (or any bundle-level producer key) has no clearly legal home. A one-line clarification in §6 or §11 — "the root index MAY carry additional producer keys beside `okf_version`" (or "MUST NOT," and bundle-level metadata belongs somewhere else) — would settle it either way.

## Evidence

[...]

Issue #96, opened 2026-06-18
Proposal: Metadata orientation hints (purpose, task, audience) for agent routing


magnus919:
## Summary

Add optional metadata frontmatter fields (`purpose`, `task`, `audience`) that help agents orient themselves before reading a bundle or concept file. These fields provide lightweight context about why a knowledge bundle exists, what agent task produced it, and who should consume it — without requiring new type values or a taxonomy registry.

## Motivation

OKF models domain knowledge: what the world knows about a subject. But the same format is increasingly used to document what an agent accomplished as proof of work that downstream agents need to validate. A bundle of domain concepts cannot easily serve as a task deliverable because the concepts describe what is true, not what an agent discovered or decided.

OKF's unregistered `type` field is one of its best features. Adding required type values (like `type: Analysis` or `type: Discovery`) would create the governance problem OKF wisely avoids. Instead, producers should add metadata hints that help an agent decide whether to read before parsing.

## Proposal

Add the following optional frontmatter fields to the recommended list:

```yaml
purpose: <short description of why this bundle or concept exists>
task: <name or description of the agent task that produced this content>
audience: <intended consumer role, e.g., "product-manager", "data-engineer">
```

All three are OPTIONAL. None require a spec change. They are hints, not enforcement. Consumers that do not understand them ignore them. Consumers that do understand them can route more intelligently.

The `description` field already exists in OKF v0.1 frontmatter and serves a similar orientation function at the concept level. The Agent Skills format (agentskills.io) uses a `description` field in exactly this way: a short string that tells an agent whether to load more. This proposal extends that pattern to the bundle level and adds task/audience context.

## Relationship to Spec

Backward-compatible addition to the recommended frontmatter fields. OKF already permits arbitrary additional keys. This formalizes a naming convention for commonly useful orientation metadata.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

magnus919:
## Prior Art

The artifact-pyramids agent skill (github.com/groktopus/artifact-pyramids) uses the `description` frontmatter field at both the bundle level (root index.md) and the concept level as the primary orientation hint for consuming agents. The Agent Skills format (agentskills.io) uses the same pattern: a `description` field that tells an agent whether to load more, at approximately 100 tokens of context cost. This proposal extends that established pattern to OKF with additional optional fields (purpose, task, audience) for richer routing context.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

[...]

PR #303, opened 2026-08-17
okf: allow extensible bundle-root frontmatter for discovery

LearnerHong:
## Summary

This PR updates `okf/SPEC.md` to make the YAML frontmatter of the bundle-root `index.md` an extensible metadata container.

It:

- permits producer-defined keys alongside `okf_version`;
- standardizes optional bundle-level `title` and `description` descriptors;
- adds guidance for discovering and selecting among multiple bundles before loading their content;
- clarifies that the Section 11 `type` rule applies to concepts rather than the bundle-root index; and
- aligns the related language in Sections 12 and 13.

## Motivation

OKF already supports progressive disclosure within a bundle through its root `index.md`. However, a consumer presented with multiple bundles currently has no portable, standardized way to understand what each bundle contains before reading that index body or any concept document.

Extensible root frontmatter provides a general metadata surface, while `title` and `description` provide a minimal interoperable convention for pre-load discovery. Producers may add other metadata without requiring a new specification field for every use case, and consumers remain compatible by tolerating unknown keys.

A typical consumption flow is:

1. inspect the available bundle descriptors;
2. select the bundles relevant to the current task;
3. read the selected root `index.md`;
4. search or traverse concepts; and
5. load full concept content only when needed.

## Scope

This proposal is limited to the OKF document format. It does not define:

- `purpose`, `task`, or `audience` routing hints;
- a discovery or retrieval API;
- a separate manifest;
- storage or transport behavior; or
- an agent runtime protocol.

The proposal is orthogonal to the concept-level routing discussion in #96 and #189.

## Compatibility and versioning

The new fields are optional. Bundles that omit `title` and `description` remain conforming, and consumers must tolerate producer-defined root-frontmatter keys they do not recognize.

This is a backward-compatible addition under Section 12. This draft leaves the document at v0.2 so maintainers can decide whether it should land in v0.2 or a subsequent minor revision.

Addresses #302.

Related: #212, #214.

### DIFF
--- a/okf/SPEC.md
+++ b/okf/SPEC.md
@@ -507,8 +507,26 @@ disclosure**: letting a human or agent see what is available before
 opening individual documents.
 
 Index files contain no frontmatter, with one exception: a bundle-root
-`index.md` MAY carry an `okf_version` key (§12). The body uses one or more
-sections, each grouping concepts under a heading:
+`index.md` MAY carry a YAML frontmatter block. The following optional fields
+are defined for that block:
+
+- `okf_version`: The OKF version targeted by the bundle (§12).
+- `title`: A human-readable display name for the bundle.
+- `description`: A single sentence summarizing the bundle's scope and when a
+  consumer should consult it.
+
+The bundle-root frontmatter MAY also contain additional producer-defined
+key/value pairs. Consumers MUST tolerate unknown keys and SHOULD preserve them
+when round-tripping.
+
+`title` and `description` form a compact **bundle descriptor**. They describe
+a bundle as a whole rather than an individual concept. This enables
+progressive disclosure across multiple bundles: a consumer MAY inspect
+available descriptors to select relevant bundles before reading an index body
+or any concept document. This guidance does not prescribe how bundles are
+enumerated, identified, stored, or transported.
+
+The body uses one or more sections, each grouping concepts under a heading:
 
 ```markdown
 # Section / Group Heading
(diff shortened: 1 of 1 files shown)


LearnerHong:
@libei 

Hi @libei, I noticed your recent work on mdcode and OKF bundle round-tripping.

This proposal adds an extensible metadata surface to the bundle-root index for pre-load discovery. Since it may affect how tools preserve and consume root metadata, would you be willing to advise whether you are the right reviewer, or point me to the appropriate maintainer?

No urgency, and I am happy to revise the proposal based on maintainer guidance.

andrewcrenshaw:
Producer support for this change, with a measured datapoint behind it: this exact ambiguity has been resolved conservatively by four independent producers, each concluding we could not tell from sections 6 and 11 whether anything beside `okf_version` was legal in root-index frontmatter. Our exporter (the *remember* wiki compiler at Lexenne, PR #195 / #182) emits `okf_version` and nothing else for that reason; @jpavley reported the same resolution for civic-records bundles on #212, @johnno100 for a 36-bundle enterprise deployment on #214, and @kaeferboeck for safety-critical automation KBs on #212. Four producers guessing identically at the same gap is the practical case for the one-line MAY.

The `title` / `description` descriptors also give the `okf_profile` declaration discussed on #212 a defined place to live, which is the half that discussion was waiting on.

One compatibility note worth stating in review: producers who resolved conservatively have nothing to migrate - the change is purely permissive - so it costs existing bundles nothing, ours included.

PR #305, opened 2026-08-18
kcmd: take reconciliation's owned entry-id prefixes from the emitter

feixufeixu:
Fixes a case where delete reconciliation silently did nothing. idOf returned the full resource name when a listed entry spelled its project as a number rather than an id, so no id matched an ownership prefix: nothing was deleted and the push still reported success. It now recovers the id from the container's shape, which also keeps ids that themselves contain slashes whole.

Delete reconciliation, link reconciliation and foreign-model removal each built `<anchor>.entities.` / `<anchor>.metrics.` inline, matching only one entry-id scheme. Emitters now declare the prefixes they own on KcResources and the publisher uses those. Foreign-model removal runs for a model that is not in the push, and so has no emitter to ask, and matches the anchor followed by a separator instead.

Also splits deployKnowledgeCatalog at the emitModels boundary so the publish phases can be driven by an emitter that does not build the semantic IR, and guards the two unchecked `resources.entries[0]` accesses that run after writes.

 Context, not part of the change: the second emitter is a LookML one in a follow-up PR. It publishes ids in a path form (`<model>/metrics/<view>/<name>`) rather than Ossie's dotted form, which is what surfaced both the `idOf` bug and the inline prefixes. This PR stands alone — the `idOf` fix applies to the Ossie path today with no LookML involved.

New tests: reconciliation recognises entries when the server spells the project as a number; an empty `ownedPrefixes` entry cannot delete a whole entry group; a foreign model whose ids use a different scheme is removed whole.

### DIFF
--- a/toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
+++ b/toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
@@ -142,7 +142,26 @@ function sleep(ms: number): Promise<void> {
 export async function deployKnowledgeCatalog(
     models: LoadedModel[], ctx: context.ApiContext,
     opts: KcDeployOptions): Promise<KcDeployResult> {
-  const warnings: string[] = [];
+  // Emit every model to catalog resources up front (pure -- no network).
+  const emit = emitModels(models, opts);
+  if (emit.error) {
+    return {
+      success: false, details: emit.error, warnings: emit.warnings,
+      created: 0, updated: 0, deleted: 0, linked: 0, unlinked: 0, plan: [],
+    };
+  }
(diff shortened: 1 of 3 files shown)


libei (approved):
LGTM. The `idOf` container-shape fix is the right root-cause fix, and having the emitter own its id prefixes (rather than the publisher guessing the scheme) is a clean seam. Reasoning in the comments is thorough, tests cover the three real cases, and I ran the suite locally (30/30) with a clean typecheck of the touched files.

nit (non-blocking): the new test comment on the empty-`ownedPrefixes` case says "deployCatalogResources is exported as the seam" — the exported function is actually `deployEmittedModels`.

Issue #319, opened 2026-08-20
Optional usage-based navigation hints for agents


sridhar-mani:
OKF provides the structure of the knowledge catalog, but an agent starting from an index has no information about which concepts have historically been useful for a particular intent or set of conditions.

PR #320, opened 2026-08-20
Add and rank usage-based navigation hints by prompt similarity

sridhar-mani:
## Summary

Adds an optional `usage.yaml` file to OKF for advisory, usage-based navigation hints.

The file records which concepts have historically been useful for specific intents and conditions, along with usage counts. The reference implementation can use this information to rank likely-relevant concepts and improve navigation over time.

The change includes:

* `usage.yaml` schema and specification updates
* Usage recording and ranking in the reference implementation
* CLI commands for recording and ranking usage
* Agent tools for usage recording and ranking
* Example usage-hint bundle
* `rapidfuzz` for lightweight fuzzy matching

Usage metadata is advisory only and does not change the underlying source of truth. Existing OKF bundles remain valid without `usage.yaml`.

Related to #319


sridhar-mani:
@nikhilk @libei  check this implementation out and please suggest some insights into the implementations if possible.

Issue #242, opened 2026-07-28
`stale_after` comparison has no timezone anchor, so staleness depends on where a bundle is read


erekola:
## Summary

§5.5 defines staleness as a date comparison:

> Optional. An absolute date (`YYYY-MM-DD`). A concept is stale when `today >= stale_after`. An absolute date, not a relative TTL, keeps the staleness decision a plain date comparison with no reference to when the concept was read.

§10.5 step 6 makes the same comparison load-bearing:

> **Gate**: refuse to display a failing attestation; warn or refuse when `today >= stale_after`.

"today" is not defined anywhere in the document.

## Why it matters

The reasoning in §5.5 is that an absolute date removes any dependence on when the concept was read. Without a timezone it introduces a dependence on where it is read instead. A bundle carrying `stale_after: 2026-12-31` is stale in Auckland while still fresh in Los Angeles, for about 26 hours in the general case. §10.5 turns that difference into a display-or-refuse decision, so two consumers acting correctly can disagree about whether the same value may be shown.

The reference implementation resolves it silently to the consumer's local timezone. `reference_agent/bundle/document.py`:

```python
return (today or date.today()) >= stale_after
```

`date.today()` is system-local. An independent consumer has no way to know that is the intended reading, and a different choice is equally defensible from the text.

## Suggestion

Name the timezone in §5.5. UTC is the obvious candidate. It costs one sentence. If the intent is deliberately that the consumer's local date governs, saying that explicitly is just as good, because the ambiguity rather than the choice is what makes this unimplementable consistently.

PR #323, opened 2026-08-20
okf: make every timestamp an ISO 8601 datetime with an explicit offset

amirhormati:
## Summary

OKF v0.2 mixed two time formats. `generated.at` and `verified[].at` were ISO 8601 datetimes, while `stale_after`, `sources[].last_modified`, and `usage_window.from`/`to` were bare dates. For `stale_after` that is ambiguous rather than merely untidy: `2026-12-31` names a different instant in every timezone, so the same bundle can be stale in one office and fresh in another. The v0.3 proposal already files this as an open issue (#242).

Every timestamp-valued key in OKF is now an ISO 8601 datetime with an explicit UTC offset.

## SPEC

The format is stated once, in the §5 preamble, and nowhere else. `stale_after`'s rule becomes `now >= stale_after`.

Consumers are expected to be strict rather than lenient about a value that arrives without an offset: a midnight-UTC fallback would quietly reintroduce the disagreement this change exists to remove. The reference agent ignores such a value, and the demo's storage layer rejects it outright (below).

The §9 log date headings are unchanged. `## 2026-05-22` groups a day's entries and is not a field value.

## Reference agent

This exposed a live bug. PyYAML follows YAML 1.1, so `safe_load` coerced both `2026-12-31` and `2026-06-30T14:00:00Z` into `date`/`datetime` objects, and `safe_dump` wrote them back as `2026-06-30 14:00:00+00:00`. The agent was already corrupting `generated.at` on any bundle it rewrote, which is visible in the committed `viz.html` files. Making `stale_after` a datetime would also have walked `is_stale` into a `date >= datetime` TypeError.

Dropping the timestamp resolver from the loader keeps every frontmatter value the string the author wrote, matching the YAML 1.2 core schema the TypeScript side already uses. `is_stale` then takes a `now: datetime` and rejects any value lacking a `T` separator or an offset.

## kcmd demo: native `datetime`

The six timestamp fields on the `okf` aspect now use Dataplex's native `datetime` type instead of `string`. Both relevant properties were confirmed against the API rather than assumed:

| input | result |
|---|---|
| `2026-12-31T00:00:00Z` | stored verbatim |
| `2026-07-10T21:15:20+00:00` | stored verbatim |
| `2026-07-10T21:15:20+05:30` | stored verbatim |
| `2026-12-31T00:00:00.500Z` | stored verbatim |
| `2026-12-31T00:00:00` (no offset) | rejected at write |
| `2026-12-31` (date only) | rejected at write |

So the round-trip stays lossless, and the format is enforced by the storage layer rather than by convention. A bundle carrying a bare date fails the push with `Text '2026-12-31' could not be parsed at index 10` and leaves no partial entry.

**Migration:** Dataplex accepts only backwards-compatible template changes, and a field's type is not one of them. A project that has run this demo before needs the `okf` aspect type dropped before `setup.ts` can recreate it. `cleanup.ts` already prints that command.

## Test plan

- [x] `pytest` in `okf/`: 39 pass, with `test_is_stale` covering the date-only and naive-datetime rejections
- [x] `bun test ./tests/libts/scenarios.ts`: 21 pass
- [x] No timestamp coercion or loss across all 54 frontmatter-bearing bundle documents (parse, serialize, re-parse)
- [x] All 63 documents across every bundle carry an explicit offset, so none is rejected by the stricter type
- [x] Live `hormati-bqml` / `us-central1` from an empty project: `acme_retail` push/pull is 17 files, 17 ok, 0 lossy
- [x] The default `ga4` bundle in `catalog/` is byte-identical on pull
- [x] `aspect-types describe okf` reports all six fields as `datetime`
- [x] A bundle with a bare date is rejected at push, with no partial entry created
- [x] Cloud state torn down

## Note

Regenerating `stackoverflow/viz.html` also picks up an unrelated tags fix that predates this change; the committed copy had `tags` as one joined string. Happy to split that out if preferred.


firat-elbey (approved):
LGTM

PR #331, opened 2026-08-22
Push OWL/extends class hierarchy to BigQuery Graph (labels + field flattening)

libei:
## What

Makes `Entity.extends` (from OWL `rdfs:subClassOf`, merged in #329) actually shape the BigQuery Graph push, so a class hierarchy is queryable.

Three mechanisms:

- **Labels mimic the hierarchy.** A subclass node table declares its own default label plus one `LABEL <Ancestor>` per transitive supertype, so `MATCH (:Person)` also matches `Customer`/`Manager` nodes.
- **Fields flow down (flattening); edges do not.** Supertype fields are copied onto subclasses (nearest-definition-wins), so each shared label's property signature matches — BigQuery requires an identical signature on every table that binds a shared label. Relationships are never inherited.
- **Abstract (table-less) superclasses.** An entity marked `abstract: true` has no physical table/source/key, produces no node table, and survives only as a `LABEL` on its concrete descendants (whose tables carry its flattened fields). An abstract class that is no concrete entity's ancestor is warned and dropped. `unbound:` remains a hard error, deliberately distinct from the explicit `abstract` marker (overloading it would silently drop an entity someone merely forgot to bind).

## Shared-label DDL constraints (discovered by live validation)

A label bound by more than one element table is a rigid contract in BigQuery. Live probing against `sqlgen-testing` pinned three rules the emitter now honors:

1. **No OPTIONS on a shared label.** A label carrying `OPTIONS(...)` "cannot be bound to another element table," so a supertype's own `DEFAULT LABEL` drops its description/synonyms (with a warning).
2. **Identical property signature everywhere.** Every table binding a shared label must list the same property set — which holds because supertype fields are flattened onto each subclass.
3. **No MEASURE on a shared label.** A measure cannot be replicated across the tables sharing a label ("defined as MEASURE, but there are other declarations with the same name"), so a metric targeting a supertype is skipped with a warning.

## Scope

- **BigQuery leg only.** Inheritance resolution lives in a new leg-agnostic pass (`resolve_inheritance.ts`) run only by the BigQuery generator; the Knowledge Catalog push output is unchanged.
- **No golden drift.** Existing corpus goldens are byte-for-byte unchanged (entities with no `extends` render identically).

## Example

A `Manager` (extends `Employee` extends `Person`) node table:

```
`…manager` AS Manager
  KEY(id)
  DEFAULT LABEL
  PROPERTIES(id, team_size, department, full_name OPTIONS(...), email)
  LABEL Employee PROPERTIES(id, department, full_name OPTIONS(...), email)
  LABEL Person   PROPERTIES(id, full_name OPTIONS(...), email)
```

## Tests

- `resolve_inheritance.test.ts` — single/multi/transitive parents, diamond dedup + ordering, child-override, cycle, unknown parent, edges-/keys-not-inherited.
- Abstract-elimination block in `bigquery.test.ts` — no node table for the abstract class, label + flattened fields on descendants, no-descendant warning.
- Supertype shared-label block in `bigquery.test.ts` — dropped label OPTIONS + skipped supertype metric, with a leaf metric still emitting.
- Bound `hierarchy_graph.yaml` fixture + `hierarchy_graph.bigquery.golden.sql` in the e2e corpus.
- Full suite: 443 pass, `tsc --noEmit` clean.

## Live validation

Pushed the full 5-class DDL to `sqlgen-testing` (accepted), then ran traversals:

- `MATCH (:Person)` → base + all three subclasses, inherited `full_name`/`email` resolve.
- `MATCH (:Employee)` → Employee + Manager (transitive).
- `MATCH (:Manager)` → Manager only, own `team_size` + inherited `department`/`full_name`.
- `livesIn` edge stays Person-only — edges are not inherited.

Scratch dataset cleaned up afterward.

### DIFF
--- a/toolbox/mdcode/docs/semantic-model.md
+++ b/toolbox/mdcode/docs/semantic-model.md
@@ -186,6 +186,26 @@ per supertype**, walking the full transitive chain. A node then matches its
 supertype in a query — `MATCH (:Person)` returns `Person`, `Customer`,
 `Employee`, and `Manager` nodes alike.
 
+You author only each entity's **own** fields plus the one `extends` keyword; the
+push does the rest:
+
+```yaml
+datasets:
+  - name: Person
+    source: proj.ds.person
+    primary_key: [id]
+    fields:
+      - {name: id,        expression: {dialects: [{dialect: BIGQUERY, expression: id}]}}
+      - {name: full_name, expression: {dialects: [{dialect: BIGQUERY, expression: full_name}]}}
+      - {name: email,     expression: {dialects: [{dialect: BIGQUERY, expression: email}]}}
+  - name: Customer
+    source: proj.ds.customer
+    primary_key: [id]          # each subclass keeps its OWN key; keys do not inherit
+    extends: [Person]          # the one keyword you add
+    fields:
+      - {name: loyalty_tier, expression: {dialects: [{dialect: BIGQUERY, expression: loyalty_tier}]}}
+```
+
 For those shared labels to work, **the supertype's fields flatten down** onto the
 subclass. BigQuery requires every table exposing a label to expose the **same
 property signature**, so a subclass's `LABEL Person` block must list exactly what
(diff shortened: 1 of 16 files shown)


libei:
Follow-up hardening pass (commit 8120b20) addressing self-review findings:

- **Inherited-expression localization** (`resolve_inheritance`): an inherited field written as `<Ancestor>.col` is rewritten to the descendant's local `col`, so subclass and supertype render the property identically — required because BigQuery enforces one identical definition per property across every element table binding a shared label.
- **Supertype-authoritative rendering** (`bigquery`): every inherited property is rendered from the supertype's canonical definition in both the DEFAULT LABEL and each `LABEL <ancestor>` block; a divergent subclass override (unrepresentable under a shared label) is warned and dropped. A metric targeting an *abstract* entity now reports an accurate "table-less supertype" message instead of the generic no-KEY one.
- **KC leg skips abstract entities** (`knowledge_catalog`): the Knowledge Catalog leg does not model inheritance, so an abstract (table-less) entity is skipped with a warning rather than published as an empty-resource entry; concrete subtypes publish normally.

Added resolve/bigquery/KC unit tests (+6) and a docs note. Existing goldens unchanged; `tsc` clean; 449 tests pass.

PR #355, opened 2026-08-29
fix(mdcode): tolerate un-addressable existing entry links on kc re-push

libei:
## Problem

`kcmd push --target kc` fails on **re-push** (once the schema-join entry
links already exist) with a 403 / `ENTRY_LINK_NOT_FOUND`:

```
entry link 'sales-orders-to-customer': ... Entry Link
'projects/.../entryLinks/sales-orders-to-customer' does not exist.
Error metadata code: ENTRY_LINK_NOT_FOUND
```

A first push succeeds; the second one fails.

## Root cause (verified live against the autopush EAP endpoint)

On a re-push, `createEntryLink` returns **409 ALREADY_EXISTS** (the server
dedups by source/target/link-type). `writeEntryLink` then falls back to
`updateEntryLink` to refresh the schema-join aspect.

But on some Dataplex surfaces — including the autopush EAP endpoint used
for the semantic-model EAP — entry links are reachable **only** via
`createEntryLink` and `:lookupEntryLinks`. The individual **get / update /
delete by resource name** methods return `NOT_FOUND` (404) or a masked
`PERMISSION_DENIED` (403) *even for a link that `:lookupEntryLinks`
returns*. So the in-place aspect update can never succeed there, and the
`!isOk(upd)` branch turned an idempotent re-push into a hard failure.

I confirmed this directly: `:lookupEntryLinks` returns
`sales-orders-to-customer`, while `GET`/`PATCH`/`DELETE` on that exact
resource name all return `ENTRY_LINK_NOT_FOUND`. `ListEntryLinks` is
likewise "not visible to labels: {PUBLIC}".

## Fix

A 409 means the link is already present with its correct, **immutable**
endpoints and link type — the only thing a re-push could change is the
schema-join aspect. So a best-effort aspect update whose failure is the
"link not addressable" class (404 / 403) is treated as a successful no-op
instead of an error. Genuine failures (any other status) still fail the
push. Mirrors how `deleteOwnedLinks` already tolerates a 404 on delete.

## Testing

- New unit tests: an existing link whose aspect update returns 404 (and
  403) is a no-op success (`linked: 1`, push succeeds).
- Full suite: 498 pass.
- Live-validated end to end against the autopush endpoint: the re-push
  that previously failed now reports `linked 2 relationships` and exits 0.

### DIFF
--- a/toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
+++ b/toolbox/mdcode/src/libts/semantic/deploy_knowledge_catalog.ts
@@ -642,11 +642,19 @@ async function createEntryLinks(
   return {linked: links.length};
 }
 
-// Writes one entry link: create, then fall back to an in-place aspect update if
-// it already exists. A link's entry references and type are immutable, so a
-// re-push only refreshes the aspect (the join detail). A relationship whose id
-// changed writes a new link and leaves the old one; reconcileLinks deletes such
-// orphaned links after this model's links are written.
+// Writes one entry link: create, then fall back to a best-effort in-place aspect
+// update if it already exists. A link's entry references and type are immutable,
+// so a re-push only refreshes the aspect (the join detail). A relationship whose
+// id changed writes a new link and leaves the old one; reconcileLinks deletes
+// such orphaned links after this model's links are written.
+//
+// A 409 means the link is already present with its correct (immutable) endpoints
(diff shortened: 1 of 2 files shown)


libei:
Closing: the fix references autopush (internal-only) behavior in code comments, which shouldn't be documented in a public repo. Will revisit with a public-safe framing if/when needed.

PR #359, opened 2026-08-29
mdcode: merge a binding profile onto the logical model at push time

libei:
Second of three stacked PRs implementing **binding profiles** (design: \`docs/semantic-model/profiles.md\`).

> **Stacked on #358.** Until #358 merges, this PR's diff also shows #358's commit. Review #358 first; the first commit here is that PR.

This adds the actual mechanism: one **logical model** paired at push time with a chosen **binding profile**, deployed to BigQuery + Knowledge Catalog exactly as before. Deploy legs are unchanged — this is loader/push work.

## What changes

- **Profile discovery** — \`SemanticModelLayout\` reads \`<model>.profiles/*.yaml\` beside the model. The model glob is single-level, so a profile file is never mistaken for a model.
- **Merge** (\`resolve_profiles.mergeProfile\`) — overlays a profile onto the logical model **by name** (entity \`source\`, field \`expression\`, \`deployment_target\`) and enforces the binding-only contract: a profile may set physical bindings and may leave a field unbound; it may not add/remove elements or change what anything means. A field the profile omits is carried through as unbound.
- **Availability pruning** (\`resolve_profiles.pruneUnavailable\`) — drops each unbound field and everything that depends on it: a metric that reads it, a relationship whose join column is unbound, a cross-entity metric over a dropped relationship. Returns a per-profile report of withheld coverage.
- **\`kcmd push --profile <name>\`** — merges the named profile before load, prunes after, then runs the existing validation + deploy. Orthogonal to \`--target\`. The implicit \`default\` profile keeps today's inline single-file behavior untouched.
- **Resource-name URI sources** — a \`//bigquery.googleapis.com/...\` table normalizes to \`project.dataset.table\`; a non-BigQuery URI (Spanner/AlloyDB/…) rides through and is skipped by the BigQuery pre-flight probe (it validates and reports, but only BigQuery is an execution backend).

## Test

- \`resolve_profiles.test.ts\` — by-name merge, contract violations, omission→unbound, non-mutation; field/metric/relationship pruning + report.
- \`semantic_model_layout.test.ts\` — profile discovery; a \`.profiles/\` dir is not discovered as a model.
- End-to-end run of the guide's three-file example: \`--profile analytical\` prunes \`availableCredit\` and emits DDL from the BigQuery-URI sources; \`--profile operational\` binds \`availableCredit\`, withholds \`avg_lifetime_value\`, and reports why.
- \`npx tsc --noEmit\` clean; full semantic suite green (513 pass).

\`kcmd profiles\` (list + report) and \`default_profile\` in \`catalog.yaml\` follow in the next PR.

### DIFF
--- a/toolbox/mdcode/src/libts/layouts/semantic-model.ts
+++ b/toolbox/mdcode/src/libts/layouts/semantic-model.ts
@@ -91,6 +91,35 @@ export class SemanticModelLayout implements CatalogLayout {
     return docs;
   }
 
+  // The binding-profile documents beside a model: `<model>.profiles/*.yaml`
+  // under the model's EntryGroup dir. Each is a partial `semantic_model`
+  // document carrying physical bindings that `--profile <name>` merges onto
+  // the logical model; `name` is the file basename (the profile name). The
+  // directory is not globbed for model documents (init's `*.yaml` glob does
+  // not descend), so a profile file is never mistaken for a model.
+  profilePaths(model: string): {name: string; path: string}[] {
+    if (!this._entryGroup) return [];
+    const dir = path.join(
+        this._catalogPath, 'EntryGroups', this._entryGroup,
+        `${model}.profiles`);
+    if (!fs.existsSync(dir)) return [];
+    const out: {name: string; path: string}[] = [];
+    for (const entry of fs.readdirSync(dir)) {
+      if (!entry.endsWith('.yaml')) continue;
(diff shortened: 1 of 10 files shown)


libei:
Folding into #358, which now carries the whole binding-profiles feature as one PR (four commits) after end-to-end live validation. Closing to keep a single review thread.

PR #361, opened 2026-08-29
feat(mdcode): add Spanner Graph as a kcmd push target

libei:
## What

Adds `--target spanner` to the semantic-model `kcmd push`, alongside the existing BigQuery Graph and Knowledge Catalog legs. A model deploys to Spanner Graph by declaring a `//spanner.googleapis.com/projects/<p>/instances/<i>/databases/<db>/propertyGraphs/<g>` deployment target; the target's host selects the leg, so **the same authored model that deploys to BigQuery Graph deploys to Spanner Graph unchanged**.

```
kcmd push --target spanner        # deploy the Spanner-targeting model(s)
kcmd push --target bq,spanner     # both graph backends
kcmd push                         # all: bq, spanner, kc
kcmd push --target spanner --print --validate-only   # inspect the DDL, deploy nothing
```

This complements the binding-profiles work: a profile chooses the *physical binding*; `--target spanner` now provides the *Spanner deploy backend* a Spanner-bound profile lowers to.

## How Spanner Graph differs from BigQuery Graph (and how this handles it)

- **Bare table names.** A property graph lives inside one Spanner database and names tables in that same database, so the generator maps each entity's `dataSource` to its final segment (`proj.ds.Orders` → `Orders`); the graph name is bare too (from the URI's `propertyGraphs/<g>`).
- **No `MEASURE`.** Spanner Graph has no measures, so model-level metrics are dropped with a warning — the graph structure (nodes, edges, labels, properties, inheritance labels) still deploys.
- **No per-element `OPTIONS`.** description/synonyms are not emitted; they ride into Knowledge Catalog instead (mirroring how BigQuery's graph-level OPTIONS is dropped).
- **Async DDL.** Applied via the Spanner Admin `updateDatabaseDdl` long-running operation, polled to completion (vs BigQuery `jobs.query`).

## Structure

| File | Role |
|---|---|
| `deployment_target.ts` (new) | Shared GOOGLE deployment-target reader; classifies each URI as a BigQuery or Spanner Graph target. `deploy_bigquery.ts` re-exports the BQ views, so existing importers are unchanged. |
| `spanner.ts` (new) | Spanner Graph DDL generator — sibling to `bigquery.ts`. |
| `gcp/spanner.ts` (new) | `SpannerClient` (`updateDatabaseDdl` + `getOperation`). |
| `deploy_spanner.ts` (new) | Spanner deploy leg — sibling to `deploy_bigquery.ts`. |
| `commands.ts` | Routes each model to the leg matching its target; the BigQuery data-source pre-flight now runs only over BigQuery-targeting models. |
| `validate.ts` | Accepts a BigQuery **or** Spanner target; the metric-resolves-to-one-entity rule stays BigQuery-only. |

## Design decisions (flagging for review)

1. **A model declares exactly one deployment target** (unchanged rule); its host selects the leg. A model can't target both BQ and Spanner in one document today — that fits the one-binding-per-profile model. Easy to relax later.
2. **Source → Spanner table = final dotted segment.** The loader does not yet parse `//spanner.../tables/<T>` source URIs (that's the proposed profiles feature), so the existing `project.dataset.table` source is reduced to its table name. One authored model works for both backends.
3. **Metrics dropped, not errored,** for Spanner (warned per metric).

## Tests

- `spanner.test.ts` — generator units + M:N association golden.
- `spanner.e2e.test.ts` — golden DDL for star / hierarchy / fan-out fixtures (shared corpus with the BQ suite, so the two goldens sit side by side).
- `deploy_spanner.test.ts` — target parsing + deploy leg with a mocked client (op polling, error, exhaustion, no-target).
- `validate.test.ts` / `target.test.ts` — Spanner target acceptance and `--target` resolution.

`tsc --noEmit` clean; full `bun test` suite green (537 tests); CLI binary builds.

## Docs

This PR now also documents the Spanner Graph target across the semantic-model guide (it previously shipped code only):

- **Deploy guide** (`docs/semantic-model/README.md`) — push reframed as "a property graph (BigQuery **or** Spanner, selected by the target host) + Knowledge Catalog"; adds the Spanner target URI form, the bare-table `source` reduction, and `--target spanner`.
- **Reference** (`docs/semantic-model/reference.md`) — `--target bq|spanner|kc|all`; a new **What gets created in Spanner** section (bare names, no `MEASURE`, no `OPTIONS`, async `updateDatabaseDdl`); validation now accepts a BigQuery **or** Spanner target and scopes the metric-resolves-to-one-entity rule and the live source pre-flight to BigQuery-targeting models; Spanner IAM permissions.
- **Fidelity** (`docs/semantic-model/fidelity.md`) — a new **To Spanner Graph** section (structure kept, metrics/OPTIONS dropped) plus a note that the BigQuery-graph matrix column is BigQuery-specific.
- **Codelab** (`docs/semantic-model/codelab.md`) — a runnable step that swaps the same model's target to Spanner and prints the generated DDL under `--validate-only`.
- Top-level `toolbox/mdcode/README.md` and the `docs/semantic-model.md` index updated to match.

Docs-only commit; the code, tests, and `bun test` results above are unchanged.

### DIFF
--- a/toolbox/mdcode/README.md
+++ b/toolbox/mdcode/README.md
@@ -171,22 +171,26 @@ kcmd status
 kcmd push
 ```
 
-For the semantic-model scope, `kcmd push` deploys the model's BigQuery Graph
-(`CREATE OR REPLACE PROPERTY GRAPH`) to the project and dataset named by the
-model's GOOGLE deployment target; pass `--validate-only` to print the generated
-DDL without executing it. This path additionally reads the target dataset's
-metadata (`bigquery.datasets.get`) to pin the query job's processing location.
-The call degrades gracefully when the permission is absent, so a narrower
-service account still works but falls back to BigQuery's own location inference.
-A dataset `source` that omits its project is qualified with the scope's
-declared project (the `<projectId>` in the init scope), not the ambient
-`gcloud` project; write a fully-qualified `project.dataset.table` source when
-the tables live elsewhere.
+For the semantic-model scope, `kcmd push` deploys the model as a property graph
+(`CREATE OR REPLACE PROPERTY GRAPH`) to the backend named by the model's GOOGLE
+deployment target — **BigQuery Graph** or **Spanner Graph**; the same authored
+model deploys to either, and swapping the target URI switches backends. Pass
+`--validate-only` to print the generated DDL without executing it. For a BigQuery
+target this path additionally reads the target dataset's metadata
+(`bigquery.datasets.get`) to pin the query job's processing location; the call
+degrades gracefully when the permission is absent, so a narrower service account
+still works but falls back to BigQuery's own location inference. A dataset
+`source` that omits its project is qualified with the scope's declared project
+(the `<projectId>` in the init scope), not the ambient `gcloud` project; write a
+fully-qualified `project.dataset.table` source when the tables live elsewhere.
+(A Spanner target names its tables inside one database, so a `source` is reduced
+to its final segment, and Spanner Graph has no `MEASURE`, so metrics are dropped
+with a warning.)
 
 `kcmd push` also deploys the model to Knowledge Catalog (entries for the model,
 its entities and metrics, and `schema-join` links for its relationships). Use
-`--target bq|kc|all` (default `all`) to choose destinations, `--print` to dump
-each destination's generated artifact, and `--force-remove` to delete models
+`--target bq|spanner|kc|all` (default `all`) to choose destinations, `--print` to
+dump each destination's generated artifact, and `--force-remove` to delete models
 left in the entry group that the push no longer includes. See
 [docs/semantic-model/](docs/semantic-model/README.md) for the full guide,
 including how re-push reconciles removed entities, metrics, relationships, and
--- a/toolbox/mdcode/docs/semantic-model.md
+++ b/toolbox/mdcode/docs/semantic-model.md
@@ -8,8 +8,8 @@ now split by what you came to do:
 - **[End-to-end codelab](semantic-model/codelab.md)** — author, govern, hydrate,
   and query one model, start to finish.
 - **[Reference](semantic-model/reference.md)** — every flag, what push creates in
-  BigQuery and Knowledge Catalog (including class hierarchies), validation, and
-  permissions.
+  BigQuery Graph, Spanner Graph, and Knowledge Catalog (including class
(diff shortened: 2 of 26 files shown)


libei:
### Live-validated against real Spanner ✅

Deployed end-to-end to a throwaway **ENTERPRISE**-edition Spanner instance (`sqlgen-testing`, since deleted): `kcmd` loader → generator → `deploySpanner` → real `updateDatabaseDdl`, then confirmed the graph exists (`INFORMATION_SCHEMA.PROPERTY_GRAPHS`) and is queryable via GQL:

```
GRAPH commerce_graph
MATCH (o:orders)-[:orders_to_customer]->(c:customer)
RETURN o.OrderId AS order_id, c.Name AS customer_name
-- 10, Ada
```

The live run caught two bugs the mocked tests couldn't (fixed in the follow-up commit, with a regression test):
1. `updateDatabaseDdl` is a **PATCH** on `.../ddl`, not a POST (POST 404s).
2. Spanner's DDL parser **rejects a trailing `;`** in a statement (BigQuery's `jobs.query` accepts it) — the leg now strips it for the wire while the printed/golden DDL keeps it.

Note for reviewers/users: **Spanner Graph requires the ENTERPRISE edition** — a STANDARD instance fails with `Feature GRAPH is not available … minimum required Edition … ENTERPRISE`.

libei:
Addressed code-review findings in 5e47952:

1. **KC dropped the Spanner target** — `modelAspectData` read only BigQuery targets, so a Spanner-bound model recorded no `deploymentTargets` in Knowledge Catalog and pulled back UNBOUND. Now records both BigQuery and Spanner target URIs.
2. **Silent no-op success** — an explicitly named graph leg (`--target spanner`/`bq`) that no model targets now hard-errors instead of reporting success having deployed nothing. A leg pulled in only by `all`/default still skips quietly.
3. **Cross-target pre-flight** — the BigQuery source reachability probe now runs only when the BigQuery leg is actually requested, so a Spanner-only or kc-only push no longer probes BigQuery tables for a target the user didn't select.
4. **Quoted-identifier mangling** — bare-table extraction is now quote-aware; a backtick-quoted table name containing a dot is no longer split.
5. **Redundant parsing** — `googleDeploymentTargets` is memoized per model (WeakMap) so the several callers in one push share a single parse.

Unit coverage added for each. `tsc --noEmit` clean; full suite 532 pass.

PR #364, opened 2026-08-30
docs(mdcode): codelab deploys to Spanner via a binding profile

libei:
Reworks the semantic-model codelab and binding-profiles guide to show the Spanner Graph target (#361) and binding profiles (#358) working together.

> **Depends on #363** (now merged into `main`): the Spanner physical-column fix that lets remapped key/join columns deploy. This branch contains the docs commit only.

## Codelab (`docs/semantic-model/codelab.md`)

- **Step 1** names the deployment target with the first-class `deployment_target:` key instead of a `GOOGLE custom_extensions` JSON string.
- **Section 5** replaces the `sed` deployment-target swap with a real **binding profile** (`sales.profiles/operational.yaml`): a Spanner store with different table names, remapped columns (including keys), and an unbound `net_amount` so `revenue` is withheld operationally. It shows `kcmd profiles`' availability report, the generated Spanner DDL, a live deploy, and a GQL query.
- **Cleanup** drops the Spanner database.

Every command output in section 5 was captured from a live run against an ENTERPRISE Spanner instance.

## Binding profiles (`docs/semantic-model/profiles.md`)

- Scope note updated: Spanner graphs now deploy (was "non-BigQuery not yet supported").
- Spanner sources use **bare table names**, not `//spanner.../tables/X` resource URIs (which a Spanner graph cannot reference); added a note on the rule in "Sources are URIs".
- `--target` documented as `bq|spanner|kc|all`; command examples deploy the operational (Spanner) profile.

## Why now

The codelab previously swapped backends with `sed`, which is exactly the manual step binding profiles remove. This makes the codelab demonstrate the feature as intended: one logical model, an analytical (BigQuery) binding and an operational (Spanner) binding, switched with `--profile`.

### DIFF
--- a/toolbox/mdcode/docs/semantic-model/codelab.md
+++ b/toolbox/mdcode/docs/semantic-model/codelab.md
@@ -1,9 +1,10 @@
 # End-to-end codelab: one semantic model, from authoring to query
 
-A self-contained walkthrough. You author one semantic model, govern it in
-Knowledge Catalog, hydrate its data, and query a metric through BigQuery Graph.
-The metric is defined once and computed the same way at every step. The expected
-output is shown after each command so you can check as you go.
+A self-contained walkthrough. You author one *logical* semantic model, govern it
+in Knowledge Catalog, then bind it to physical stores — BigQuery and Spanner —
+and query the same metric in each. The metric is defined once and computed the
+same way at every step. The expected output is shown after each command so you
+can check as you go.
 
 For the deploy mechanics on their own (author, push, update, pull), see the
 [deploy guide](README.md); for every flag and permission, see the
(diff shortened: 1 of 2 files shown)


libei:
### Update: verified live, `entities:` throughout, and a dependency

Ran the whole codelab end to end against **live instances** — Knowledge Catalog on the autopush instance, live BigQuery, and a live ENTERPRISE Spanner instance. Reconciled the doc with what the commands actually print, and switched the authored YAML to `entities:` to match the prose.

**New dependency: #368.** Step 2 (govern the logical model *before* binding it) failed against the live server — the `semantic-entity` aspect template requires `source.resources`, but the emitter omitted the `source` block for a logical-only entity (`Required field missing source`). #367 verified this only with `--validate-only`, so it was never caught live. #368 fixes the emitter (`source: {resources: []}`); with it, `kcmd push --target kc` on the logical model writes all 5 entries and `kcmd pull` round-trips.

**Doc changes in the new commit:**
- Author the model and both profiles under `entities:` instead of `datasets:` (the prose already calls them entities). The two keys are aliases; the OWL-import block still shows `datasets:`, and a note flags it.
- Step 2's KC plan shows the real output: the `Validating...` framing, the two relationship-name-normalization warnings (with a note that they're expected and repeat on every push), and the completion footer.
- Step 3's BigQuery DDL block shows the `Pushing...` line, `-- BigQuery Graph --` banner, and target-URI comment it was missing — matching step 4's Spanner block and the real console.

**Live results (all matched the doc after `$PROJECT`/`$DATASET` substitution):** BigQuery measure query `Acme 290 / Globex 40` vs. the hand-join's wrong `380 / 120`; `kcmd profiles` withholds `revenue` on the operational profile; Spanner GQL returns `Acme 2 / Globex 1`. All test resources cleaned up afterward.

PR #371, opened 2026-08-30
fix(mdcode): correct the OWL-import UNBOUND push hint for --target kc

libei:
## What

`#367` let a purely logical model push to Knowledge Catalog (`kcmd push --target kc`), but the `kcmd owl import` hint printed after conversion still said:

> `kcmd push` is rejected until you bind each entity's source table and add a BigQuery deployment target -- validation needs both, for every --target.

That is now false for `--target kc`. This corrects the printed note and the two places that reproduce it verbatim:

- `src/tool/commands.ts` — the note printed by `owl import`.
- `docs/semantic-model/codelab.md` — the reproduced console output (which otherwise contradicts the same codelab's "govern the model right now with `--target kc`" step).
- `tests/libts/semantic/owl_converter.test.ts` — the behavior-spec comment.

New wording:

> `kcmd push --target kc` publishes it to Knowledge Catalog as-is. To deploy a graph, bind each entity's source table and add a deployment target first.

## Notes

- No logic change — only a printed string, a doc block, and a comment.
- The developer-guide prose for this (`owl-import.md`, `reference.md`) is corrected separately in the docs PR #365.
- Dependencies were not installed in this environment, so the suite was not run locally; the change touches no code paths.

### DIFF
--- a/toolbox/mdcode/docs/semantic-model/codelab.md
+++ b/toolbox/mdcode/docs/semantic-model/codelab.md
@@ -139,8 +139,8 @@ kcmd owl import /tmp/parts.ttl --out /tmp/parts_osi.yaml
 converted 2 classes, 1 object property, 2 datatype properties
 wrote /tmp/parts_osi.yaml
 note: this model is UNBOUND (placeholder `unbound:` sources, no deployment target).
-      `kcmd push` is rejected until you bind each entity's source table and add
-      a BigQuery deployment target -- validation needs both, for every --target.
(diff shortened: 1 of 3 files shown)


libei:
Superseded by #374, which reframes this note around the logical model instead of keeping the "this model is UNBOUND" lead. #374 updates the same three sites (commands.ts note, codelab output, test comment) plus the surrounding codelab section. Closing to avoid a conflicting edit; happy to reopen if preferred.

PR #396, opened 2026-09-05
feat(mdcode): tighten the semantic-model spec (version gating, fail-fast, unbound-by-omission)

libei:
## What

Tightens the mdcode semantic-model loader and tooling to the behaviors we
settled in review, and updates the user guides to match. Each settled item is
its own commit.

### Settled behaviors implemented

1. **Version is required.** Every document must declare `version`. A missing or
   unknown version is a hard error (not a warning). Two accepted values:
   `0.2.0.dev0` (vanilla Apache Ossie) and `0.2.0.dev0/google` (kcmd's extended
   profile).
2. **Fail-fast validation.** Duplicate names throw, an unknown `extends`
   supertype throws, and every object is `.strict()` so an unknown/misspelled
   key is an error.
3. **No `unbound` flag.** A field is unbound exactly when it has no
   `expression`. See the design note below.
4. **Measures only on leaf types.** A supertype that is extended by another type
   may not carry measures.
5. **`entities:` alias gated to `/google`.** The alias for `datasets:` is
   accepted only under the extended profile, and pull emits `entities:` (and
   the `/google` version).
6. **OWL importer is import-only.** OWL/RDF-specific constructs are no longer
   carried through as round-trippable custom extensions; pull never re-emits
   them.
7. **Version gates the extension surface.** Vanilla accepts `datasets:` +
   `custom_extensions:` and rejects native keys; `/google` accepts the native
   keys (`entities`, `extends`, `abstract`, `deployment_target`) and rejects
   `custom_extensions:`.

Open/unsettled items from the thread are intentionally left untouched.

### Design note (the one non-trivial interpretation)

Removing the `unbound` flag (#3) collapses "a field you forgot to bind" and "a
field this store legitimately lacks" into one syntactic thing: a field with no
`expression`. The two ways to resolve that:

- **(A)** an expression-less field is a hard error on a graph leg (unbound
  fields allowed only on a Knowledge-Catalog-only push), or
- **(B)** an expression-less field is unbound and pruned before generation.

This PR takes **(B)**. It preserves the binding-profiles value prop -- one
logical model serving several stores, each binding a different subset of
columns, with a metric reported unavailable where its field is unbound -- which
(A) would remove. A profile is authoritative: selecting it clears the base
model's inline bindings, so a field the profile does not re-declare is unbound
under it. The only binding-completeness error left in strict/graph mode is a
non-abstract dataset with no `source` at all.

## Testing

`bun test` -- 585 pass, 0 fail across 30 files. (Local `tsc` typecheck is
unavailable in this environment -- TypeScript is not installed offline -- but
`bun` transpiles and runs every touched file.)

### DIFF
--- a/toolbox/mdcode/docs/semantic-model/README.md
+++ b/toolbox/mdcode/docs/semantic-model/README.md
@@ -82,7 +82,7 @@ business and nothing physical — the entities, their fields, the relationships
 between them, and the metrics computed over them:
 
 ```yaml
-version: "0.2.0.dev0"
+version: "0.2.0.dev0/google"
 
 semantic_model:
   - name: sales                      # keep equal to the <model>.yaml filename (pull round-trips to that name)
@@ -111,7 +111,7 @@ semantic_model:
 Field and relationship names are the business vocabulary — `order_id`,
 `placed_by` — never physical column names; the physical binding comes later. A
 metric's `expression` may be a bare formula over the logical fields or the fuller
-per-dialect form. `entities` may also be written `datasets`.
+per-dialect form. `entities` may also be written `datasets` (the two are interchangeable under the `/google` version).
 
 Entities can **extend** other entities (`extends: [Parent]`); push flattens the
 supertype's fields down and expresses the hierarchy as graph labels, so a query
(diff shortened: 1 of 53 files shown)


libei:
Addressed the code-review findings in ce98752:

- **OWL unknown supertype** — a `rdfs:subClassOf` whose superclass is not an `owl:Class` in the ontology is now dropped with a warning (the loader hard-fails on an unknown supertype, so recording it would emit an unpushable model). Known parents in a mixed list are kept.
- **Profile inline-binding strip** — `stripInlineFieldExpressions` now clears inline bindings only for the models the profile names; a model the profile does not target keeps its bindings.
- **Leaf-measure check** — the non-leaf set is now built from table-emitting entities, so an abstract / empty-KEY subtype no longer marks its parent non-leaf.
- **Stale comments** — corrected to match shipped behavior.

Two findings were reviewed and left out of scope as narrow edges unrelated to the settled agreements: (a) a `deployment_target` on a vanilla-version profile failing `.strict()` load (a user error the version gate correctly rejects), and (b) sibling extension data dropped without a warning on `osi_converter` pull. Happy to fold either in if you'd prefer.

PR #399, opened 2026-09-07
docs(mdcode): a user guide for modeling write operations

libei:
Stacked on #398, which is stacked on the actions prototype, #346. The diff
against `main` therefore includes those two; the new material is
`docs/semantic-model/actions.md` plus two lines in
`docs/semantic-model/README.md`.

The guide is the missing page between the deployment README, which introduces
actions and constraints in two short sections, and `demo/action/`, which runs
them but assumes you already know what they are. It walks one model end to
end: declare an action, state the rules as constraints, push it, run it
against a live store, and expose it to an agent. It also states what the
prototype does not do yet — no per-action `precondition` or `affects`, no
executor dispatch, a Spanner write path, and no CLI command that runs an
action.

Verification: the example model was run through the loader and the constraint
evaluator on this branch. It loads with no warnings, its parameters type as the
guide says (two object references and one scalar), and all four constraints
lower to probes. The grammar limits, the mixed AND/OR precedence, and the NULL
reading were confirmed the same way, with negative cases for cross-entity
comparisons, aggregates, and function calls. Every quoted output is copied from
`demo/action/README.md`.

### DIFF
--- a/toolbox/mdcode/demo/README.md
+++ b/toolbox/mdcode/demo/README.md
@@ -256,3 +256,26 @@ bun pull.ts --bundle /tmp/acme_pulled
 ```bash
 bun cleanup.ts
 ```
+
+
+## Actions and Constraints
+
+This demo shows the write side of a semantic model: an action that moves money
+between accounts, four constraints that say what has to stay true, and a runtime
+that applies the write in a Spanner transaction and rolls it back if any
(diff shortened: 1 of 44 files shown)


libei:
Superseded: the guide is split to match the PR split. The actions-only page now rides #346 (commit 0c70399), and the constraint gate is documented on the constraints PR instead, so neither page describes something its own PR does not ship.

PR #409, opened 2026-09-11
mdcode: enforce action constraints at runtime, with a runnable credit-flow demo

libei:
A semantic model has always described what things mean. Actions describe what
can be done and constraints describe what must stay true, but until now nothing
checked a constraint when an action ran, and a constraint could not read an
action's arguments at all. This adds the runtime that does both, and a demo that
runs it end to end against a live Spanner database.

## Reading order

The six commits are meant to be read in order, and the first three are the ones
that matter for the library. The last three are the demo and can be skimmed.

| Commit | What it does | Library lines |
| --- | --- | --- |
| `43761bf` | Ports the semantic runtime onto main's action model. | ~750 |
| `db7ca5b` | Adds the `sql` executor and lets a constraint read action parameters. | ~1400 |
| `62efb35` | Checks parameter guards before the write; runs the declared DML. | ~150 |
| `3f83743` | The demo, plus 14 runtime tests and four Spanner admin methods. | ~115 |
| `62748c5` | README wording. | 0 |
| `d0a861c` | Keeps the ADK out of the manifest. | 0 |

Everything under `toolbox/mdcode/demo/credit/` is demo code. It is not imported
by the CLI and sits outside the tsconfig `include`.

## What the runtime does

`runAction` runs five steps, all but the first inside one Spanner read-write
transaction:

1. **Resolve.** An entity-typed argument is an object reference, so `--order
   12345` is looked up and turned into the row it denotes. "No such order" and
   "more than one such order" both stop the call.
2. **Guard.** A constraint whose predicate reads an action parameter describes
   the proposed call rather than the stored data, so it is checked before the
   write with the arguments bound. Once the write has happened there is no
   longer an `amount` for it to read.
3. **Apply.** The action's writes run inside the transaction.
4. **Gate.** Every constraint over stored state is lowered to a probe and run in
   the same transaction, so it observes the uncommitted rows.
5. **Decide.** `reject` rolls back and returns the constraint's own description.
   `escalate` rolls back and names what an approver must sign off, re-submittable
   with `approvals`. `warn` commits and is reported.

The gate fails closed. A constraint that cannot be lowered aborts the action
before a transaction is opened, because an unevaluated invariant cannot be told
apart from a satisfied one. Probes wrap the predicate in `COALESCE(..., FALSE)`
so a NULL answer is a violation.

## The `sql` executor

An action can now carry its own DML. The other executor kinds name a write that
commits inside a system this runtime does not control, so a check around one is
advisory; a statement run in the runtime's own transaction can be rolled back,
and it runs where the probes run. `runAction` refuses a non-`sql` executor
unless the caller supplies a handler that performs the write as DML in the open
transaction.

`validate.ts` enforces the narrowness that makes this safe: one DML verb per
statement, no `;`, and every `@parameter` either declared by the action or
generated by the runtime. A row the action creates gets a runtime-generated key,
because a caller that picks its own primary key can overwrite a row that already
has it.

## Verified live

Against `sqlgen-testing / graph-unified-solution-demo / semantic_credit_demo`,
with a model carrying three constraints:

- A $30 credit on a $147.85 order escalates on the review threshold and rolls
  back; re-submitted with the approval it commits, and the order becomes $117.85
  across five lines.
- A $30 credit on an $18.00 order escalates on two constraints at once.
- A $10 credit on a $200.00 order commits with no review.
- A caller that writes its own SQL, inserting the credit line and forgetting to
  recompute the order total, is rejected by the invariant and rolled back. The
  order is unchanged afterwards.

The demo also ships a Google ADK 2.0 agent that calls `runAction` in process.
Told "give him a 60 dollar credit right now, I authorize it", it came back with
a review request it could not grant. A caller asserting its own authority does
not get past a gate that lives in the transaction.

`toolbox/mdcode/demo/credit/README.md` is the runbook for all of the above.

## Tests

774 pass, up 14. The new ones cover guard-before-write ordering, the `escalated`
outcome and `approvals` (including that approving one constraint does not carry
over to another), executor parameter binding and generated keys, and the refusal
of an executor the runtime could not roll back.


libei:
Two commits added for review of the end-to-end flow, ahead of building it.

`toolbox/mdcode/demo/credit/PIPELINE.md` is the flow as a reader would run it,
command by command from an empty directory: `kcmd init` the scope, author the
logical model plus a Spanner binding profile, `kcmd push --profile` to send the
graph to Spanner and the model, its action and its rules to Knowledge Catalog,
`kcmd init --pull` into a fresh workspace to see what a consumer can read back,
then run the CLI and the agent against the catalog scope. Its last section says
which steps ship today and which are being built.

The other commit closes a gap that flow depends on. A `semantic-constraint`
aspect carried the expression but not the severity, and the OSI serializer
dropped it too, so every rule round-tripped as a plain refusal and the
supervisor-review half of the demo disappeared. The pull golden now shows
`severity: escalate` surviving push and pull.

libei:
Reworked the flow so it runs on commands rather than on demo TypeScript. `demo/credit/PIPELINE.md` is the thing to read.

Nine steps, and every one of them is a `kcmd` command, a `gcloud` command, or a file you author. What the demo still owns is five files, all readable in full: the logical model, its Spanner binding profile, `schema.sql`, `seed.sh`, and the agent.

The mapping out of TypeScript:

| Was | Becomes | Status |
| --- | --- | --- |
| `setup.ts`, 146 lines | `schema.sql` via `gcloud spanner databases create --ddl-file`, and `seed.sh` | done in this commit, `setup.ts` deleted |
| `credit.ts`, 247 lines | `kcmd action run IssueCredit --arg order=12345 --arg amount=30` | runtime ships here, command to build |
| `credit.ts --list` | a `SELECT` through `gcloud spanner databases execute-sql` | nothing to build |
| `credit.ts --by-hand` | an `INSERT` through `gcloud` | nothing to build |
| `config.ts`, `model.ts` | the workspace `kcmd init --pull` creates | to build |
| `cleanup.ts` | `gcloud spanner databases delete` | nothing to build |

`seed.sh` is a script rather than a `.sql` file because `gcloud` reads DDL from a file and has no equivalent for DML. Its body is six visible `gcloud ... execute-sql` calls, so any one of them can be copied out and run alone.

Two changes to the flow itself since the last draft, both worth a look:

**The invented `--scope` flag is gone.** Step 5 does `mkdir agent-workspace && cd agent-workspace && kcmd init --semantic-model ... --pull`, and everything after runs in that directory. `kcmd action run` and the agent read the model the way every other `kcmd` command does, out of `catalog.yaml`. That proves the catalog round trip by construction instead of asserting it, and it avoids adding a flag that exists only for this demo.

**New step 7 writes to Spanner directly.** A raw `gcloud` INSERT adds a credit line to order 12346 without recomputing the total, and it succeeds, because nothing was asked. The next legitimate `kcmd action run` against that order is then refused by the invariant, with no approval offered. That is the argument for giving an agent the action and withholding the database credentials, shown rather than claimed.

`npx tsc --noEmit` clean, 776 pass / 0 fail.

libei:
Unpacked the agent. `demo/credit/AGENT.md` builds it from nothing in five steps and reports what each one cost.

The answer to "I hope it's a simple process": it is, and here is the measurement.

| Step | Code | Whose work |
| --- | --- | --- |
| Imports | 6 lines | boilerplate |
| 1. Load the model | 4 lines | same for any model |
| 2. Derive the tools | 4 lines | same for any model |
| 3. Adapt them to ADK | 18 lines | written once per framework |
| 4. Write the persona | 13 lines | yours, and should stay yours |
| 5. Run it | 23 lines | ADK boilerplate |

68 lines total. No line of code names a credit, an order, a threshold, a table or a column. Two lines carry anything domain-shaped, both inside the persona.

**What made it simple.** The hand-written tools are gone. `src/libts/semantic/agent_tools.ts` derives them: `actionTools()` gives one write tool per action, `entityTools()` one lookup tool per entity from the binding profile. Deriving `issue_credit` needs nothing invented, because the model already said all of it — the description, the `ai_context.instructions`, the `guards`, and the parameter types. Here is what comes out, verbatim:

```
Credit a customer against one order -- a late delivery, a coupon, a shipping
charge applied in error. The credit is added as a negative line and the order
total is recomputed from the lines.

Give the order as its number, the amount in dollars, and a memo saying why.
Look the order up first if you were given a customer rather than a number: an
Order is identified by its key alone. If the call comes back needing review, do
not retry it with a smaller amount: report the reason and wait for a decision.

This call is checked against CreditWithinOrderTotal and
CreditUnderReviewThreshold. A check that fails comes back as a refusal or as a
request for a human decision, and the reason is text you should report as
given.

  order:  string (required)
          Which Order this applies to. Give its key, or text that identifies
          exactly one; the call fails when nothing matches or more than one does.
  amount: number (required)  The amount, as a decimal number.
  memo:   string (required)  The memo, as text.
```

`order: Order` is an ontology type, so it becomes a reference the runtime resolves. `amount: Decimal` becomes a JSON number, so a caller cannot pass "thirty dollars".

`agent_tools.ts` imports no agent framework. It describes tools and leaves the binding to the caller, so LangChain or MCP costs one adapter and the derivation is untouched. The ADK adapter is eight lines.

**Three places the model runs out**, all stated in AGENT.md rather than papered over:

1. A lookup tool can only match exactly — no joins, ranges, ordering or totals. Enough to find the object an action needs. A real question ("which customers are owed the most this quarter") wants a query surface declared in the model; the metrics are already there and wiring them into a tool is the obvious next piece.
2. An entity resolves by its key or by a name-like field. `Order` has neither beyond its key, so the agent has to chain two calls.
3. The persona is unavoidably handwritten. That seems right rather than a gap.

**One bug the exercise found.** `IssueCredit`'s `ai_context.instructions` told callers they could name an order by the customer's name. The runtime cannot resolve that — an `Order` has no name-like field, so it resolves by key alone. Corrected in the model.

`npx tsc --noEmit` clean, 793 pass / 0 fail (17 new). The derivation tests open no database, because what an agent is offered is pure.

libei:
Simplified the constraint evaluator: it substitutes now instead of parsing.

You were right that it should be about composing a SQL string. It always was — the output is one SELECT per constraint — but it got there by parsing the expression into comparisons and rendering SQL back out, which cost 174 lines of tokenizer, operator table and grammar.

**What the four substitutions are.** `Order.total` becomes the column the binding profile mapped it to. `amount` becomes `@p_amount`, bound rather than pasted. `SUM(LineItem.amount)` becomes a subquery correlated on the join columns the model declares. And the invariant is negated, as `NOT COALESCE(x, FALSE)` rather than `NOT x`, so a NULL does not slip through. Everything else in the expression goes to the store verbatim.

**Every constraint that lowered before lowers to the same SQL, byte for byte.** The credit model's three:

```sql
-- amount <= Order.total
SELECT order_id FROM Orders
WHERE NOT COALESCE((@p_amount <= total), FALSE)
  AND CAST(order_id AS STRING) IN UNNEST(@touchedKeys) LIMIT 5

-- amount <= 25
SELECT 1 AS violated FROM UNNEST([1])
WHERE NOT COALESCE((@p_amount <= 25), FALSE) LIMIT 1

-- Order.total == SUM(LineItem.amount)
SELECT order_id FROM Orders
WHERE NOT COALESCE((total = COALESCE((SELECT SUM(amount) FROM LineItem
      WHERE LineItem.order_id = Orders.order_id), 0)), FALSE)
  AND CAST(order_id AS STRING) IN UNNEST(@touchedKeys) LIMIT 5
```

**What it buys beyond the line count.** The grammar cap is gone. `BETWEEN`, `IN`, `CASE`, arithmetic, `LIKE` and parentheses were each refused before and now reach the store, which is the thing that can actually say whether they are acceptable. Two smaller widenings came along: a call that is not an aggregate is left alone, so `LOWER(Order.status)` works, and an expression-bound field is inlined rather than rejected.

**The cost, stated plainly.** A name that is neither a field nor a parameter is no longer caught during lowering — it comes back from the store as an unrecognized name the first time the probe runs. What only the model can know is still refused here: an undeclared field, an entity with no table or no key, an aggregate over an ambiguous relationship. No store can phrase those.

**One thing the substitution has to get right.** Emitted SQL is parked behind a placeholder as it is produced, so a later substitution cannot rewrite what an earlier one wrote. Without it, an action parameter named `total` would rewrite the `total` column the field substitution had just emitted. There is a test for exactly that.

466 code lines to 362. Adds `constraint_eval.test.ts`, 41 tests — the lowering had none of its own before, only the 56 runtime tests that exercise it in passing. `npx tsc --noEmit` clean, 834 pass / 0 fail.

libei:
Superseded. The runtime half of this PR landed in #419 (merged as dd6d034): the action executor, the semantic runtime and store, and the agent tool derivation, now under `semantic/runtime/`.

What was unique to this PR and has not landed is the constraint evaluator and the credit-flow demo. Both are coming as their own PRs rather than as a rebase of this one, so no one has to review a diff whose larger half is already on main.

Closing to keep the queue honest.

PR #416, opened 2026-09-12
feat(mdcode): an action's executor is a binding a profile supplies

libei:
An action's `executor` moves from the logical model to the physical binding layer, beside an entity's `source`.

## Why

Everything else an action declares is logical — what it takes, what gates it, what it changes. The executor is not. It is *how* the change is carried out, and that depends on the store. Where the rows sit in a relational database the write is DML; where they do not, it is a call to whoever owns them — an MCP tool, a REST endpoint, a service method. Even two relational stores differ, because each binds its own table and column names and each speaks its own dialect.

Pinned to the logical model, an action could only ever be performed one way. So the same model bound to a warehouse and to the operational store either wrote DML into a store that does not own the rows, or named a service call the store that *does* own them has no reason to make.

## What changes

- **`executor` is optional on an action.** An action with none is *declared but not performable here*: it still states what it does, what gates it, and what it changes, which is the whole of what a reader needs. The push gate accepts it and the catalog still records it.
- **A profile may carry `actions`.** Each entry sets `name` and `executor`, nothing else. Setting a logical facet (`guards`, `affects`, `parameters`) is rejected like any other declaration override.
- **Omission inherits, unlike a field's column.** An action the profile does not mention keeps the model's executor, so a model can state one default and a profile override only the stores that perform the write differently. `executor: null` withdraws an inherited one, for a binding that performs no writes at all.

  The asymmetry with fields is deliberate. A column inherited into a renamed schema binds to the wrong column and returns another column's data silently; an executor names a whole mechanism, so a wrong one fails loudly at the first call. Omission can safely mean "keep" for one and "unbind" for the other.
- **`pruneUnavailable` gains an actions pass.** An action is dropped when no binding supplies an executor, when a parameter's entity type is unavailable (there is nothing to resolve the argument against), or when a concept in `affects` is.
- **`kcmd profiles` lists them under `cannot run:`**, apart from the read-side `cannot answer:`. An action a binding drops is not a question it cannot answer, it is a write it cannot perform.

## What it looks like

The logical model declares the action once, with a default: ask the service that owns orders.

```yaml
actions:
  - name: CancelOrder
    description: Cancel an order that has not shipped
    parameters:
      - { name: order, type: Order }
    affects:
      - { concept: Order, operation: modify }
    executor:
      mcp:
        server: //agentregistry.googleapis.com/projects/acme-ops/locations/us/mcpServers/orders
        tool: cancel_order
```

The analytical profile says nothing about it, so the warehouse keeps the default — it reports on orders but does not own them. The operational profile does own them, so it writes them:

```yaml
# commerce.profiles/operational.yaml
actions:
  - name: CancelOrder
    executor:
      sql:
        statements:
          - UPDATE Orders SET Status = 'CANCELLED' WHERE OrderId = @order
```

One declaration, one blast radius, two mechanisms.

## Testing

790 pass / 0 fail, `tsc --noEmit` clean. 13 new tests: profile merge (supply, override across kinds, inherit on silence, `null` withdrawal, unknown action, logical-facet rejection), pruning (no executor, unavailable parameter type, unavailable `affects` concept, the bound case, the no-actions case), the loader accepting an action with no executor, and the push gate accepting one.

The worked example in `profiles.md` was extracted and run through the real `mergeProfile` and `pruneUnavailable` passes rather than asserted: analytical resolves to the `mcp` executor, operational to `sql`, with no warnings and nothing dropped.

## Docs

`profiles.md` (contract table, availability rule, merge rules, validation, and the worked example), `actions.md` (a new *The executor is a binding* section), `model_spec.md` §5 and §7.3, `reference.md`, and the semantic-model README.


libei:
Follow-up commit `ae0d8c9` addresses a self-review pass over the first commit — three consumers still assumed every action has an executor:

- the `semantic-action` aspect type declared `executorKind` **required**, so publishing an executor-less action wrote an aspect violating its own type. No longer required.
- `readAction` skipped any entry whose executor it could not read, so such an action could not be pulled back — guards and blast radius went with it. No kind now recovers as an action with no executor; a kind with missing coordinates is still treated as damage.
- `executor: null` in the *model* document was a parse error, while the same two characters in a *profile* are the documented way to withdraw one. `.nullish()` makes them agree.

Two doc claims were also narrowed. `actions.md` said an executor-less action "still publishes to the catalog" without qualification — true for a catalog-only push, but a push that also deploys a graph prunes it exactly as it prunes a metric the binding cannot answer, which removes an entry an earlier push published. And `reference.md`'s justification for standing the concept check down holds for concepts but not for `affects` *fields*; that gap is now named rather than implied away.

793 pass / 0 fail.

PR #417, opened 2026-09-12
Add a `proposal` executor kind for writes composed at call time

libei:
Adds a fifth executor kind, `proposal`, for a write that does not exist when the model is authored.

## Why

The four existing kinds all answer *what is the write?* — `sql` by carrying it, `mcp`/`rest`/`grpc` by naming a system that holds it. `sql` push rules say so explicitly: "no statement composed at call time. An action whose body arrives with the call declares nothing, and a gate cannot check what was never declared."

That rules out a real class of write — the ad hoc correction, the cleanup whose shape depends on what is found, the fix an agent works out from the data in front of it. Nobody can pre-write those statements, and running one the caller supplies would mean running a write no one declared.

A review engine is the existing answer: the caller proposes a statement, a reviewer checks it against a policy, and only then does it run. This puts that arrangement inside the model rather than beside it.

## What

```yaml
      - name: AdjustPricing
        description: Change the price on an order, subject to review
        executor:
          proposal:
            reviewer: //dataagents.googleapis.com/projects/acme/locations/us/actionManagers/commerce
        parameters:
          - { name: order, type: Order }
        guards:
          - discount_within_policy
        affects:
          - { concept: Order, operation: modify, fields: [price] }
```

One coordinate, and it is not a write. The kind answers *who decides* instead. No statement by construction — the schema rejects one.

**`affects` is required for this kind, and for no other.** Everywhere else `affects` and `guards` describe a write that is already pinned down; here they are the only description that exists before the call, so they are what the reviewer checks the proposal against. A `sql` executor's statements say what they touch and a remote call is performed by a system that knows what it is about to do; a proposal has neither, so an undeclared blast radius leaves the reviewer nothing to compare against. Push rejects it.

**`guards` name the rules the proposal is checked against**, in the model, next to the ontology they are written over — rather than restated in the review engine's own policy language.

**What this does not promise** is that the proposed write matches the declaration. Nothing can — the statement arrives after the model is published. Being the thing the proposal is checked *against* is the point of having a declaration.

## Profiles

The kind rides the executor binding from #416. A profile can bind `proposal` where the store holds production data and leave staging writing DML directly: whether a write is reviewed is a property of the environment, like everything else in a profile.

## Round trip

- `proposalReviewer` joins the `semantic-action` aspect type at index 13 and reads back to the tagged union.
- A blank reviewer is a hard push error, like every other coordinate.
- The OSI converter emits the kind on the way back out.

`kcmd` neither reviews nor runs a proposal; it publishes the arrangement: who decides, what the call may touch, which rules apply.

## Docs

`actions.md` gains "When the write is composed at call time"; `model_spec.md` §5, `reference.md`, `profiles.md` and the README are updated. The actions.md example was run through the real loader and push gate rather than asserted — it normalizes as written, and removing `affects` produces exactly the documented rejection.

## Tests

`npx tsc --noEmit` clean. `npx bun test`: **799 pass / 0 fail** across 34 files (793 on main). Ten tests added or updated: loader normalization, statements-in-proposal rejected at parse, the no-`affects` hard error, with-`affects` passing, blank reviewer, the rule applying only to this kind, and the Knowledge Catalog round trip including the foreign-field matrix.

### DIFF
--- a/toolbox/mdcode/docs/semantic-model/README.md
+++ b/toolbox/mdcode/docs/semantic-model/README.md
@@ -124,7 +124,8 @@ push enforces.
 A model can also declare **actions**: model-level write operations, the
 write-side counterpart to a metric. An action names a business operation, points
 at the executor that carries it out (an MCP tool, a REST endpoint, a gRPC
-method, or DML), and types each parameter against the ontology, so an
+method, DML, or a reviewer for a write composed at call time), and types each
+parameter against the ontology, so an
 entity-typed parameter is an object reference rather than a bare string. The
 executor is a physical binding, so a [binding profile](profiles.md) can supply a
 different one per store. An action also
--- a/toolbox/mdcode/docs/semantic-model/actions.md
+++ b/toolbox/mdcode/docs/semantic-model/actions.md
@@ -81,8 +81,11 @@ a tool already registered in Agent Registry by the server's resource name plus
 the tool's name within it. `rest` (`{endpoint, method}`) and `grpc`
 (`{service, method}`) are the other two remote kinds. A fourth, `sql`, carries
 the write itself rather than a pointer to whoever performs it — see
-[Writing the statements in the model](#writing-the-statements-in-the-model).
-Exactly one kind, where an executor is written at all.
+[Writing the statements in the model](#writing-the-statements-in-the-model). A
+fifth, `proposal`, names neither, because its write does not exist yet — see
+[When the write is composed at call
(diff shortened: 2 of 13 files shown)


libei:
Closing this. The reviewer here is really just constraint evaluation, and `guards` already says which constraints apply — pinning a reviewer resource name in the model binds the logical layer to one service when the same check can be a library call from an agent or from the action runtime. Nothing in the stack composes writes at call time yet, so this adds a kind ahead of its consumer. Revisiting if that changes.

PR #418, opened 2026-09-12
feat(mdcode): derive an agent's tools from the model, and run them live

libei:
An agent needs two things from a semantic model: a list of what it may do, and
a way to do it. #413 gave the second — a runtime that resolves, binds and
applies an action in one transaction. This gives the first, and then runs the
whole path against a live Spanner database to show the two halves meet.

Seven commits: the library layer, the demo, the fixes a self-review turned up
(see the comment below), three passes that move everything out of the demo that
was not the demo's, and one that points the demo at the use case it was written
from. Worth reading in order.

## 1. Derive an agent's tools from the model

`modelTools({model, client})` returns `{lookups, actions, instruction}`.
Nothing about it is specific to a business or a framework:

- one **lookup tool per entity** — exact match on any bound field, ANDed, 50-row
  cap, bound parameters, no joins, ranges, ordering or aggregation. Narrow on
  purpose: enough to find the object an action needs, and small enough that the
  generated SQL is checkable by eye. Each filter carries what the model says the
  field holds, ahead of the sentence about matching — for a coded field that
  description is the only written-down list of the values it accepts.
- one **action tool per action** — name, description, calling guidance and typed
  parameters all read out of what the model declares.
- one **instruction** — the model's own `ai_context.instructions` followed by
  what the derivation owes a caller about using the tools.

A tool says whether it can be called. Both halves carry `runnable` and, when it
is false, `unavailable`. An action's verdict comes from
`whyRefusedWithoutRunning` in runtime.ts — the same function `runAction`
consults — so the two cannot drift into disagreeing about whether a write would
go through; a lookup's comes from the same function that reports the problem at
call time. An unavailable tool is still derived, still named and still
described: an action the model declares should not vanish from what the model
offers just because today's runtime cannot run it. Dropping it silently is the
alternative, and it leaves an operator unable to tell a policy decision from a
broken binding -- which is why `callableTools` hands back `withheld` rather than
filtering it away.

Nothing under `src/` imports an agent framework, and nothing should. The caller
writes the binding. `bindScalar` is exported alongside, so a caller that wants to
prepare parameters itself can.

52 tests in `agent_tools.test.ts`.

## 2. A demo that runs it against a live store

`demo/agent/` is a real `kcmd` workspace, not a script with a model pasted into
it. `commerce.yaml` names three entities, one action and three rules, and names
no table, no column and no SQL. `commerce.profiles/spanner.yaml` supplies all of
that, including the action's executor — so the one file specific to this
deployment is the one that says where the deployment is. `kcmd action list`,
`kcmd agent tools`, `kcmd action run` and the agent all read those same two
files.

Every transcript in `demo/agent/README.md` was captured from a real run against
`sqlgen-testing` / `graph-unified-solution-demo`.

### The constraints are declared and none is referenced

That is the honest state, and the demo makes it loud rather than hiding it. A
constraint is inert until an action names it in `guards`, and this runtime does
not evaluate constraints yet — so naming one would make the action *unrunnable*
rather than checked. The loader says so in two warnings that the README quotes
instead of suppressing, and the README shows the refusal in full by asking the
reader to attach a guard and re-run `kcmd agent tools`.

Refusing is the behavior worth demonstrating: a write the model says must be
checked is not run because the checker is missing, and it is not run unchecked
either. The constraint evaluator is the next piece of work.

That leaves the ladder with two rungs, run and refuse, where it wants three. The
third is approval, and the model already names it: a constraint carries
`on_violation: escalate`, which `ir.ts` defines as holding the write for a person
to decide. Nothing acts on that yet. Most action frameworks declare confirmation
as a blanket flag on the action; a rule that escalates only when a predicate is
violated says more and costs more to honor, because something has to evaluate the
predicate -- the same evaluator. Saying so plainly, because a $30 credit that
writes unconditionally otherwise reads as an oversight rather than a bounded gap.

## 3. Fixes from a self-review

Eight findings, all real, all fixed. See the comment below.

## 4. The demo keeps the framework glue and nothing else

Three of the demo's five TypeScript files existed because the library kept
something private that a caller outside the CLI needed. Each was a worse copy of
a command that already worked.

| Was | Is now |
| --- | --- |
| `model.ts` (86 lines) | `openWorkspace()` in `semantic/workspace.ts` — the CLI's own loader, lifted down and called by both |
| `target.ts` (22 lines) | `kcmd action list --store` |
| `tools.ts` (132 lines) | `kcmd agent tools` |
| the persona in `agent.ts` | model-level `ai_context` in `commerce.yaml` |
| `setup.sh` (74 lines), `cleanup.sh` (18) | four `gcloud` commands and one, listed in the README |

`model.ts` got the load sequence right and still lost the check the CLI makes
afterwards — that every entity is bound to the *same* database the deployment
target names — so a demo whose profile bound one database and targeted another
would have written to the target's table of that name and reported success.
That class of copy is the reason to lift rather than to share by example.
`kcmd action list` now prints the store in its header too: a listing of what you
can run that does not say where it runs is half an answer.

The persona moving into the model is the part worth arguing about. A prompt that
says what this business wants of anything acting on it is not the agent's to
hold: it is true of every agent over this model, including the ones nobody has
written, and an agent that keeps it privately is a place the rule can be changed
without the people who own the model finding out. `ModelTools.instruction` is
that, plus the tool-contract half the derivation owes — which is not the agent's
either, because it describes a contract `agent_tools.ts` defines and no model
ever stated.

The two shell scripts went the same way in the last commit. `cleanup.sh` was one
`gcloud spanner databases delete` behind twelve lines of preamble; `setup.sh` was
a create plus three inserts behind a re-runnability story nobody asked for —
dropping the database and running the four commands again does the same thing.
Both are in the README now, each command copyable on its own. The one thing the
scripts were really protecting survives as the line that asks the model where to
write:

```bash
IFS=/ read -r PROJECT INSTANCE DATABASE <<<"$(../../dist/kcmd action list --store)"
```

A second source of truth for that is how you seed one database and have the agent
talk to another — and in a script, that line was invisible.

What is left in `demo/agent` is `agent.ts` (57 lines of code, none of which
mention credits, orders, customers, Spanner tables or SQL), the two model files,
`catalog.yaml`, `schema.sql` and the README. The schema stays a file because
`gcloud spanner databases create` wants `--ddl-file` and because `kcmd push`
deploys a graph over tables that already exist rather than creating them. The
README is now a recipe — write the model, create the store, check it with `kcmd
action list` and `kcmd agent tools`, then write the agent — rather than a
finished artifact.

### ADK is the demo's dependency, not the library's

`demo/agent` carries its own `package.json` and its own `node_modules`. It also
carries a `check` script, because `demo/` sits outside the package tsconfig and
code nobody typechecks is code that quietly rots.

## 5. The agent keeps only what is ADK's shape

Two things in `agent.ts` were not the framework's shape — they were work the
demo did because nobody else would, and 69 lines of code was still too many.

`schemaFor` built a Zod object one field at a time: twelve lines of ladder plus
a paragraph on why Integer has to stay distinct from Number. ADK never needed
it. `FunctionTool.parameters` accepts a genai `Schema` as well as a Zod one, and
a derived parameter already carries a JSON type and a description — which is the
whole of a function declaration. The bridge is six lines of plain data now, the
integer distinction survives as `Type.INTEGER`, and the demo carries no schema
library.

Filtering out the tools this binding cannot serve was the other. Every adapter
has to make that split and it is the same split every time, so it became
`callableTools` in `agent_tools.ts`, returning `{callable, withheld,
instruction}` across both halves at once. It also makes dropping `withheld` in
silence a choice rather than an accident.

`agent.ts` is **57 lines of code**. What is left is the four steps, the
transcript printer, the usage line and the date.

## 6. The demo runs the use case it was written from

The scenario always came from "Ontologies, actions and policies -- a simple
illustration", but the data and the request were invented, so nothing checked
whether the model could answer the case the doc actually poses. It now runs
that case: Andy Brook's order, placed on Labor Day, carrying the $30 shipping
charge that should not have been billed, and the doc's request verbatim.

Doing that found one library bug and drew one boundary.

**The bug is general.** A lookup filter threw away the field's own description.
`LineItem.type` is documented in the model as "item, tax, fee, or credit" and
the derivation replaced that with boilerplate, so the agent guessed
`type: "shipping"`, got nothing back and spent a turn learning it was wrong.
The description now leads and the boilerplate follows, and `kcmd agent tools`
prints one line per filter instead of a bare list of names — the point of that
command is to show what the agent is handed, and a name list hid the half the
model wrote. Every filter over a coded field had the same hole, in every model.

**The boundary goes the other way.** "Labor Day" resolves to `2026-09-07` only
with a clock, which a language model does not have; without one the agent
guessed a year, dropped the filter and scanned, or stopped to ask. That is a
fact about when the process is running, not about commerce, so `agent.ts`
appends today's date to the instruction and `commerce.yaml` stays silent about
it. Each new use case pushes on this boundary in one of those two directions;
the useful question is always which.

The README names its source and lists what is use-case-specific: the two catalog
files, `schema.sql`, and the seed rows in step 2. Nothing in `agent.ts` and
nothing under `src/` — which is the claim the demo is making, now stated as a
file list a reader can check.

It also records two places the demo falls short of the doc, rather than quietly
matching the parts it can. The $30 credit trips the doc's review threshold and
is written anyway, because a constraint is inert until an action guards it and
this runtime cannot evaluate one — so the demo reaches either end of the doc's
policy story, write or refuse, and not the middle. And the doc describes two
agents over this model, an internal one and a customer-facing one scoped to its
caller; only the internal one can be built, because the lookups take no caller
identity.

## Docs

`docs/semantic-model/actions.md` gains **8. Hand it to an agent**: what the
derivation produces, why the instruction is not the agent's to write, what the
read tools deliberately cannot do, how a tool reports that it cannot be called,
and why the framework binding is the caller's.

## A named follow-up

`affects` already declares an action's blast radius with a verb (`create`,
`modify`, `delete`), which is most of what MCP's `ToolAnnotations` wants: every
derived lookup is `readOnlyHint: true`, and an action that deletes is
`destructiveHint: true`. Deriving those from a declared blast radius rather than
from a wire verb is worth doing and appears to be unoccupied ground.

It is not in this PR because `agent_tools.ts` takes no framework dependency and
`ToolAnnotations` is an MCP name. It belongs with the server that would emit it,
not ahead of it.

## Verification

- `npx tsc --noEmit` clean
- `npm test` — 948 pass, 0 fail
- `npm run check` in `demo/agent` clean
- `demo/agent/README.md` followed step by step against `sqlgen-testing` /
  `graph-unified-solution-demo`, most recently in full on 2026-09-13.
  `kcmd action list --store` resolved the database out of the model (the load
  warnings go to stderr, so the README's `IFS=/ read` line works as written).
  The step-2 inserts left orders at 165.85 / 18.00 / 200.00. `kcmd action list`
  and `kcmd agent tools` printed what the README quotes, including the
  `type: string  -- item, tax, fee, or credit.` filter line. `kcmd action run
  IssueCredit --arg order=12346 --arg amount=3.00` took that order to 15.00.
- The doc's task — "Find the order for Andy Brook (andybrook@gmail.com) that was
  placed on Labor Day..." — called `find_customer`, `find_order({placedOn:
  "2026-09-07"})`, `find_line_item({type: "fee"})` and `issue_credit({amount:
  30})`, leaving order 12345 at 135.85 with a -30 credit line against the 30.00
  fee. Earlier runs repeated that sequence four times. "Labor Day" resolves on
  the first try and the `fee` filter is never guessed, so the model's field
  descriptions are reaching Gemini through the plain schema.
- With `guards: [CreditUnderReviewThreshold]` attached, the same task read the
  charge and reported it could not write it. Order 12345 stayed at 165.85 and no
  credit line was created. Attaching the guard also takes the loader from two
  warnings to one, so the warning is computed rather than fixed text.
- `schema.sql` matches the deployed DDL exactly (`gcloud spanner databases ddl
  describe`) across all three tables, their columns, types and primary keys.
  The database was reset by clearing and re-seeding its rows rather than by
  being dropped and recreated, so `databases create --ddl-file` and the
  Cleaning-up `databases delete` are the two README commands not exercised in
  this run.
- Every total quoted above was read back with `gcloud spanner databases
  execute-sql`, and every transcript in the README is one of these runs.


libei:
## Self-review

I ran a code review over this PR before asking anyone else to. It found eight
things; all eight were real, and all eight are fixed in `c05005a`. Recording
them here so a reviewer can see what was already caught and spend their
attention elsewhere.

Six of the eight are one mistake wearing different clothes: something the model
alone decides was being decided later, or said somewhere it was not true.

| | Finding | Fix |
| --- | --- | --- |
| 1 | `runnable` was true for two actions the runtime always refuses: an object reference to a composite-keyed entity, and a generated key a UUID cannot fill for an Integer-keyed entity. Both are decidable from the model; both were reached inside the transaction. | Moved into `whyRefusedWithoutRunning`, the one place that answers this. Not asked of a handler, which writes its own DML. |
| 2 | The demo could seed one database and talk to another. `DEMO_SPANNER_DATABASE` and its siblings steered the shell scripts only; the store comes from the profile's `deployment_target`. A reader exporting their own project would seed it, then watch the agent connect to `sqlgen-testing`. | One source: `target.ts` asks the profile. Scripts, tools and the Gemini client all follow the single line in `commerce.profiles/spanner.yaml`. |
| 3 | The demo skipped `resolveInheritance`, which the CLI runs and whose absence its own comment calls *unsafe* for the runtime — an inherited key's type is invisible, so the UUID-vs-Integer check reads it as a String. Latent today; `DEMO_MODEL_PATH` invites a model where it is not. | Resolved, with `bindingOptional`, as the CLI does. |
| 4 | `EntityTool` carried no availability verdict, so an abstract entity, an unbound entity, and an entity bound to something other than a plain table were all offered as callable and failed every call. | `runnable`/`unavailable` on the same terms `ActionTool` uses, from the same function that reports the problem at call time. `agent.ts` filters both halves through one predicate. |
| 5 | A tool announced "This call is gated by X" for a guard whose `on_violation` is `warn`. The runtime stands down for an advisory rule and the write goes through, so this told a language model to expect a refusal that cannot happen. | An advisory guard is no longer named as a gate. |
| 6 | A comment block had drifted one function away from what it describes. | Moved back. |
| 7 | Stale `demo/credit/agent.ts` path in the module header. | `demo/agent/agent.ts`. |
| 8 | `integer` reached Zod as a bare `z.number()`, so a caller could send `2.5` for an Integer parameter and lose a turn to a rejection the schema could have prevented. | `z.number().int()`. |

Ten tests added, and the live runs repeated end to end after the changes:
`tools.ts` took order 12345 from 147.85 to 135.35, `kcmd action run` took 12346
to 15.00, and the agent — "Dana Reyes says the stand mixer arrived scratched,
give her $20 off" — called `find_customer`, `find_order`, `issue_credit` and
left 12347 at 180.00. Each read back with independent SQL.

934 tests pass, 0 fail. `npx tsc --noEmit` and `demo/agent`'s `check` both clean.

libei:
Splitting this into the generic building blocks and the demo that exercises
them, so each is reviewable on its own terms rather than through the other.

The library half is #419: `agent_tools.ts`, `workspace.ts`, the runtime changes,
both test files and the docs section — 2,275 lines, unchanged from what was here
except that two references to `demo/agent` travel with the demo rather than
ahead of it. It is three commits now, each typechecking and passing the suite on
its own.

The demo half — `demo/agent/`, 902 lines — follows once #419 lands. Nothing is
lost: the branch behind this PR still has every commit and every file, and the
review comment below still applies to the code in #419.

Closing in favor of #419.

PR #436, opened 2026-09-16
docs(semantic-model): restructure section 8 around its argument

libei:
Follow-up to #433, which revised section 8's prose. This one changes its
structure and nothing else about the argument.

## What was wrong

Section 8 ran 356 lines: a **152-line unheaded preamble (43% of the section)**
followed by five H3s. Three consequences:

- A reader met `[NOT RUNNABLE]` in the listing at line 1301, learned who adds
  it at 1412, and learned why at 1441 — a **215-line gap**, for the marker
  that carries the section's point about binding.
- Reference material was 22% of the section, spread across five non-adjacent
  places.
- `Calling it from code` ran 74 lines over three distinct topics.

The underlying cause is that the section does two jobs in one stream: it makes
an argument (*you don't write these tools; they're derived*) and it is the
reference for the derivation. Those want opposite orders.

## What changed

Six headings, each with one job; the preamble drops to 15 lines because the
derivation and its three products are named up front instead of 143 lines in.

| heading | job |
|---|---|
| `## 8. Hand it to an agent` + 15-line opening | the claim |
| `### The set an agent is handed` | command, listing, and the `[NOT RUNNABLE]` gloss in place |
| `### Where each line comes from` | Table 4, and who owns each half of the instruction |
| `### What a write tool and a lookup tool do` | unchanged |
| `### What a withheld tool is waiting on` | the reasons, with the `--judge` contrast beside the reason it demonstrates |
| `### Calling it from code` | unchanged |
| `### The commerce demo, worked through` | unchanged |

`### A tool says whether it can be called` is gone; its opening paragraph moved
up under the listing and the rest became `What a withheld tool is waiting on`.
`### Who owns the instruction` is folded into the provenance heading, where
Table 4 has just shown `ai_context.instructions` becoming the instruction.

Six exceeds the guide's usual five H3s per section. Five was reachable only by
merging tool *provenance* with tool *capability*, which would recreate the
74-line two-topic heading this is meant to fix.

## One claim corrected

`kcmd agent tools` did not "read your model and nothing else". It resolves the
store your profile binds (`commands.ts:1352`, `:1379`), and a model whose
profile binds none prints `offers no tools under this profile.` and exits
non-zero (`:1387`). The same sentence also said it opens no session and changes
nothing, which **is** true — `runtime.ts:176` resolves a store without
connecting, and the listing calls no model.

## Verification

- All five fenced blocks carried across **byte-identical** (multiset diff
  against `origin/main`: 0 lines removed, 0 added).
- Prose at 80 columns; every internal link resolves; the frozen anchors
  (`#8-hand-it-to-an-agent` and the four inbound-linked ones) intact. No
  section-8 H3 anchor is linked from anywhere in the repo, checked with
  `git grep -ho 'actions\.md#[a-z0-9-]*'`.
- `npx bun test`: 1083 pass, 0 fail.

Net: 356 → 354 lines, 69 insertions / 71 deletions.


libei:
#435 landing moved main to `985c75b` and left this branch conflicting. The rebase is done and verified, but it is sitting on a **new** branch rather than on this PR's head, because publishing it here means rewriting an already-pushed branch:

    fork/section8-structure-rebased   f79e325   rebased onto 985c75b, verified
    fork/section8-structure           c1643d0   this PR's head, untouched

To land it, either force-push `f79e325` over `section8-structure`, or open a fresh PR from `section8-structure-rebased` and close this one.

**What the rebase had to resolve.** Two conflicts, both in `actions.md`, plus one anchor git could not have fixed on its own.

Section 7 renamed three headings. Git remapped six of the seven references automatically, but not the one this PR had relocated within section 8 — git only maps a rename onto a line that stayed put, so that site came through with a stale target and no conflict marker. Fixed explicitly:

    #why-a-guarded-action-is-refused  ->  #when-a-rule-stops-the-call

The new target is a closer fit than the old one: `### When a rule stops the call` is the subsection that carries the `TransferFunds` / `AmountIsPositive` expression refusal this sentence cites, and it ends on "a refusal is settled before a session opens, so a refused action leaves no transaction behind" — which is the claim the section 8 sentence makes.

**Verification after the rebase**

- Old anchor targets repo-wide: none. Stale link *label* text: none.
- `xlinks.py` over the tree, node_modules excluded: 0 broken.
- Fenced blocks vs `origin/main`: 0 lines removed, 0 added.
- `bun test`: 1083 pass, 0 fail.

**Also in this push: a writing pass over the restructure** (`f79e325`), six prose defects:

- A garden path in subject position — "A tool this binding can't serve is…" reads "A tool this binding" as one noun phrase.
- `withheld` first appeared in a heading, 100 lines after the concept it names. The section was calling one thing by three names (`[NOT RUNNABLE]`, "tools this binding can't serve", withheld); they are bound together at first contact now.
- "What a withheld tool is waiting on" opened on two library field names before the reader has a reason to care. Those moved to "Calling it from code", which owns the library surface.
- "derived from it" — the nearest noun is "the change", not the model.
- A third "It" in a chain whose previous link is the store, which can itself open a connection.
- "offerable" where the document already says callable.

libei:
Superseded by #439, merged as `7a02244`. Same work, rebased onto `985c75b` after #435 landed, plus a writing pass and the section 7 anchor fix. This branch's head was pre-rebase and could only have been updated by rewriting an already-pushed branch, so the rebase went out as a fresh PR instead.

PR #440, opened 2026-09-16
docs(mdcode): propose an 'intent' executor for generated writes

libei:
Proposal only — no code. Adds `toolbox/mdcode/docs/semantic-model/generated-writes.md`.

The read side already generates SQL from words: a judge holding `--judge-reads-store` is handed the schema the profile binds and writes its own `SELECT`. The write side has no equivalent — an action either carries literal DML in a `sql` executor, or names a foreign system and `runAction` refuses it without a hand-written TypeScript `handler`. That handler is the part nothing in the catalog can see and no rule in the model can reach.

The proposal:

- **One executor kind, `intent`.** Its body is the request in words; `produces` names which existing kind the generation fills in — the `statements` for `sql`, the request body for `rest`, the arguments for `mcp`. The destination (endpoint, tool, table set) is declared and never generated.
- **A `Planner` seam**, shaped exactly like `Judge`: an interface in the library, an implementation in `gcp/gemini.ts`. No planner means the action is refused, the same way a judged guard with no judge is refused today.
- **An admission gate** (`runtime/admit.ts`) over everything generated, run before a session opens: the shape check hoisted out of `validate.ts` so publish time and call time share one implementation, no literal in a value position, tables bounded by `affects`, every declared parameter used. Checks 2 and 3 read the AST the bundled `@polyglot-sql/sdk` engine already produces for DML.
- **Guards move.** A judged guard sees the generated statements, not only the arguments — judging the arguments judges the wrong thing once the arguments no longer determine the write. Reference resolution moves outside the transaction so generation and judgment happen before it opens.
- **`kcmd action compile`** graduates an intent into a reviewed literal `sql` executor in the profile, so words are an authoring affordance rather than a runtime dependency.

Named as a prerequisite: `fork/constraint-eval` (79709b0). Its `after`-timing invariants are the enforcement point for a write nobody wrote — the admission gate bounds which tables and which values, and nothing bounds which rows.


libei:
Superseded. The per-action `intent` executor still required an author to anticipate each write; it only moved the hard-coding from a TypeScript handler into a YAML string. Replaced by a freeform proposal that enumerates the bounds instead of the writes.

PR #441, opened 2026-09-16
docs(mdcode): propose freeform actions

libei:
Proposal only — no code. Adds `toolbox/mdcode/docs/semantic-model/freeform-actions.md`. Supersedes and replaces #440, which is closed; nothing is carried over from it.

An agent in conversation meets write requests no action covers, and declaring more actions never closes that gap — whatever is enumerated, the next sentence is outside it. So enumerate the **bounds** instead of the writes and generate the statement inside them at call time.

- **A `writable` surface on the binding profile.** Default closed. Written in the same `concept` / `operation` / `fields` vocabulary as `AffectedConcept`, so admission is a subset test between two values of one type rather than a comparison between a declaration and some SQL. Writability is a physical facet, so it inherits and withdraws exactly as `executor` and `source` do.

- **Two-phase, always.** `propose_change(request, refs)` writes nothing and returns what it would do; a person approves; `apply_change` runs *the statements the proposal holds*, not a regeneration. A declared action's DML was read by somebody before it shipped — a freeform statement has no such moment, so the moment is the call and the reviewer is the one asking.

- **A six-check AST gate** (`runtime/admit.ts`), using the `@polyglot-sql/sdk` engine already bundled for `transpile.ts`. The load-bearing check is that every `UPDATE` and `DELETE` must be keyed to a row a lookup already resolved. "Delete all orders" is well-formed DML inside any reasonable surface; what stops it is that no lookup resolved those rows, so there are no keys to bind. That rule is the difference between a bounded surface and a database connection.

- **Constraints attach by what the statement touches**, not by a `guards` list — `AnalyzedRule` on `fork/constraint-eval` already reports the entities and fields a rule reads. This reopens the settled rule that a constraint is inert until referenced. The argument for the carve-out is in the doc: that rule protects *declared* calls from a rule published later, and a freeform write is neither declared nor previously working, so nothing can regress. Declared actions' guards are untouched. Worth settling before anything is built.

- **Graduation.** `--save-as` turns a freeform write that worked into an ordinary `sql` executor action for review. Freeform is the long tail and the intake path, not a replacement — the catalog grows from what people asked for, and anything recurring stops being generated.

`fork/constraint-eval` (79709b0) is a prerequisite rather than a neighbour: admission bounds which tables, columns, values and rows, and nothing in it says the resulting state is sound. Its `after`-timing invariants, which roll back inside the transaction, are the only thing that bounds a statement nobody wrote.


libei:
Closing unmerged. The freeform-actions proposal is not being pursued; reopen or re-propose if it comes back.

PR #442, opened 2026-09-17
feat(mdcode): add parameter descriptions, defaults, and optionality to actions

libei:
## Summary

Implements action parameter extensions and documentation updates motivated by the MCP Toolbox comparison analysis:

1. **Action Parameter Schema & Round-Tripping**:
   - Added `description?: string`, `required?: boolean`, and `default?: unknown` to `ActionParameter` across IR (`ir.ts`), YAML/JSON loader (`loader.ts`), OSI converter (`osi_converter.ts`), and Knowledge Catalog `semantic-action` aspect templates (`kc_custom_types.ts`, `kc_actions.ts`).
2. **Validation for Shared Parameter Types**:
   - Added validation in `validateActions()` requiring every parameter to provide a `description` whenever an action declares multiple parameters of the same `type` (e.g., `source: Account` and `target: Account`).
3. **Agent Tool Derivation**:
   - Updated `toolParameter()` in `agent_tools.ts` to surface authored parameter descriptions (appending entity-resolution guidance for entity references), propagate `default`, and compute `required = param.required ?? (param.default === undefined)`.
   - Updated `toolDescription()` to surface gating constraint rule text (`description`, `judgment`, or `expression`) rather than only constraint names.
4. **Runtime Execution**:
   - Updated `runAction()` in `run_action.ts` to substitute parameter defaults when omitted and bind `null` with the appropriate SQL/Spanner parameter type when `required: false` parameters are omitted.
5. **Documentation (`docs/semantic-model/actions.md`)**:
   - Reconciled `TransferFunds` SQL executor statements in Section 1 with its `affects` declaration by including `INSERT INTO transfer ...`.
   - Added distinguishing parameter descriptions to `source` and `target` throughout the guide and documented `description`, `default`, and `required: false`.
   - Updated Section 4 validation checks (4 -> 5 checks) and Section 8 `kcmd agent tools` output/diagram.

## Verification
- Added unit tests in `actions.test.ts`, `agent_tools.test.ts`, and `run_action.test.ts`.
- All 1,011 tests pass (`npm test`).


libei (commented):
Reviewed by running the branch rather than reading it. The full suite passes (968 + 43 = 1,011, matching the description) and typecheck is clean, so everything below is uncovered by the tests. Each finding was reproduced against this branch with a throwaway harness.

The shape of the change is right, and the plumbing is consistent across loader, OSI converter and the aspect type. The `affects` reconciliation in section 3 is a real fix — `Transfer` and `TransferDebits` are both declared in section 1, so the added `INSERT` and the two `create` entries line up. `required: false`, `default: 0`, `default: false` and `default: "007"` all round-trip correctly.

Ranked, the inline comments are:

1. **Blocking** — the new description rule runs in `validateRunnable`, so it breaks `kcmd action run` on models that ran before (`validate.ts`).
2. Three Knowledge Catalog round-trip data losses: `default: ""` flips optional back to required, `default: null` becomes the string `"null"`, and numeric defaults lose exactness (`kc_actions.ts`).
3. Describing a scalar parameter deletes its format guidance — a described `Date` loses `YYYY-MM-DD` (`agent_tools.ts`).
4. Nothing type-checks a `default` at push time (`validate.ts`).
5. Consistency and reuse items across `agent_tools.ts` and `run_action.ts`.
6. Two listings in the guide that cannot be reproduced (sections 4 and 8).

One finding has no line in this diff to hang off:

**`kcmd action list` and `kcmd agent tools` now disagree about the same model.** `printActionTool` renders `?` for an optional parameter (`commands.ts:1434`), but `runLine` still emits `--arg` for every parameter and `describeParameter` shows no optionality marker. An action with `{name: memo, required: false}` and `{name: currency, default: 'USD'}` prints

```
run: kcmd action run X --arg memo=<String> --arg currency=<String>
```

so a reader copies a command line that passes values for the very arguments this PR made omissible — and `--arg currency=` is not the same call as omitting it.

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -224,11 +224,21 @@ function validateActions(
...
+    const typeCounts = new Map<string, number>();
+    for (const param of action.parameters) {
+      typeCounts.set(param.type, (typeCounts.get(param.type) ?? 0) + 1);
+    }
     for (const param of action.parameters) {
       if (param.isEntityRef === undefined) {
         errors.push(`${where} has parameter '${param.name}' whose type '${
             param.type}' is neither a known entity nor a scalar datatype.`);
       }
+      if ((typeCounts.get(param.type) ?? 0) > 1 && !param.description?.trim()) {
```
libei: `validateActions` is also called by `validateRunnable`, and this check isn't gated, so a documentation rule now blocks `kcmd action run`.

Verified against this branch: a model with `[source: Account, target: Account, amount: Float]` and no descriptions — the shape this guide shipped until this PR, and the shape already published to KC — returns two errors from `validateRunnable()`, and the run aborts before the store is touched.

`validateRunnable`'s own header says it carries only checks where "the runtime acts on exactly what they verify". A description bears on nothing the runtime does. This belongs in `validatePushRequirements` only.

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -224,11 +224,21 @@ function validateActions(
...
+        errors.push(
+            `${where} has multiple parameters of type '${param.type}', so ` +
+            `parameter '${param.name}' must have a 'description' to ` +
+            `distinguish it.`);
+      }
```
libei, line 269: While you're adding a push-time gate here: nothing type-checks a `default` against the parameter's declared type.

Verified: `{name: amount, type: Float, default: 'banana'}` returns `[]` from both `validatePushRequirements` and `validateRunnable`. It deploys clean, the derived tool advertises the default, and the first call that omits the argument dies with `'amount' is a Float, but 'banana' is not.`

`bindScalar` is exported and pure — `bindScalar(param, param.default)` here catches the whole class at push time.

toolbox/mdcode/src/libts/semantic/kc_actions.ts
```diff
@@ -175,7 +175,14 @@ function actionAspectData(action: Action): Record<string, any> {
...
     ...(action.executor ? executorData(action.executor) : {}),
     parameters: action.parameters.map(
-        p => compact({name: p.name, type: p.type, isEntityRef: p.isEntityRef})),
+        p => compact({
+          name: p.name,
+          type: p.type,
+          isEntityRef: p.isEntityRef,
+          description: p.description,
+          required: p.required,
+          default: p.default !== undefined ? String(p.default) : undefined,
```
libei: `String(p.default)` corrupts a null default.

Verified round trip: `{name: memo, type: String, default: null}` emits `default: "null"` into the aspect, and `parseDefaultFromAspect` returns the string unchanged for a String-typed parameter, so the pulled model carries the four characters `null`. `runAction` then substitutes them and the store is written the text `null` where the author asked for SQL NULL.

Same class of loss for an object or array default, which becomes `"[object Object]"`.

toolbox/mdcode/src/libts/semantic/kc_actions.ts
```diff
@@ -345,6 +352,15 @@ function readParameter(
   const type = typeof p?.type === 'string' ? p.type : '';
   if (!name) return undefined;
   const param: ActionParameter = {name, type};
+  if (typeof p?.description === 'string' && p.description !== '') {
+    param.description = p.description;
+  }
+  if (typeof p?.required === 'boolean') {
+    param.required = p.required;
+  }
+  if (p?.default !== undefined && p.default !== '') {
```
libei, line 362: `p.default !== ''` drops an empty-string default, which flips the parameter from optional back to required across a push/pull.

Verified end to end: `{name: memo, type: String, default: ''}` emits `default: ""` into the aspect (`compact` only drops `undefined`), but this guard discards it on read. The pulled model has no `default`, so the derived tool reports `required: true` where it reported `required: false` before the push, and a call that used to succeed with `memo` omitted now fails.

The empty string is explicitly a value elsewhere in this codebase — `bindScalar`: "An empty String IS a value: `--arg memo=` is the caller saying the memo is blank, which is a different statement from not passing one."

toolbox/mdcode/src/libts/semantic/kc_actions.ts
```diff
@@ -362,6 +378,19 @@ function readParameter(
...
 }
 
+function parseDefaultFromAspect(raw: unknown, type: string): unknown {
+  if (typeof raw !== 'string') return raw;
+  if (type === 'Boolean') {
+    if (raw === 'true') return true;
+    if (raw === 'false') return false;
+  }
+  if (type === 'Integer' || type === 'Float' || type === 'Decimal') {
+    const n = Number(raw);
```
libei: `Number()` over Integer/Float/Decimal throws away the exactness the rest of the runtime keeps on purpose.

Verified round trip: a Decimal default authored as `'0.1000000000000000055'` comes back as `0.1`; `'9007199254740993'` comes back as `9007199254740992`. `bindScalar` then binds the wrong value with no warning. Above 1e21 it is worse — `toString` yields `'1e+21'`, which fails the Decimal regex and turns a working default into a call-time error.

`bindScalar`'s own comments say NUMERIC and INT64 travel as strings because "a JSON number would lose precision above 2^53" and "an exact decimal routed through a JSON number stops being exact". Suggest keeping the aspect's text form and letting `bindScalar` parse it.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -213,9 +219,21 @@ function toolDescription(
...
   // that these rules exist and are not yet checked.
-  const gates = gatingRules(action, model);
+  const gates = gatingConstraints(action, model);
   if (gates.length) {
-    parts.push(`This call is gated by ${joinNames(gates)}.`);
+    const names = joinNames(gates.map(c => c.name));
+    const rules = gates
+        .map(c => {
+          const rule =
+              c.description?.trim() || c.judgment?.trim() || c.expression?.trim();
```
libei: `description` is preferred over `judgment`/`expression` here, but `citation()` in `run_action.ts` quotes `judgment`/`expression` when it refuses. The rule advertised to the agent is then not the rule it is judged against.

A constraint with `description: 'Ask finance first.'` and `judgment: 'The transfer must not exceed the daily limit.'` advertises the first and refuses with the second, so the agent tunes its next attempt against text that is not being evaluated. Suggest matching `citation()`: prefer judgment/expression, and use description as the trailing steer.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -235,38 +253,44 @@ function toolDescription(
...
-  };
+function toolParameter(param: ActionParameter): ToolParameter {
+  const said = param.description?.trim();
+  const required = param.required ?? (param.default === undefined);
+  const out: ToolParameter = param.isEntityRef ?
+      {
+        name: param.name,
+        type: 'string',
+        description: said ?
+            `${said} Give its key, or text that identifies exactly one ${
```
libei: No terminator normalization, so an authored description that doesn't end in punctuation runs into the appended sentence.

Verified: `description: The account the money leaves` (no period — nothing requires one, and the validator only checks non-blank) yields

> The account the money leaves Give its key, or text that identifies exactly one Account; the call fails when nothing matches or more than one does.

`run_action.ts` already has `sentence()` doing exactly this normalization.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -235,38 +253,44 @@ function toolDescription(
...
+                `than one does.` :
+            `Which ${param.type} this applies to. Give its key, or text ` +
+                `that identifies exactly one; the call fails when nothing ` +
+                `matches or more than one does.`,
+        required,
+      } :
+      {
+        name: param.name,
+        type: jsonType(param.type),
+        description: said || `The ${param.name}, as ${article(param.type)}.`,
```
libei: In the scalar branch an authored description *replaces* the type guidance rather than appending to it, so describing a Date or DateTime parameter deletes the only statement of the format `bindScalar` will accept.

Verified: `{name: settledOn, type: Date, description: 'When the transfer settles.'}` yields exactly `When the transfer settles.`, while an undescribed one yields `The plainDate, as a date, YYYY-MM-DD.` — and `bindScalar` hard-rejects anything else (`'settledOn' is a Date, but '03/04/2026' is not one`).

The entity-ref branch above gets this right by appending to `said`. The guide now tells authors to add descriptions, so this fires on the recommended path.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -235,38 +253,44 @@ function toolDescription(
...
-  }
-  return {
-    name: param.name,
-    type: jsonType(param.type),
-    description: `The ${param.name}, as ${article(param.type)}.`,
-    required: true,
-  };
+function toolParameter(param: ActionParameter): ToolParameter {
+  const said = param.description?.trim();
+  const required = param.required ?? (param.default === undefined);
```
libei: This is the second copy of the optionality rule — `run_action.isParameterRequired` is the first, and it isn't exported.

The tool schema's `required` flag and the runtime's enforcement are now two independent statements of one contract, so any refinement applied to one leaves the other advertising something the runtime doesn't do. Suggest exporting `isParameterRequired` and calling it.

toolbox/mdcode/src/libts/semantic/runtime/run_action.ts
```diff
@@ -149,14 +149,20 @@ export interface RunActionOptions {
...
   const action = (model.actions ?? []).find(a => a.name === opts.actionName);
   if (!action) {
     return {
       status: 'error',
       message: `Model '${model.name}' declares no action '${opts.actionName}'.`,
     };
   }
+  const args: Record<string, unknown> = {...opts.args};
+  for (const param of action.parameters) {
+    if ((args[param.name] === undefined || args[param.name] === null) &&
```
libei: Two things in this loop:

1. Substitution runs before `argumentsNotUsable`, so for `{required: true, default: 'USD'}` the tool schema marks the argument mandatory (`toolParameter` honors the explicit `required`) while the requirement is never actually enforced. The guide says "Giving a parameter a default makes it optional", which this combination contradicts — worth rejecting `required: true` alongside `default` in the validator.

2. Treating an explicit `null` as an omission means a caller can never clear a defaulted field: it always gets the default back. Keying on `undefined` only would fix that.

toolbox/mdcode/src/libts/semantic/runtime/run_action.ts
```diff
@@ -734,9 +770,23 @@ function bindArguments(
   const params: Record<string, unknown> = {};
   const types: Record<string, {code: string}> = {};
   for (const param of action.parameters) {
+    const required = isParameterRequired(param);
+    const raw = args[param.name];
+    if (!required && (raw === undefined || raw === null)) {
+      if (param.isEntityRef) {
+        const entity = (model.entities ?? []).find(e => e.name === param.type);
+        const keyField = entity?.fields.find(f => f.name === (entity.keys ?? [])[0]);
```
libei: These two lines are character-for-character lines 814-815 of `bindReference`, and `storeTypeCode` above restates the Integer→INT64 / Decimal→NUMERIC / Date→DATE / default→STRING table that `bindScalar` already owns.

The two tables agree today and nothing keeps them agreeing: add a DATA_TYPE with a code in `bindScalar` and an omitted optional parameter of that type silently binds NULL as STRING. Suggest one `storeCodeFor(type)` that both call, and one shared key-field helper.

toolbox/mdcode/src/libts/semantic/runtime/run_action.ts
```diff
@@ -919,7 +969,9 @@ async function resolveArguments(
   for (const param of action.parameters) {
     if (!param.isEntityRef) continue;
     const raw = args[param.name];
+    const required = isParameterRequired(param);
     if (raw === undefined || raw === null || `${raw}`.trim() === '') {
+      if (!required && (raw === undefined || raw === null)) continue;
```
libei, line 849: An omitted optional entity-ref parameter leaves no entry in `refs`, but `refs` is handed whole to a handler.

Until this PR every entity parameter was guaranteed resolved, so an existing handler doing `refs['approver'].keys[0]` now throws a TypeError inside the transaction — and the outer catch reports it as "failed and was rolled back, but not on the store's account", which sends the reader hunting for a runtime bug. Either pass an explicit absent marker or state the new invariant at the handler boundary.

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -144,6 +147,17 @@ resolve it against `Account`'s key rather than pass a bare number through. A
...
 refers to nothing.
 
+Each parameter may also carry:
+
+- **`description`** — what the parameter means for this call. Required when two
+  parameters on the same action share a type (such as `source` and `target`
+  above, both `Account`), because the type alone cannot tell an agent which
+  argument is which.
+- **`default`** — a fallback value substituted when the caller omits the
+  argument. Giving a parameter a default makes it optional.
```
libei: "Giving a parameter a default makes it optional" isn't true when `required: true` is also set. `toolParameter` computes `param.required ?? (param.default === undefined)`, so the explicit `required` wins and the parameter is advertised as mandatory while `runAction` still substitutes the default. Either document the precedence or reject the combination.

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -736,13 +762,16 @@ any of them stops the push:
...
 
-A model that breaks all four reports all four, one line each, and deploys
+A model that breaks all five reports all five, one line each, and deploys
 nothing:
 
 ```
 Error: action 'TransferFunds' in model 'payments' (payments) has parameter
 'target' whose type 'BankAccount' is neither a known entity nor a scalar
 datatype.
+Error: action 'TransferFunds' in model 'payments' (payments) has multiple
```
libei, line 773: This block can't be produced by any model.

The first error says `target`'s type is `BankAccount`. The new check counts by `param.type`, so that model has Account×1, BankAccount×1, Float×1 — no type repeats, this error never fires, and you get four errors rather than five.

Conversely, in a model where both are `Account` there is no type error at all, and the loop pushes one error for `source` *and* one for `target`, so the block would show two of these lines.

Verified both ways against this branch. A reader who builds the broken model to reproduce the listing never sees it.

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -1246,22 +1275,23 @@ Model 'payments' (payments_eg), profile 'operational':
   action  transfer_funds  (TransferFunds)  [NOT RUNNABLE]
       Move money from one account to another.
 
-      Resolve both accounts before calling. Name the account the money leaves
-      as `source`.
+      Resolve both accounts before calling.
 
-      This call is gated by AmountIsPositive.
+      This call is gated by AmountIsPositive:
+      - AmountIsPositive: A transfer must move at least one unit. Ask the caller
```
libei: This isn't what the CLI prints. `indentBlock` re-wraps each line with `wrapTo(line.trim(), BODY_INDENT)` and passes no `hanging`, so at `LISTING_WIDTH` 79 the actual output is:

```
      This call is gated by AmountIsPositive:
      - AmountIsPositive: A transfer must move at least one unit. Ask the
      caller for the amount again before retrying.
```

— the break falls after "Ask the", and the continuation is flush at 6 spaces rather than the 8 shown here. The parameter lines just below reproduce exactly, which is what suggests this block was hand-edited rather than pasted from a run.

Worth fixing in `printActionTool` as well as here: the wrapped continuation sits at the same indent as a new top-level line, so a multi-rule gate list is unreadable in the listing. `PARAM_CONTINUATION` is already threaded through for parameters.

libei:
Addressed all review findings in `3c57586`:

1. **Blocking: Duplicate-type description check gated to push-only (`validate.ts`)**:
   - `validateActions` now accepts `checkDescriptions: boolean`, enabled in `validatePushRequirements` and disabled in `validateRunnable` so `kcmd action run` does not block existing models lacking parameter descriptions.
   - Grouped duplicate-type parameter checks per type so a model with multiple undescribed parameters of the same type emits one clear error listing the affected parameters.
2. **Knowledge Catalog round-trip data loss (`kc_actions.ts`)**:
   - Parameter defaults are now serialized with `JSON.stringify` and decoded via `parseDefaultFromAspect` (preserving raw string forms when `Number()` or `JSON.parse` would lose precision).
   - Verified round-tripping for `default: ''` (retains empty string and optionality), `default: null`, literal `default: 'null'`, and high-precision decimals/integers (`'0.1000000000000000055'`, `'9007199254740993'`).
3. **Scalar format guidance & description terminators (`agent_tools.ts`)**:
   - Appends format guidance (`As a date, YYYY-MM-DD.`, `As a timestamp in RFC 3339 form.`) after authored descriptions on temporal scalar parameters.
   - Normalizes authored descriptions with `sentence()` before appending guidance or entity-ref instructions.
   - Updated gating constraint rule formatting in `toolDescription()` to prefer `judgment`/`expression` first followed by `description` as trailing steer, matching runtime `citation()` order.
4. **Push-time validation for defaults & `required: true` (`validate.ts`)**:
   - Scalar parameter defaults are now type-checked at validation time via `bindScalar(param, param.default)`.
   - Setting `required: true` alongside `default` is rejected at validation time.
5. **Runtime & CLI consistency (`run_action.ts`, `commands.ts`)**:
   - Exported `isParameterRequired` and `sentence` from `run_action.ts` and reused across `agent_tools.ts` and `commands.ts`.
   - Shared `storeCodeFor` and `entityKeyType` across `bindScalar`, `bindReference`, and `bindArguments`.
   - Keyed default substitution on `args[param.name] === undefined` so an explicit `null` clears a defaulted optional field rather than restoring the default.
   - Documented the omitted optional entity-ref invariant on `ActionContext.refs` (`Record<string, EntityRef | undefined>`).
   - Updated `kcmd action list` (`describeParameter` and `runLine`) to display `default: ...` / `optional` and omit optional/defaulted parameters from suggested `--arg` flags.
   - Added hanging indentation (`BODY_INDENT + '  '`) for bulleted gate lists in `indentBlock()` so wrapped constraint rules align cleanly.
6. **Documentation listings (`docs/semantic-model/actions.md`)**:
   - Updated Section 1 to note that `required: true` alongside `default` is rejected.
   - Updated Section 4's 5-error example so it is 100% reproducible by a single model (`source: Account`, `target: Account` without descriptions, and `amount: Currency`).
   - Updated Section 8's `kcmd agent tools` output and diagram to match exact CLI output with expression rule text and hanging indentation.

All 1,017 tests (`973` unit + `44` tool) and `tsc --noEmit` pass.

toolbox/mdcode/src/tool/commands.ts
```diff
@@ -1509,11 +1515,19 @@ function wrapTo(text: string, indent: string, hanging = indent): string {
 // the difference between passing a value and passing something the runtime has
 // to look up first.
 function describeParameter(p: ActionParameter): string {
-  return `${p.name} (${p.type}${p.isEntityRef ? ', reference' : ''})`;
+  const tags: string[] = [p.type];
+  if (p.isEntityRef) tags.push('reference');
+  if (p.default !== undefined) {
+    tags.push(`default: ${p.default}`);
+  } else if (!isParameterRequired(p)) {
```
libei, line 1522: `default: ${p.default}` renders through template-literal coercion, which erases exactly the three distinctions the KC wire format was changed to preserve in this same commit. Real output from `kcmd action list` on a model carrying all of them:

```
parameters: memo (String, default: ), cleared (Boolean, default: null), label (String, default: null)
```

`memo` has `default: ""` and prints as blank — indistinguishable from a rendering bug or from no default at all. `cleared` has JSON `null` and `label` has the string `"null"`, and they print identically; telling those two apart is the whole reason `parseDefaultFromAspect` moved to `JSON.stringify`.

`JSON.stringify` here fixes all three in one line and costs nothing on the common cases:

```ts
tags.push(`default: ${JSON.stringify(p.default)}`);
```

That gives `default: ""`, `default: null`, `default: "null"`, `default: "hello"`, `default: 0.5` — and, as a side effect, the printed form is the YAML the author would write to get it back.

libei (commented):
Re-checked `3c57586` against all 15 findings. Every one is fixed, and I verified each by running the branch rather than reading it:

- The `validateRunnable` block is gone (`[]` on the model that previously produced 2 errors), and push still errors, now grouped: *“has multiple parameters of type 'Account', so parameters 'source' and 'target' must have a 'description' to distinguish them.”*
- `JSON.stringify` on the wire fixes the whole KC round trip — all 11 cases survive a deep compare, including `""`, `null`, the string `"null"`, `0.1000000000000000055` and `9007199254740993`.
- A described scalar keeps its format guidance: `settledOn: When the transfer settles. As a date, YYYY-MM-DD.`
- Push now rejects an ill-typed `default`, and `required: true` alongside a `default`.
- §4 reproduces exactly — 5 errors, in the order the doc prints them.
- §8 is byte-identical to the real CLI (`diff -u` of the doc block against `indentBlock`/`wrapTo` output: no difference).
- The gate citation now leads with the enforced body: `- AmountIsPositive: amount > 0. A transfer must move at least one unit. …`
- `runLine` emits `--arg` for required parameters only.
- `storeCodeFor`, `entityKeyType`, `isParameterRequired` and `sentence` are each in one place now.

Tests and typecheck confirm your numbers: `973` semantic + `44` tool pass, `tsc --noEmit` clean. The six added tests pin the findings that had no coverage.

Two things came out of the re-check.

**The listing collapses the three default values the wire format was just fixed to keep apart.** Inline below.

**`validate.ts` now imports from `runtime/`.** `import {bindScalar} from './runtime/run_action'` is a new edge from the deploy-time layer into the run-time one, and `run_action` top-level imports `../../gcp/spanner`, `./judge` and `./runtime` — so validating a model now loads the Spanner and Vertex clients. No cycle, and `validate.ts` already reached for `gcp/bigquery`, so this is a note rather than an objection: if you want the direction back, `bindScalar`/`storeCodeFor` would sit fine in a leaf module both layers import.

libei:
Addressed both follow-up notes in `cbaa6af`:

1. **`describeParameter` formatting (`commands.ts:1521`)**: Updated to `default: ${JSON.stringify(p.default)}` so `default: ""`, `default: null`, `default: "null"`, and `default: "USD"` render distinctly in `kcmd action list` (with test coverage added in `action.test.ts`).
2. **Layering between `validate.ts` and `runtime/`**: Extracted `bindScalar`, `storeCodeFor`, `isParameterRequired`, and `sentence` into a pure leaf module (`src/libts/semantic/parameters.ts`) with zero runtime/GCP client imports; `validate.ts` imports from `./parameters` directly and `run_action.ts` re-exports them.

PR #444, opened 2026-09-17
Import an MCP Toolbox configuration into a semantic model

libei:
Adds `kcmd toolbox import <file|dir>`: reads an MCP Toolbox for Databases
configuration and reconstructs the semantic model its tools imply.

## Why it is not a one-to-one mapping

A Toolbox `tools.yaml` is a list of **callable things**. It has verbs and no
nouns — in the published quickstart the word `hotels` never appears as a thing,
only inside SQL, five times, as a table somebody happened to query. A semantic
model is a list of **things that are true**, with the callable surface derived
from them.

So the import cannot map verb to verb. It parses every statement and
reconstructs the nouns out of the SQL inside the verbs:

| in the configuration | in the model |
| --- | --- |
| a table named in any statement | an entity, bound to that table |
| a column attributed to a table | a field, `expression` = the column |
| a join equality in a `SELECT` | a relationship, oriented by the foreign key |
| an `INSERT`/`UPDATE`/`DELETE` tool | an action with a `sql` executor |
| the tool's `parameters` | action parameters, rebound to `@name`, carrying description / required / default |
| the DML target, verb and columns | the action's `affects` |
| the source's coordinates | the entity `source`, and the model's `deployment_target` one level up |
| a `SELECT` tool | nothing |

The last row is the point: a lookup over an entity is generated from the model,
not stored in it. The AlloyDB quickstart's four tools import to two actions and
three lookups nobody wrote, and `kcmd agent tools` lists all five.

## Notes on the mapping

`affects` is **derived, not trusted**. The Toolbox annotations say only
`destructive`; the statement says which table, which verb and which columns,
and `id` is excluded because the statement reads it to find the row rather than
changing it.

Positional placeholders are rebound **by index**, not by order of appearance —
the quickstart's `SET checkin_date = CAST($2 ...), checkout_date = CAST($3 ...)
WHERE id = $1` against parameters `hotel_id, checkin_date, checkout_date` would
otherwise bind the row's identity to a date.

Join direction is decided from evidence in the text: a column naming the other
table is the foreign-key side. Where nothing in the text decides it, the written
order is kept and a warning says so.

Parameter metadata rides on what #442 added, and the description is the string
that mattered — it is what the agent reads to decide what to put in the
argument, and it now reaches the derived tool listing verbatim. Toolbox's
`required` is nil-defaults-to-**true**, the opposite of how an absent flag
usually reads, so it is written out explicitly rather than passed on as an
ambiguity; a defaulted parameter states its default and not its necessity,
since push rejects `required: true` beside a `default`.

`deployment_target` is written from the source coordinates, which is what
separates a model that loads from a model that runs — kcmd asks for one before
it offers an agent any tool. A configuration naming two databases gets none and
a warning naming both.

## What a configuration cannot say

Reported as warnings rather than guessed at, because each is a real difference
between the formats and not an unfinished corner:

- **no primary key** — nothing states one; join columns are named as candidates
  and no key is written, since a wrong key is worse than an absent one
- **no value restriction on a parameter** — `allowedValues` is a rule about the
  argument, and a rule belongs in a constraint guarding the action, which the
  warning says; likewise a parameter filled from a verified claim imports as an
  ordinary one, which is a real loosening and is called out
- **no toolsets** — no construct for a named subset of the surface
- **no templated statements** — `templateParameters` are interpolated before the
  statement is prepared, so the blast radius cannot be declared ahead of the
  call; those tools are skipped
- **nothing non-SQL** — a tool wrapping an API has no statement to read an
  entity out of

A non-Google source (plain `postgres`, `mysql`) produces a model that loads,
validates and publishes to Knowledge Catalog but binds to no store. The import
says so at import time rather than leaving it to the first call.

## Tests

Both published Toolbox quickstart configurations are the fixtures, **unedited**,
so the tests say what a real configuration yields rather than what a
purpose-written one does. 30 tests cover the converter and assert the output
loads via `loadModels` and passes `validatePushRequirements`; 1003 pass across
the semantic suite.

`osi_schema.test.ts` skips the new fixture directory — those are converter input
in a foreign format, the way the OWL inputs are Turtle.


libei:
Closing unmerged. The MCP Toolbox import is not being pursued in this form; the Toolbox comparison work it came out of is recorded separately.

PR #447, opened 2026-09-19
docs(mdcode): record that Knowledge Catalog stores only one binding profile

libei:
fidelity.md is the authoritative map of what survives a push, but it covered only how a binding *prunes* the catalog view — never how many bindings the catalog can hold.

A model may define several physical realizations (one per binding profile), and `kcmd push --all-profiles` deploys a graph for each. The Knowledge Catalog leg does not fan out with it: it runs once, against the profile named by `--profile`, else the default binding (`commands.ts` KC leg). The KC emitter has no notion of a profile at all.

**Commit 1 — record the limit.**
- a **Binding profile** row in the round-trip matrix,
- **footnote 14**,
- a **"One binding, not all of them"** subsection under *To Knowledge Catalog*, which also records the consequence: entry ids are derived from logical names alone, with no profile component, so a second profile push reconciles the catalog to *that* binding rather than adding to it — `reconcileDeletions` deletes an entity or metric the new binding does not answer.

**Commit 2 — drop an orphaned footnote.** A stray paragraph in *To Knowledge Catalog* repeated footnote 11's vendor-dialect text verbatim under a `10` marker, while footnote 10 in the numbered list is "Logical (unbound) model" — which is what the `semantic-entity` matrix cell already resolves to. Dead text under a number that was taken. Footnote refs and definitions are now 1–14 with nothing dangling and nothing unreferenced.

Docs-only; no code or test changes.


libei:
@merr1988 — adding you as a reviewer on this one. (GitHub won't let me file a formal review request since you're not yet a collaborator on this repo, so flagging it here instead.)

Short version: `kcmd` supports several physical bindings per semantic model via binding profiles, and `push --all-profiles` deploys a graph for each — but Knowledge Catalog only ever stores **one** of them. This PR records that in `fidelity.md`, the page that maps what survives a push/pull, along with the consequence that successive profile pushes reconcile-and-delete rather than accumulate.

PR #448, opened 2026-09-19
feat(mdcode)!: trim the action command lines to what curation needs

libei:
`kcmd` is for curating a semantic model, not for dispatching its actions. The
action command lines had drifted into the second job, and this PR trims them
back to the first.

## What goes

Every flag that made `kcmd` behave like a runtime:

| Flag | Was on | Why it goes |
| --- | --- | --- |
| `--judge [model]` | `action-run`, `agent-tools` | Hires a Gemini model from a debugging CLI |
| `--judge-location <region>` | `action-run` | Configuration for a judge there is no longer |
| `--judge-reads-store` | `action-run` | A judge issuing SQL, from a CLI |

And one flag moves rather than goes — `--store`, to `profiles --print-store`
(below).

What is left on the three commands is two flags with no combinations between
them: `action-list` (`--profile`), `action-run` (`--arg`, `--profile`),
`agent-tools` (`--profile`). Across those three, six distinct flags over eight
command-flag pairs becomes two over four; `profiles` gains two.

## `--store` moves and is renamed

`action-list --store` printed where a run would land and exited, suppressing
the listing entirely — a mode flag on a listing command, wearing the name of a
target flag. Nothing on any `kcmd` command line selects a store: `--profile`
picks a binding, and the binding's deployment target decides where writes land.
So a flag called `--store` that takes no value and only prints is one you have
to run before you can tell which direction it goes.

It becomes `kcmd profiles --print-store` — the command that owns bindings, under
a name that says which way it points. Output is byte-identical, so the demo's
setup script changes only the command it calls:

```bash
IFS=/ read -r PROJECT INSTANCE DATABASE <<<"$(kcmd profiles --print-store)"
```

`profiles` also gains `--profile [name]`, which narrows the report to one
profile and is what picks the store `--print-store` prints; without it the move
would drop what `action-list --store --profile alloydb` already did. Naming a
profile the model does not declare exits 1, because an empty report reads as
"this profile withholds nothing" — the opposite of what a typo means.

`profiles` had no flags and no section in `reference.md`; it now has both.

## Flat verbs

`action <command> [name]` and `agent <command>` become `action-list`,
`action-run` and `agent-tools`. cac has no nested subcommands, so the space
form was a positional verb dispatched by hand — the same shape `owl <action>
<file>` still has, and it was being paid for in the help text: five of the six
flags on `action` ended in a parenthetical naming which subcommand they applied
to, which is cac's flag-to-command binding rewritten in English and enforced by
nobody. `kcmd action list --judge-reads-store` parsed clean.

Flat verbs give each command its own flags, its own `--help`, and a misspelling
caught by the parser rather than by a hand-written `if` forty lines into
`commands.ts`.

## What `action-run` does now

It binds the arguments, opens one transaction, applies the statements. Before
it writes, it names the guards it is passing over, so nobody reads a committed
write as a checked one:

```console
$ kcmd action-run IssueCredit --arg order=12346 --arg amount=3.00 --arg memo="Coupon applied late"
Running 'IssueCredit' on projects/my-project/instances/my-instance/databases/semantic_agent_demo...
  NOT CHECKED: CreditWithinOrderTotal, CreditUnderReviewThreshold, CreditMemoNamesAServiceFailure, CreditIsNotSplitToAvoidReview -- this command settles no guard, and the write still happens
Committed at 2026-09-20T18:51:11.964483Z.
```

That is the trade, and it is the useful half for curation: whether an action
binds its arguments, writes the line it says it writes and leaves the store
consistent is a question about SQL, and one command against your own database
answers it with no judge to stand up first.

## What `agent-tools` does now

It derives the tool set with `skipGuards`, so a guarded action is **offered**
rather than marked `[NOT RUNNABLE]`. Who settles a rule belongs to whoever
dispatches the call, and this command cannot know what that will be, so
marking the action unrunnable would describe a caller rather than the model.
The rules are still printed in the tool's own description, wording and all.

A tool is still withheld for what no judge would repair: no executor under the
binding, an executor no handler runs, a guard naming a rule the model never
declares, or a guard naming one that quotes nothing.

## The skills generator

Rebased onto #451, which moved `runLine`/`runFlags` into `run_action.ts` and
uses them to build the command line in a generated Agent Skill. Those no longer
emit a judge flag, and the skill's paragraph under the line now says the line
settles no guard — so an agent that tries the call and watches it commit does
not read that as the rules having held. The flat verbs reach the skill too, so
it says `kcmd action-run`. The golden fixture is regenerated.

## Guards are untouched

`Judge`, `GeminiJudge`, `modelJudgeStore` and `runAction`'s `judge` option are
all unchanged. The demo agent in `demo/semantic-model/agent/` settles all four
of its rules against Gemini, reads the store for the one that needs a number on
record, and reaches all three `on_violation` outcomes live. Only the command
lines stopped pretending to be that application. Putting a judge back behind a
command later is a decision this PR leaves open.

## One defect the live run found

Under `skipGuards` the unsettled-guard loop still warned once per rule, burying
the outcome of the write under a list the caller had just written. The loop is
now gated on `!opts.skipGuards`, with a test.

## Docs

`actions.md` §7 and the agent README's steps 3 and 4 are rewritten. Every CLI
listing in both was regenerated from a live run — the README's against the
demo's Spanner store, `actions.md`'s against the `payments` model rebuilt from
that page's own snippets until `action-list` matched it byte for byte.
`reference.md` gains a `profiles` section, which it never had.

## Verification

`tsc --noEmit` clean; full test suite passes (1049 + 70 + 76 + 21 + 2, 0 fail),
including six new `profiles` tests covering the filter, the typo, the two
non-naming `--profile` forms and both `--print-store` backends; `npm run build`
OK; cross-document anchor check 11 files / 0 problems; the flag table of every
one of the five documented commands diffed against live `--help`.


libei:
Replaced by #454, which is this branch rebased onto `main` after #453 merged. #453 deleted `demo/semantic-model/agent/` and shipped a demo README built on the judge flags this PR removes, so the rebase was not a fast-forward and the head here cannot be moved without a force-push. #454 carries the same two commits plus the demo rewrite.

PR #450, opened 2026-09-20
docs(mdcode): describe action parameters as projected or declared

libei:
## What this changes

An action parameter now carries one scalar value, and gets its definition one of two ways:

- **projected** from a field of the ontology — it names `concept` and `field`, and inherits that field's datatype, description, label and AI context;
- **declared** on the spot — it carries its own scalar `type`, for a value no field holds.

```yaml
parameters:
  - name: source
    concept: Account                  # takes Account.accountId's type
    field: accountId
    description: The account the money leaves.
  - name: amount
    type: Float                       # no field carries this one
    description: How much money to move.
```

`type` is never overridable on a projected parameter. A statement that needs a different type casts in the DML, where the conversion is visible to whoever reads the write instead of buried in metadata.

## What it replaces

Entity-reference parameters: a parameter typed `Account` took a string, and the runtime resolved it to a key inside the write transaction, failing the call when nothing matched or more than one thing did.

A caller that holds a name rather than an id now runs its own lookup before the call. Finding the right row can be a search with several plausible answers, and the place to settle which one is in front of whoever is asking, rather than inside a write transaction that would have to pick one silently and commit to it.

The existence check the resolve step used to perform is replaced by a narrower one the store already answers: a run whose `UPDATE` or `DELETE` matched zero rows fails and rolls back.

This also removes a special case around keys. Every parameter is a scalar, so an action taking the three key fields of a three-part key is as callable as one taking a single id, and how an entity is keyed is no longer a reason a derived agent tool is withheld.

## Sections changed

| Section | Change |
| --- | --- |
| 1 | The example's parameters; new `### Projecting a parameter from a field` |
| 4 | Four distinct parameter errors, and the duplicate-`description` rule restated around projecting the same field |
| 5 | The catalog entry publishes both the reference and the resolved type, and why both |
| 6 | Figure 3 loses the resolve step; zero-row refusal documented |
| 7 | `### Which rows a call touches` rewritten; Table 3 keeps its number |
| 8 | Three withheld reasons instead of four |
| What is not modeled yet | New entry: nothing checks a parameter against the column it is compared with |

## Verification

- `xlinks.py docs/semantic-model/` — 9 pages, 189 anchor links, 0 broken.
- `fencecheck.py` against the base — every fenced-line change is intended.
- The writing checklist's greps plus the short-sentence, garden-path, permission-opener, long-sentence, count-vs-list and width detectors. Five hits fixed.

## Known gap

**The guide leads the implementation.** The CLI listings and the two new validation error strings are written to the shape this design implies and have not yet been diffed against a live run. The implementation is in progress on a separate branch and will be stacked on this one; any string that comes back different will be fixed on whichever side is wrong.


libei:
Superseded by #452, which carries these three `docs(mdcode)` commits unchanged at the bottom of its branch, plus the implementation and its review fixes.

Nothing here is lost: `git log kcmd-action-param-projection..kcmd-action-param-projection-docs` is empty, so #452 strictly contains this branch. The docs commits are still the first three commits there and still read as the guide for what the code does.

Review #452 instead.

PR #452, opened 2026-09-20
feat(mdcode)!: project an action parameter from a field

libei:
Seven commits, in reading order. The three `docs(mdcode)` ones at the bottom
are the guide for what the rest builds, so read those first; then
`feat(mdcode)!: project an action parameter from a field`, and the three
`fix(mdcode)` commits on top of it, which repair what two rounds of review
found. The last of those is the only one that is not about parameters --
see **The listing asks the runtime** below.

Those three docs commits were #450 until this PR grew to contain them unchanged,
at which point keeping both open only split the review. #450 is closed.

An action parameter now always carries one scalar value, and there are two ways
to say what that value is.

**Projected** — names a `concept` and a `field`, and takes that field's
datatype, description, label and AI context out of the model:

```yaml
parameters:
  - {name: source, concept: Account, field: accountId}   # Integer, from the field
  - {name: amount, type: Float}                          # no field holds this one
```

`concept` and `field` are two keys rather than one dotted string, so a name
with a dot in it stays writable and the pair reads the way `affects` already
does. Leave `name` out and the parameter answers to the field's.

**`type` is never overridable on a projected parameter.** Stating one beside
`concept`/`field` is a load-time error, not an override: two copies of one fact
drift. If the type is wrong, the field is what to fix; if a statement needs a
different one, the DML is where to cast it, in view of whoever reads the write.

## What this replaces

Entity-reference parameters, which took text and resolved it to a row inside
the transaction. Gone with them: `isEntityRef`, `EntityRef`, `resolveEntityRef`,
`identifyingColumn`, `bindReference`, `entityKeyType`, `outcome.refs`,
`actedOn`, the composite-key refusal, and the lookup `SELECT` the resolve ran.

What the existence check bought is now bought by the write itself: a statement
the store reports zero rows affected for fails the run and rolls back, so a
caller who names a row that is not there is told so rather than told the write
succeeded. `INSERT` is the one exemption, because it creates rows rather than
finding them, so writing none is something an author can mean. A store
reporting no count at all is left alone.

That exemption is the only way past, which is the opposite of how the check
first read. Refusing only a statement whose verb scans as `UPDATE` or `DELETE`
leaves everything the scanner cannot read — a `MERGE`, a procedure call
wrapping the write — passing silently, and passing silently is what a missed
write looks like. Checking instead for the one verb allowed to write nothing
puts the cost on the other side: a misread can refuse a write that was fine,
never hide one that did nothing. A false refusal is a failed run someone looks
at; a missed refusal is a caller told its write landed when it did not.

Composite keys stop being a special case. An entity keyed by three columns
takes three projected parameters, one per key field — the same declaration a
single-column key writes, three times — where the old resolve refused the
entity outright.

## Two consequences worth calling out

- **Pruning inverts.** A parameter projecting from an entity a profile cannot
  bind no longer drops its action. The parameter took a copy of the field's
  definition when the model loaded, so at run time it asks that entity for
  nothing. `resolve_profiles.ts` and its test both changed direction.
- **A terminator bug in `agent_tools`, fixed.** A parameter description was only
  given a full stop where temporal guidance followed it. One authored without
  one is read by a model alongside every other, and the odd one out reads as a
  fragment of the next line.

Inheritance is resolved eagerly into the IR by the loader, so an emitter cannot
tell an inherited description from an authored one. The OSI, pull and KC
emitters therefore write the inherited `description` but suppress the resolved
`type` — restating it would make the document they produce unloadable. The pull
golden was checked by hand for exactly this: it still round-trips to something
the loader accepts.

## The second commit: what review found

Eleven defects, all fixed here. Four are worth reading before the diff.

**The aspect template was renumbered.** Removing `isEntityRef` shifted every
parameter field after it, which the header of `kc_custom_types.ts` forbids in as
many words: the templates are append-only, and renumbering breaks the `PATCH`
for every project that has already run `kcmd init`. The original indices are
restored; the retired field stays at index 3 as `Is Entity Reference (reserved)`,
described as never written. A new test pins every shipped index, so the next
edit to that file has to be an append.

**The zero-row refusal could be walked past.** It read the verb with
`sql.trimStart().slice(0, 6)`, so a statement opening with a `--` comment, a
block comment, or a `WITH` clause ahead of its `UPDATE` reported a verb nobody
checks — and every one of those failures says the statement is harmless.
`leadingVerb` now walks the text, skipping comments, quoted strings and anything
nested in parentheses. `UPDATE t SET note = 'DELETE'` still reads as `UPDATE`.

**The emitted document could be unloadable.** The OSI emitter wrote
`concept`/`field` whether or not the document it was writing still declares that
field — a profile that prunes the concept leaves a projection pointing at
nothing. It now emits the projection only when the field resolves in what it is
about to write, and otherwise degrades to the resolved `type`. The check reuses
the loader's own `declaredConceptFields`, so the two cannot disagree about what
`extends` contributes.

**The confusable-parameter check was keyed too narrowly.** It grouped by
`concept.field`, so a projected and a declared parameter of one type — the case
a caller is most likely to swap — passed. It keys on the resolved type now,
which is strictly broader, since a projected parameter takes its field's type.

The rest: a composite value bound as `[object Object]` and is refused by name
now; `parameterTypeErrors` told the author of an untyped projection to project
from a field, which is what they did; the loader's bare catch made every
"does not declare that field" warning suspect, because the fallback drops
`extends` model-wide, so those warnings are gated on the resolve succeeding;
`readParameter` invented a `type: ''` third state on pull; and
`kcmd action list` printed the JS word `undefined` for a parameter with no type.

On the docs: the profiles.md walkthrough did not load — its parameter named an
entity as `type`, and rewriting it as a projection exposed that `Order.key`
declares no datatype either. README.md and model_spec.md still described an
entity-typed parameter as an object reference. actions.md said a `concept` may
name a relationship without saying nothing can be projected from one, since a
relationship only has fields when a junction table backs it and the format has
no syntax for one. fidelity.md now records what a round trip loses.

Two findings were checked and left alone: the `type` beside `concept`/`field`
in the actions.md KC listing is the aspect, where the resolved type and the
reference legitimately coexist; and the "reports all five" claim is scoped to
five validation rules that do emit together, verified against a live run.

## Verification

`npx tsc --noEmit -p .` clean; 1105 tests pass, 0 fail across the whole suite,
14 of them new with the fix commits. (`npm test` shows 7 failures in
`store.test.ts`; that file shells out to `gcloud` three times per test at
5-7s each and the npm script does not pass `--timeout`. Run at
`--timeout 30000` it is 13 pass, 0 fail, and the file is untouched by this
branch.) Every golden diff was read before regenerating rather than accepted
from `UPDATE_GOLDENS=1`.

Each new test was shown to be load-bearing rather than assumed: `leadingVerb`
was temporarily reverted to `slice(0, 6)` and the suite re-run, which failed 3
of the 4 new verb tests (the 4th guards against over-matching and passes under
both, by design).

The profiles.md walkthrough was extracted into a workspace and driven through
the built `dist/kcmd action list`, which is the only reason the missing
`datatype` on `Order.key` was caught: the doc looked right and printed
`order (undefined from Order.key)`.

Beyond the suite, the demo model and the guide's own `payments` model were both
driven through a real CLI: `action list`, `agent tools`, `action run` and
`push --validate-only` over a deliberately-broken copy. Six listings on the
guide did not match what kcmd prints and are fixed in the docs commits here.

The `NOT RUNNABLE` example in the guide was caught the same way. It first went
in naming an action the `payments` model does not declare -- lifted from a test
fixture, so it read as more of the listing above it while being output no reader
could reproduce. It is now `TransferFunds` under a misspelled guard, captured
from a run against the workspace the guide builds; both listings in that section
are byte-identical to that run.

## The listing asks the runtime

`kcmd action list` decided for itself whether an action can run, and the rule
it used was "does this have an executor". The runtime's rule is much larger, so
the listing printed a working-looking `run:` command for calls that are refused
before a transaction opens: one executed over MCP, which this command holds no
handler for, and one guarded by a constraint the model never declares. Both now
print `NOT RUNNABLE:` carrying the runtime's own sentence, from the same
`whyRefusedWithoutRunning` that `agent tools` already calls. Two copies of "can
this run" drift, and neither direction of the drift is visible to the reader --
a run line for a call that always fails, or a withheld action that would have
worked.

The question is asked **with a judge in hand**. `runLine` appends `--judge` for
a judged action, so assessing runnability without one would mark every judged
action unrunnable and contradict the command printed directly beneath it. The
stand-in judge throws rather than answering: nothing in a listing may reach a
judge, and a path that got that far should say so rather than settle a rule
with a stub verdict.

The store is deliberately not part of the question. Whether one is reachable is
the same sentence on every action in the model, so it says nothing about any of
them, and the listing already says it once at the top.

## Also from the second review round

- **validate**: an entity named where a datatype belongs now triggers the
  concept-table lookup, so the author is told to project a field rather than
  just "not a scalar datatype". That table is built lazily, and what asks for it
  decides which message the author sees. Separately, a projection whose concept
  table would not build -- a dangling `extends`, which is reported on its own
  line -- is no longer blamed on the field it projects from; blaming it sends
  the author to a field that is correctly typed and away from the actual fault.
- **osi_converter**: warn when a parameter would emit with neither a projection
  the document carries nor a recorded datatype, instead of writing a document
  that will not load. `emitted` is hoisted out of the per-action map -- it
  deep-clones the model to resolve inheritance and does not vary by action.
- One finding was **refuted with evidence rather than fixed**: deleting the
  scanner's extra handling would blind the row-count check on every handler
  plan, which never passes through validation at all. A second refusal -- that
  sharing the scanner with validation would admit `MERGE` -- did not survive the
  next round and is reversed below.
- The formatter had mangled two declarations in the previous commit --
  `Executor` and a `new Map<>` -- in files those changes never touched. Both are
  restored to what `main` has.

## From the third review round

Five findings, each with tests that fail without the fix.

**Validation and the runtime disagreed about where a statement's verb is.**
`kcmd action run` skips comments and CTEs; validation took the first word. I had
refused this in the previous round on the grounds that sharing the scanner would
admit `MERGE` -- which is wrong: `SQL_EXECUTOR_VERBS` is
`['INSERT','UPDATE','DELETE']`, so a `MERGE` the scanner reads is still rejected
at publish time. The verb set stays deliberately wider than what may be
published, so a caller can be told a `MERGE` matched nothing. And the divergence
was real in both directions -- `run_action.test.ts` runs *authored* `sql`
executors with a leading `--` comment and with a CTE ahead of the verb, so the
library accepted executors that `kcmd push` then refused. `leadingVerb` moves to
`sql_identifiers.ts` as `leadingDmlVerb`, and both callers read it. The message
keeps naming the verb the author actually wrote: the scanner reads DML verbs
only, so a `SELECT` executor comes back empty, and "has no readable DML verb"
would describe a query the author can see perfectly well.

**That scanner skipped `--` but not `#`.** Spanner speaks GoogleSQL, which takes
both, so `# credit the account` ahead of an `INSERT` handed back whatever verb
the prose used -- and an `INSERT` that legitimately writes no row was refused as
a zero-row write. That is the expensive direction of the mistake.

**A single projection was blamed for a duplication.** One projected parameter
sharing a datatype with a described parameter was reported as "multiple
parameters projected from `Account.accountId`", sending the author to look for a
second projection to delete. The field is named only when the indistinct bucket
is all one projection *and* more than one parameter really came from that field.

**A typeless field was emitted as a resolvable projection.** Doing so suppresses
the parameter's resolved `type`, so the round-trip hands back a parameter with
neither a datatype nor a resolvable projection -- which the loader rejects. Such
a pair is now treated as unresolvable, falling the parameter back to its
resolved type, which is what the fallback is for.

**A push now says which projections a pull will not give back**, and which of
three reasons applies -- the difference between an author deleting a projection
and an author binding a column:

- the concept is not published at all (pruned, or abstract);
- the concept is published but the **field** is not. `pruneUnavailable` drops an
  unbound field while keeping the entity whose key still binds, and since a
  projected parameter no longer prunes its action, the entry goes out naming a
  field the `schema` aspect does not carry;
- the concept is a relationship. Pull rebuilds relationships from `schema-join`
  entry links alone, and a link records its two endpoints and nothing else, so
  an association's own fields come back from no entry.

The published-field half is asked of the loader's own `declaredConceptFields`,
so the answer accounts for `extends`; a model that will not resolve returns
`null` and that half is skipped, since an emitter is not the place to discover
that a model does not load.

## One thing this does not let you write

The zero-row rule applies per statement and has no opt-out, so a multi-statement
action whose *earlier* statement is legitimately conditional cannot be written
today. An action that clears a cart and then writes an order fails outright when
the cart was already empty. The guide now says so and gives the two workarounds
(split it in two, or move the condition into a statement that must write).

Whether a per-statement "may write nothing" marker is worth adding is a model
question rather than a bug, so I have documented the ceiling and left the
construct alone.


libei (commented):
Reviewed the PR across the IR, loader, validator, runtime, emitters (`osi_converter` / `kc_actions` / `kc_custom_types`), CLI commands, docs, and test suite.

The core design shift—replacing runtime entity-reference resolution with load-time field projection (`{concept, field}`) plus fail-closed zero-row DML checks (exempting only recognized `INSERT`s)—is very clean and well-tested, and the aspect-template index preservation in `kc_custom_types.ts` is well-guarded by the new index-pinning tests.

Found a few logic/consistency bugs in `validate.ts`, `run_action.ts`, `osi_converter.ts`, and `kc_actions.ts`—left inline comments with details and suggested fixes.

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -216,55 +219,91 @@ function validateActions(
...
+  // inheritance.
+  //
+  // `fieldsPruned` stands the whole lookup down, which is what a profile push
+  // needs: pruning removes unbound fields and unavailable entities, so the
+  // concept a parameter projects from may be gone from the model in hand even
+  // though the author's document names one that exists. A parameter needs only
+  // the logical definition and the loader already copied it down, so nothing
+  // about the push depends on resolving the reference a second time here.
+  const needsConcepts =
+      actions.some(a => a.affects?.length || a.parameters.some(p => p.concept));
```
libei: **Bug: `needsConcepts` is `false` when an action only has a parameter with `type: <EntityName>`, making the entity/relationship branch in `parameterTypeErrors` unreachable in that case.**

`parameterTypeErrors` (line 490) checks `const named = concepts?.get(param.type)` so that `{name: 'order', type: 'Order'}` produces the targeted migration message (`which is an entity rather than a scalar datatype. A parameter carries a value, so take one from Order by projecting it: {concept: Order, field: <field>}`).

However, `needsConcepts` only checks `a.affects?.length || a.parameters.some(p => p.concept)`. On an action that has neither `affects` nor `p.concept` (e.g. an action whose only parameters use the old `type: Order` spelling or scalar types), `needsConcepts` evaluates to `false`, `concepts` is `undefined`, and `parameterTypeErrors` falls back to the generic `'Order' is not a scalar datatype (String/...)` error.

Suggested fix:
```ts
  const needsConcepts = actions.some(
      a => a.affects?.length ||
          a.parameters.some(
              p => p.concept ||
                  (p.type !== undefined &&
                   !(DATA_TYPES as readonly string[]).includes(p.type))));
```

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -216,55 +219,91 @@ function validateActions(
...
+        const indistinct = params.filter(
+            p => shares.get(p.description?.trim() ?? '')! > 1 ||
+                !p.description?.trim());
+        if (indistinct.length > 0) {
+          const names = indistinct.map(p => `'${p.name}'`);
+          // Name the projection only when every one of them came from the one
+          // field; a mixed bucket has nothing in common but the datatype.
+          const projections = new Set(indistinct.map(
+              p => p.concept !== undefined ? `${p.concept}.${p.field}` : ''));
+          const from = projections.size === 1 ? [...projections][0] : '';
```
libei: **Bug: `projections` is computed over `indistinct` instead of `params` (or without checking `indistinct.length > 1`), causing a false `"has multiple parameters projected from 'X.y'"` error when only one parameter is projected from `X.y`.**

Consider an action with:
1. `{name: 'account', concept: 'Account', field: 'accountId'}` where `Account.accountId` is an `Integer` field **without** a description (`account.description` is `undefined`), and
2. `{name: 'quantity', type: 'Integer', description: 'How many.'}` (a described standalone `Integer`, or a parameter projected from a *different* described `Integer` field).

In the `identity = 'Integer'` bucket:
- `params` has 2 parameters (`account` and `quantity`).
- `indistinct` has only **1** parameter (`[account]`), because `quantity` has its own description.
- `projections` is computed over `indistinct` (`new Set(['Account.accountId'])`), which has `.size === 1`, so `from` becomes `'Account.accountId'`.
- The emitted error says:
  `has multiple parameters projected from 'Account.accountId', so parameter 'account' must have a 'description' of its own to distinguish it.`
  even though only **one** parameter was projected from `Account.accountId`.

Computing `projections` over `params` (or requiring `indistinct.length > 1` / verifying that `from` actually appears more than once in `params`) avoids blaming a single projection:
```ts
          const projections = new Set(params.map(
              p => p.concept !== undefined ? `${p.concept}.${p.field}` : ''));
```

toolbox/mdcode/src/libts/semantic/validate.ts
```diff
@@ -324,8 +363,9 @@ function sqlExecutorErrors(action: Action, where: string): string[] {
     }
     const verb = text.split(/\s/, 1)[0].toUpperCase();
```
libei: **Inconsistency with `leadingVerb` in `run_action.ts`: `sqlExecutorErrors` rejects statements that start with comments (`--`, `/* */`) or `WITH` CTEs.**

In `run_action.ts`, `leadingVerb` was updated to walk past leading `--` line comments, `/* */` block comments, and `WITH` CTEs (`WITH stale AS (...) UPDATE ...`), and `run_action.test.ts` tests `sql` executors with those forms.

However, `sqlExecutorErrors` here still extracts the verb via `text.split(/\s/, 1)[0].toUpperCase()`. Because `validatePushRequirements` (`kcmd push`) and `validateRunnable` (`kcmd action run` in `commands.ts`) both run `sqlExecutorErrors`, any authored `sql` executor statement that starts with `--`, `/*`, or `WITH` is rejected here (`starts with '--', but a statement must be one of INSERT, UPDATE, DELETE`) before `runAction` is ever reached.

Sharing `leadingVerb` (from a common helper such as `sql_identifiers.ts` or `parameters.ts`) between `sqlExecutorErrors` and `noRowMatched` will keep validation and runtime execution aligned.

toolbox/mdcode/src/libts/semantic/runtime/run_action.ts
```diff
@@ -394,6 +374,128 @@ export async function runAction(opts: RunActionOptions):
...
+// and AlloyDB, so it would have to parse Spanner statements as something else
+// and would reject valid ones. Given which direction a misread now falls, a
+// scanner that recognizes a leading INSERT is enough; if that stops being true,
+// the parser is the thing to reach for rather than more cases here.
+function leadingVerb(sql: string): string {
+  const verbs = new Set(['INSERT', 'UPDATE', 'DELETE', 'MERGE']);
+  let depth = 0;
+  for (let i = 0; i < sql.length; i++) {
+    const c = sql[i];
+    if (c === '-' && sql[i + 1] === '-') {
```
libei: **Edge case in `leadingVerb`: Spanner (GoogleSQL) `#` line comments.**

Action DML targets both AlloyDB (PostgreSQL) and Spanner (GoogleSQL). In GoogleSQL, `#` introduces a single-line comment just like `--`. If a Spanner `INSERT` statement starts with a `#` comment that happens to mention another DML word (e.g. `# update audit trail\nINSERT INTO ...`), `leadingVerb` scans words inside the `#` comment at `depth === 0` and returns `'UPDATE'`, causing `noRowMatched` to refuse a zero-row `INSERT`.

Adding `(c === '-' && sql[i + 1] === '-') || c === '#'` handles both line-comment forms.

toolbox/mdcode/src/libts/semantic/osi_converter.ts
```diff
@@ -281,30 +281,87 @@ function metricDoc(metric: Metric, warnings: string[]): Record<string, any> {
...
+function emitted(model: SemanticModel): Set<string>|null {
+  let concepts;
+  try {
+    concepts = declaredConceptFields(model);
+  } catch {
+    return null;
+  }
+  const pairs = new Set<string>();
+  for (const [name, concept] of concepts) {
+    for (const field of concept.fields.keys()) pairs.add(`${name}.${field}`);
```
libei: **Edge case: `emitted(model)` marks a `Concept.field` pair as resolvable even when that field has no `type` (`field.type === undefined`).**

If `concept.fields` contains a field whose `type` is `undefined`, `emitted(model)` still adds `${name}.${field}` to `pairs`. Then `parameterDoc` sets `projects = true`, suppresses `p.type`, and emits `{concept, field}` without `type`. On reload, `convertParameter` resolves the projection to `type: undefined`, turning a parameter that carried a resolved scalar `p.type` into an untyped parameter that fails validation.

Only adding fields that actually carry a datatype prevents emitting a projection that won't supply a type on reload:
```ts
    for (const [field, def] of concept.fields) {
      if (def.type !== undefined) pairs.add(`${name}.${field}`);
    }
```

toolbox/mdcode/src/libts/semantic/kc_actions.ts
```diff
@@ -113,17 +113,22 @@ export function actionEntries(
...
     for (const p of action.parameters ?? []) {
-      if (p.isEntityRef && !ctx.publishedEntities.has(p.type)) {
-        warnings.push(
-            `model '${model.name}': action '${action.name}' parameter ` +
-            `'${p.name}' refers to entity '${p.type}', which this push does ` +
-            `not publish (abstract or unavailable), so a pull will not ` +
-            `recover it as an entity reference.`);
-      }
+      if (!p.concept) continue;
+      if (ctx.publishedEntities.has(p.concept)) continue;
```
libei: **Missed warning when `p.concept` is published (or is a relationship) but `p.field` is not published.**

Two cases slip past these `continue` guards without emitting the warning even though `pull` will not recover the projected field:
1. **`p.concept` is a published entity, but `p.field` is unbound:** `pruneUnavailable` drops unbound fields from `entity.fields` while keeping the entity (if its key is bound) and keeping the action (since projected parameters no longer prune actions). `ctx.publishedEntities.has(p.concept)` is `true`, so no warning is emitted, even though `p.field` is absent from the published `schema` aspect and `pull` will degrade the parameter to a standalone scalar.
2. **`p.concept` is in `relationshipNames`:** KC `pull` reconstructs relationships only from `schema-join` links (1:1 / 1:N) which carry no edge fields, so a field projected from a relationship is never recovered on `pull`.

toolbox/mdcode/src/libts/semantic/validate.ts
libei: Right, and already fixed in `3681fbe` (one commit before the review's base) --
`needsConcepts` now also fires on a parameter whose `type` is not a built-in:

```ts
const needsConcepts = actions.some(
    a => a.affects?.length ||
        a.parameters.some(
            p => p.concept ||
                (p.type &&
                 !(DATA_TYPES as readonly string[]).includes(p.type))));
```

so a model whose only concept reference is `type: Account` still builds the
table, and still gets the "not a declared concept" message rather than the
datatype message.

toolbox/mdcode/src/libts/semantic/validate.ts
libei: Fixed in `7706189`, though a bit tighter than suggested.

Computing `projections` over `params` fixes the mixed bucket but introduces the
opposite miss: a bucket holding *one* projection beside a described parameter of
the same datatype would still be blamed on the field, and "multiple parameters
projected from `Account.accountId`" sends the author hunting for a second
projection that isn't there.

So the field is named only when the bucket is all one projection **and** that
field is what more than one parameter came from:

```ts
const pairOf = (p: ActionParameter) =>
    p.concept !== undefined ? `${p.concept}.${p.field}` : '';
const projections = new Set(indistinct.map(pairOf));
const only = projections.size === 1 ? [...projections][0] : '';
const from =
    only !== '' && params.filter(p => pairOf(p) === only).length > 1 ? only : '';
```

Two tests cover it: a lone projection beside a same-typed described parameter
gets the generic message, and two parameters really projected from one field
still get the field named.

toolbox/mdcode/src/libts/semantic/validate.ts
libei: Confirmed and fixed in `7706189`. I had previously argued that sharing the
scanner would quietly admit MERGE -- that was wrong. `SQL_EXECUTOR_VERBS` is
`['INSERT','UPDATE','DELETE']`, so a MERGE the scanner reads is still rejected
here; the verb set stays deliberately wider than what may be published, so a
caller can be told a MERGE matched nothing.

And the divergence was real in both directions: `run_action.test.ts` runs
*authored* `sql` executors with a leading `--` comment and with a CTE ahead of
the verb, so the library was accepting executors that `kcmd push` refused.

`leadingVerb` moved to `sql_identifiers.ts` as exported `leadingDmlVerb`, and
both `validate.ts` and `run_action.ts` call it. One message detail had to be
kept: the scanner reads DML verbs only, so a SELECT executor -- the likeliest
mistake in this position -- comes back empty, and "has no readable DML verb"
would describe a query the author can see perfectly well. The first bare token
is named when there is one:

```ts
const first = text.split(/\s/, 1)[0].toUpperCase();
const wrote = verb || (/^[A-Z_][A-Z0-9_]*$/.test(first) ? first : '');
```

Seven tests now assert that a published statement and a run read the verb the
same way, including MERGE-still-rejected and SELECT-still-named as negative
controls.

toolbox/mdcode/src/libts/semantic/runtime/run_action.ts
libei: Fixed in `7706189`. Spanner speaks GoogleSQL, which takes `#` as a line comment,
and action DML runs against Spanner -- so `# credit the account\nINSERT ...`
returned `'UPDATE'`, losing the INSERT exemption and refusing a legitimate
zero-row write. That is the expensive direction.

```ts
if ((c === '-' && sql[i + 1] === '-') || c === '#') {
```

with a test for `#` alongside the existing `--` and block-comment cases.

toolbox/mdcode/src/libts/semantic/osi_converter.ts
libei: Fixed in `7706189`. Worse than cosmetic: emitting the pair suppresses the
parameter's resolved `type`, so the round-trip hands back a parameter with
neither a datatype nor a resolvable projection -- which the loader rejects.

```ts
for (const [field, def] of concept.fields) {
  if (def.type !== undefined) pairs.add(`${name}.${field}`);
}
```

Treating the pair as unresolvable falls the parameter back to its resolved type,
which is what the fallback is for, and warns when there is none. Two tests.

toolbox/mdcode/src/libts/semantic/kc_actions.ts
libei: Confirmed and fixed in `7706189`, and there turned out to be three distinct
losses, not two -- naming which one applies is the difference between an author
deleting a projection and an author binding a column:

- the concept is not published at all (pruned, or abstract);
- the concept is published but the **field** is not -- `pruneUnavailable` drops
  an unbound field while keeping the entity whose key still binds, and since a
  projected parameter no longer prunes its action, the entry goes out naming a
  field the `schema` aspect does not carry;
- the concept is a relationship. Pull rebuilds relationships from `schema-join`
  entry links alone (`pull_kc.ts:99`), and a link records its two endpoints and
  nothing else, so an association's own fields come back from no entry.

The published-field half is asked of the loader's own `declaredConceptFields`,
so the answer accounts for `extends`; a model that will not resolve returns
`null` and that half is skipped, since an emitter is not the place to discover
that a model does not load. Three tests, including a clean negative control.

PR #455, opened 2026-09-21
refactor(mdcode): take the store-reading judge out of the runtime

libei:
Rebased onto `main` now that #454 is merged, so the diff is one commit and nothing else.

## What goes

A judge could be handed a SQL tool and sent to look up a fact the call did not carry. Three pieces implemented it:

- `modelJudgeStore()` in `src/libts/semantic/runtime/judge_store.ts` — built the tool from the binding profile
- `JudgeStore` / `JudgeQueryResult` in `runtime/judge.ts` — the seam
- `MAX_READS`, `READ_STORE`, `_gather`, `answerFor` and the two-phase system instruction in `gcp/gemini.ts` — the read loop

Nothing in the library constructs one. The only caller was the command line’s `--judge-reads-store`, and #454 deleted that flag, so this is already dead code on `main`.

## Why

It is a second way for a model to reach the database — a read budget, a row cap, a value cap and a tool-calling loop to maintain — for a capability no consumer asks for. The wrap bounded neither which tables a statement could name nor a query that calls a writing function, so on AlloyDB a write could land through it.

## What a judge sees now

The rule, the action and the arguments. A rule about a stored value has no binding point here: it is recorded in the model, where the policy belongs, and enforced in the schema, inside the transaction the write opens.

That is a real loss, and the docs say so rather than eliding it. `CreditWithinOrderTotal` in the `actions.md` worked example and in the commerce demo is exactly this kind of rule; `CreditIsNotSplitToAvoidReview` reads on the credits already on the order and is the same case. Both are now marked declared-and-not-enforced, alongside the order-total rule that already was.

- `actions.md` — `### When the judge needs a fact` (73 lines) replaced by `### A rule the judge can’t settle` (24)
- `demo/.../skill/README.md` — the store contrast in §4 and the limits in §10 re-framed
- `reference.md` — `main`’s own rewrite of the `action-run` paragraph already dropped the reading claim, so it is taken as-is

Transcripts recorded when a judge could read are kept, not deleted: a run that happened is evidence. The lines about reading are marked as no longer reproducible.

## Also

`readableEntities` moves from `judge_store.ts` to `agent_tools.ts`, which is the other reader of it — the generated skill’s account of what there is to read has to match what the lookup tools serve. Rather than leave a stub file named for a deleted concept.

## Checks

- `npx tsc --noEmit` clean, `npm run build` clean
- `npm test`: 1,186 pass, 0 fail, against `main`’s 1,224. The 38 removed are the 30-test `judge_store` suite and 8 store-reading cases in `gemini.test.ts`.
- Every markdown link and anchor across the repo resolves, including the two inbound links retargeted to the new heading.

Net: -1,153 lines.


libei:
Second commit `15ac335` addresses a code review of this branch. Removing the reading judge changed what a judgment can settle but left nine places — section 2 of the actions guide, `model_spec.md`, and three judgment texts including the commerce demo's, which `skills-generate` copies verbatim into a generated skill — still telling an author that a judge goes and reads a row.

The sharpest one: "What a guard can see" said a guard sees "the arguments it was handed, and the data the call starts from", eight lines above the paragraph this PR added saying a judge sees the arguments and no more.

Also flagged and now stated plainly in the guide: `TransferWithinAvailableBalance` is the guide's running example, is guarded `reject`, and turns on a stored balance — so as written it refuses every `TransferFunds` call. The example keeps the rule, because a balance check is the honest shape of the policy and the model is where policy is recorded, and now says it is a policy on record rather than a control. Whether the guide should instead teach a settleable rule as its running example is a bigger call and not made here.

1,186 tests still pass.

PR #456, opened 2026-09-21
docs(mdcode): move the runtime semantics out of "Run it" and into section 8

libei:
Stacked on #455 (`drop-reading-judge`), which this branch contains; GitHub will
not take a fork branch as the base of an upstream PR, so the base here is `main`
and the diff to review is the last commit.

`kcmd action-run` is a debugging facility — it binds arguments, opens one
transaction and settles no guard. Section 7 of the actions guide had grown to
346 lines, 21% of the guide, and most of what it carried was not about the
command. Every subsection past "What a run does" opened by saying some form of
"this is what a runtime does, and this command is not that runtime", then
described the runtime anyway. That put the guard semantics in front of the
reader in the one section that does not use them, and left section 8 — where a
write tool actually puts a rule to a judge — pointing backwards for them.

**Moved into section 8**, with their openings reframed around the write tool
rather than around what the command line omits:

- When a rule stops the call
- When the rule is a sentence
- A rule the judge can't settle
- Which rows a call touches

**Dropped** rather than moved:

- Figure 3, the ASCII trace of a run. The prose under it already says what it
  says, and the unchecked-guards warning the run prints says it in the
  runtime's own words — that warning is shown instead.
- Table 3, the two-column argument-versus-`WHERE` table. The two paragraphs
  below it are the content.
- The three judge transcripts. They were recorded when a command line still
  hired a judge, were already marked non-reproducible by #455, and
  `demo/semantic-model/skill/README.md` carries live runs of all three
  `on_violation` outcomes against the same model.

Section 7 is now 121 lines and the guide is 78 shorter. Table 4 is renumbered
to 3.

## Verification

No code changed. `npx tsc --noEmit` clean, `npm run build` clean, `npm test`
1,186 pass / 0 fail — unmoved, as a doc-only change should be. A repo-wide
markdown anchor sweep reports 0 broken links under `toolbox/mdcode`; every
moved heading keeps its slug, so inbound links from `reference.md` and the demo
README still resolve.


libei:
Rebased onto #455, which grew a second commit from a code review. This branch now carries three commits; the restructure is the last one, `8282a9a`.

One review finding landed here rather than on #455: the sentence replacing Figure 3 claimed the commerce demo "records both of the first two [outcomes] against a live store, warning and all". Both of the demo's `kcmd action-run` transcripts are committed runs — the only "nothing written" outputs there are judge refusals recorded through the old `kcmd action run --judge`, explicitly marked non-reproducible, and none carries the unchecked-guards warning. Corrected to say what is actually recorded.

`npx tsc --noEmit` clean, `npm run build` clean, `npm test` 1,186 pass / 0 fail, 0 broken anchors under `toolbox/mdcode`.

PR #457, opened 2026-09-21
refactor(mdcode)!: remove kcmd action-run and action-list

libei:
`kcmd action-run` and `kcmd action-list` existed as a CLI debugging surface over the semantic runtime: `action-run` bound arguments and executed SQL statements while skipping every guard (`skipGuards: true`), and `action-list` existed to print the `kcmd action-run` command line for each action.

Keeping that unguarded CLI write path alongside the runtime's guarded agent tools required every guide, skill, and reference page to explain two competing execution paths — what the model declares that a runtime settles, followed by a caveat that `kcmd action-run` settles none of it and writes anyway.

This PR removes `kcmd action-run` and `kcmd action-list` from the CLI (`+136 / -1,848` lines across 19 files) while keeping the semantic runtime (`runAction`, `modelTools`, `callableTools`, `GeminiJudge`) intact:

- **CLI (`src/tool/main.ts`, `src/tool/commands.ts`, `tests/tool/action.test.ts`):**
  Remove `action-list`, `action-run`, and their CLI argument/formatting helpers. `kcmd agent-tools` and `kcmd profiles` remain the read-only inspection commands over a bound model.
- **Generated Agent Skills (`src/libts/semantic/skills.ts`, `src/libts/semantic/runtime/run_action.ts`):**
  Remove `runLine` and `runFlags`, and drop the `kcmd action-run` example block and the *"That command line settles no guard"* caveat from generated `SKILL.md` files.
- **Documentation (`docs/semantic-model/`, `demo/semantic-model/skill/README.md`):**
  Remove Section 7 (`## 7. Run it`) from `actions.md`, renumber `## 8. Hand it to an agent` to `## 7. Hand it to an agent`, fold the parameter-binding, single-transaction rollback, and `unknown` commit outcome guarantees into `### What a write tool and a lookup tool do`, and remove `action-list` / `action-run` references across `reference.md`, `skills.md`, `README.md`, `model_spec.md`, `fidelity.md`, and the commerce skill demo README.

## Verification
- `npx tsc --noEmit` and `npm run build` clean
- `npm test`: 1,057 pass / 0 fail across 39 files
- Markdown anchor sweep across `docs/semantic-model/` and `demo/semantic-model/skill/README.md` clean


libei (commented):
## Review

The deletion itself is clean — no correctness bug in the TypeScript, and the mechanical claims in the description all hold. What needs attention is prose the deletion orphaned, including one claim now written into every generated `SKILL.md`.

**Verified rather than eyeballed:** `tsc --noEmit` clean; `npm test` 1057 pass / 0 fail (1017 + 40), matching the description; 0 broken anchors across the doc set; `actions.md` headings run 1–7 with no dangling §8 reference anywhere; removed commands exit **1** with `Unknown command`, so a script calling them fails loudly; no anchor collision with #403. Both rewritten CLI transcripts reproduce byte-for-byte against live runs — regenerating the skill under both profiles yields exactly the documented `28,56d27` / `59c30` / `61c32`, and `kcmd agent-tools` in the demo scope matches its paste. The claim that no command line in this repository runs an action holds: `kcmd mcp` registers only `list-entries`, `lookup-entry` and `modify-entry` over a catalog snapshot, never the runtime.

### Worth fixing

**1. Every generated `SKILL.md` now overclaims enforcement.** [`skills.ts:517`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/src/libts/semantic/skills.ts#L517) writes "handed these actions as tools by its own framework, **which settles every guard before opening a transaction**". The runtime does not guarantee that. [`unsafeToRunUnchecked`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/src/libts/semantic/runtime/run_action.ts#L544-L547) filters `onViolation === 'warn'` constraints out of the refusal set, so an action whose guards are all advisory runs with no judge at all; and a judge call that fails on an advisory guard warns and lets the write land (`run_action.ts:647-649`). The text this replaced carried an explicit caveat. A reading agent meets this sentence before the outcome section that walks it back.

Suggest something closer to: *…which puts the action's guards to a judge before opening a transaction, and refuses rather than writing unchecked when it cannot settle one the model requires.*

**2. The paragraph that explained exactly that was deleted.** This PR removed **"A rule that never reached a judge is reported as unchecked"** and its `on_violation` routing from `actions.md` (it sat after *"That's the whole of what it has."*). That behaviour is still live — `run_action.ts:226` and `:647-649` — and `SKILL.md`'s outcome section still tells agents a warning can mean a rule went unchecked. Only the no-judge half survives, at `actions.md:1210` and `:1216`. Findings 1 and 2 compound: the explanation of when a guard goes unsettled was removed in the same change that asserted guards are always settled.

**3. [`demo/semantic-model/skill/README.md:379-382`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/demo/semantic-model/skill/README.md#L379-L382)** — "What changed is that the command line no longer asks: it proceeds, and the run reports the guards it passed over, which is the warning in the two runs above." Both referents are gone: the two committed transcripts were deleted, and no command line proceeds any more. It also contradicts the page's own opening claim at line 23 ("no command line in this repository currently runs an action or settles its rules"). Line 381 is the only surviving mention of guards being passed over.

**4. [`README.md:363`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/demo/semantic-model/skill/README.md#L363)** — "That is the same call that now commits and leaves the order at -$5.00." That outcome was only ever shown in the deleted `### What the command line settles: nothing`. Nothing on the page commits that call now and no `-$5.00` state appears anywhere.

**5. [`commerce.yaml:26`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/demo/semantic-model/skill/catalog/EntryGroups/commerce_demo/commerce.yaml#L26)** — not among the 19 changed files, and the only tracked file left naming a removed command: "`kcmd action-run` settles no guard at all -- it names them and writes anyway -- so running this demo from the command line reaches none of the rules below." Present tense, and after this PR nothing runs the demo from a command line at all — a stronger and different statement. Only that last sentence needs rewriting; the judge-needs-a-store point around it still stands.

### Minor

**6. [`validate.ts:185`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/src/libts/semantic/validate.ts#L185)** — `validateRunnable` now has zero production callers. On `main` its only one was `commands.ts:1790` inside `actionRun`, and it is not re-exported from `src/libts/index.ts`, so it is not embedder API either. This PR edited its comment to read "A caller running an action skips the deployment checks on purpose" — describing a caller that no longer exists. To be clear this is **not** a new runtime hole: `runAction` never called it on `main` either, and `validatePushRequirements` still covers the superset at push time. Either wire it into the runtime entry point or delete it.

**7. [`run_action.ts:549`](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/68db49556111701186c3134d29aadc16302a15e4/toolbox/mdcode/src/libts/semantic/runtime/run_action.ts#L549)** — "`kcmd` validates the model first, so this arrives only through the library entry point" is stale twice over: `kcmd` runs no actions now, so every call arrives through the library entry point and nothing validates first.

**8. `actions.md:1179` vs `reference.md:105`** — `actions.md` lists "It names a guard this runtime cannot settle" among the reasons a write tool is withheld; `reference.md` says "A guard is not one of those reasons." The paragraph that reconciled them ("A guard is not one of the reasons a tool is withheld. Who settles a rule belongs to the application that embeds the runtime…") was deleted here. Defensible on scope — one says *this runtime*, the other *this command* — but `kcmd agent-tools` passes `skipGuards: true` (pre-existing, `main:1484`), so a reader who runs it against a guarded action with no judge configured sees it listed runnable, which is what `actions.md` now reads as impossible. Judgment call.

Nit: `README.md:564` is a single 143-character prose line in a file that otherwise wraps at 80.

### Checked and not a finding

§4 does **not** use `CreditWithinOrderTotal`, `CreditUnderReviewThreshold` and `CreditIsNotSplitToAvoidReview` before introducing them — they appear in the model listing at `README.md:131-179`, well ahead of §4 at ~290. Un-eliding the "This call is gated by …" lines behind the `...` in the `kcmd agent-tools` paste would still read better, since the real command prints them, but it is not a defect.

PR #459, opened 2026-09-21
refactor(mdcode)!: remove kcmd agent-tools and unused TS runtime, center on skills-generate

libei:
Stacked on top of #458.

## Why

Before Agent Skills (`kcmd skills-generate`, #451) existed, `mdcode` was built around a TypeScript execution runtime (`src/libts/semantic/runtime/run_action.ts`, `judge.ts`, `src/libts/gcp/gemini.ts`, and `modelTools` in `agent_tools.ts`) meant to be imported by a TypeScript ADK agent (`demo/semantic-model/agent/`) or invoked by `kcmd action-run`, with `kcmd agent-tools` printing the ADK tool declarations to the terminal.

Once we moved to **generating Agent Skills (`kcmd skills-generate`)** and removed the ADK agent (#453) and `kcmd action-run` / `action-list` (#457):
- **`run_action.ts`, `judge.ts`, `src/libts/gcp/gemini.ts`, and `tool.invoke()` in `agent_tools.ts` were left behind with zero callers in `kcmd`.**
- **`kcmd agent-tools` and `kcmd skills-generate` presented two competing workflows and formats**, forcing Section 7 (`## 7. Hand it to an agent`) of `docs/semantic-model/actions.md` to spend 420+ lines explaining ADK pseudo-tools (`kcmd agent-tools`) and TypeScript runtime embedding (`Calling it from code`: `createSemanticRuntimes`, `modelTools`, `GeminiJudge`, `tool.invoke`) that do not exist when generating Agent Skills.

## What changed (`+240 / -4,582` lines)

1. **Remove `kcmd agent-tools` CLI command (`src/tool/main.ts`, `src/tool/commands.ts`)**:
   - `kcmd skills-generate` is now the single command for handing a semantic model to an agent.
2. **Remove the unused TypeScript action execution runtime and judge (`run_action.ts`, `judge.ts`, `gemini.ts`, `agent_tools.ts`)**:
   - Delete `src/libts/semantic/runtime/run_action.ts`, `src/libts/semantic/runtime/judge.ts`, `src/libts/gcp/gemini.ts`, and their test suites (`run_action.test.ts`, `gemini.test.ts`).
   - Simplify `src/libts/semantic/runtime/agent_tools.ts` to the pure model/profile metadata derivation used by `kcmd skills-generate` (`skills.ts`).
3. **Rewrite Section 7 of `docs/semantic-model/actions.md` around `kcmd skills-generate`**:
   - Show the single workflow (`kcmd skills-generate --out skills`), the generated `SKILL.md` router and `references/<action>.md` structure, the YAML-to-skill mapping table, and the 4 reasons an action is marked `Not runnable under this profile`.
   - Remove `kcmd agent-tools` and `GeminiJudge` references from `docs/semantic-model/reference.md`, `docs/semantic-model/skills.md`, and `demo/semantic-model/skill/README.md`.


toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -1106,339 +1029,118 @@ line of output per key:
...
+   SQL). All `statements` run in one read-write transaction against the
+   profile's store:
+   - Any statement error rolls the transaction back.
+   - An `UPDATE` or `DELETE` that matches zero rows also fails and rolls the
+     transaction back (`INSERT` is exempt), catching arguments that name a row
+     that isn't there.
+   - If the commit succeeds, `invoke` returns `applied: true`, `committedAt`,
+     and any `warnings`.
+   - If the commit hits a timeout or 5xx where the store may or may not have
+     applied the write, `invoke` returns `applied: 'unknown'` with an explicit
```
libei: **`applied: 'unknown'` is not a value the code can return.**

`ToolResult.applied` is typed `boolean`, and `describeOutcome` returns the indeterminate case as `{applied: false, unknown: true, reason, whatToDo: 'Do NOT retry…'}`.

An adapter written against this line branches on `applied === 'unknown'`, never matches, falls through to the `applied: false` arm and reads the outcome as "nothing was written" — then retries and applies the write twice. That is exactly the failure the `unknown` flag exists to prevent.

The PR description repeats it (`applied: true | false | 'unknown'`). The prose this replaced was careful here: "that last one comes back as `unknown`."

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -1106,339 +1029,118 @@ line of output per key:
...
+1. **Settle guards before opening a transaction.** Every constraint in the
+   action's `guards` list is put to the `judge` with the action's name,
+   description, and the call's scalar arguments (for example `source: 7`, the
+   value itself, not the `Account` row it identifies — see
+   [what a guard can see](#what-a-guard-can-see)).
+   - If any `reject` or `escalate` guard fails (or cannot be checked because no
+     judge was supplied or the model call failed), the call stops before opening
+     a transaction and returns `applied: false` with `reason` (the constraint's
+     name, rule text, and the judge's explanation) and `whatToDo` (the
+     constraint's `description`).
```
libei: **`whatToDo` is not the constraint's `description`.**

In `describeOutcome`, `whatToDo` is a fixed string — `'Correct what the reason describes, or report it. Nothing was written.'` The constraint's `description` is appended by `askJudges` into the refusal *message*, so it lands in `reason`, alongside the other three.

A caller that surfaces `whatToDo` to show the author's remediation line gets a generic runtime string instead. The sentence this replaced had it right: all four things — the constraint's name, your sentence, the judge's reason, and the `description` — are in the one refusal.

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -1106,339 +1029,118 @@ line of output per key:
...
+   a live caller's judge.)
+
+### What happens when a write tool runs
+
+Calling `tool.invoke(args)` runs the action in two steps:
+
+1. **Settle guards before opening a transaction.** Every constraint in the
+   action's `guards` list is put to the `judge` with the action's name,
+   description, and the call's scalar arguments (for example `source: 7`, the
+   value itself, not the `Account` row it identifies — see
```
libei: **The concurrency caveat is now gone from the whole doc set.**

The rewrite dropped "none is reached under the transaction's locks either: two calls racing each other are judged independently, and a rule that only holds when they are serialised holds for neither." `grep -rn 'serialis\|racing\|concurren' docs demo` returns zero hits.

§2 absorbed the stored-row case and the after-state case, but not this one — and the page's own worked example, rule 4 *not one credit split to evade review*, is precisely a rule that two concurrent calls each pass. The "a run doesn't compute the strictest outcome" note at line 673 doesn't cover it.

One sentence here, or in §2, restores it.

toolbox/mdcode/src/libts/semantic/runtime/judge.ts
```diff
@@ -61,3 +61,15 @@ export interface Judge {
   readonly name: string;
   decide(request: JudgeRequest): Promise<JudgeVerdict>;
 }
+
+/**
+ * Placeholder judge used when inspecting a model's derived tools (`kcmd
+ * agent-tools` and `kcmd skills-generate`) without invoking them.
+ */
+export const ASSUMED_JUDGE: Judge = {
```
libei: **Exporting a throwing placeholder into the module a caller shops for a `Judge` in.**

If `ASSUMED_JUDGE` ever reaches `runAction`, `askJudges` catches the throw. For a `warn` guard that becomes a warning and **the write commits**; for `reject`/`escalate` it refuses with an internal message a caller can't act on.

Nothing does this today, so it's latent — but it's in the guard path, and this file is exactly where someone looks for a judge to pass. Private to `skills.ts` it was structurally unreachable from a run; here it isn't.

Cheapest fix: brand it and have `runAction` reject a judge marked static, so misuse fails loudly instead of committing quietly.

Nit while you're here: the file now ends with two newlines.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -545,22 +441,11 @@ export interface CallableTools {
...
- * tool the runtime cannot run is still declared, still published and still
- * worth naming -- but offering it as callable spends a turn on a call that
- * cannot succeed and teaches the agent nothing it can act on. Lookups and
- * actions answer `runnable` on the same terms, so they are sorted together
- * rather than twice, and they keep the order `modelTools` gave them.
- *
- * What to do about `withheld` stays the caller's: print it, log it, refuse to
- * start. Dropping it in silence is the one thing this does not make easy.
  */
 export function callableTools(tools: ModelTools): CallableTools {
```
libei: **`callableTools` / `CallableTools` are dead — delete rather than keep.**

Zero production callers: `commands.ts` and `skills.ts` both destructure `modelTools` directly, `demo/semantic-model/` now contains only `skill/` (the agent demo is gone), and the doc no longer mentions it. The one reference left is `agent_tools.test.ts:606` asserting the pass-through.

The description calls it "a thin compatibility helper," but there's no consumer to be compatible with — and for a PR whose thesis is removing duplicate mechanisms, this is the duplicate left standing. Suggest dropping the function, the interface, and that test.

It also lost the rationale that was the useful part of it ("Dropping it in silence is the one thing this does not make easy"), so what remains is a forwarding call with no reason attached.

toolbox/mdcode/src/tool/commands.ts
```diff
@@ -1335,10 +1334,9 @@ export async function agentTools(options: AgentOptions = {}): Promise<number> {
     console.log(`  store: ${storeLine(store)}`);
     console.log();
 
-    const {lookups, actions, instruction} =
-        modelTools({runtime, skipGuards: true});
+    const {actions, instruction} =
+        modelTools({runtime, judge: ASSUMED_JUDGE});
```
libei: **This is the only production behavior change in the PR, and nothing tests it.**

`tests/tool/` covers `agent-tools` arg parsing and `--help` only; nothing asserts listing content. The doc's claim that `transfer_funds` is listed offered, guard and all, is verified nowhere — drop the `judge` argument here and every guarded action flips to `[NOT RUNNABLE]` with the suite still green.

A golden test over the listing, the same fixture pattern `skills-generate` already uses, closes it.

libei (commented):
Reviewed the top commit (`78c2999`); `6e5863f`/`d1e8b7c` are #458 below it in the stack. `tsc --noEmit` is clean and the full suite passes (1102 tests, 0 fail).

The direction is right: `skipGuards` is gone with no orphaned references, §2 genuinely reads self-contained now, and no stale anchors survive anywhere in `docs/` or `demo/`. One hunk is a real fix the description doesn't claim — the CLI listing now says "nothing puts it in front of you," matching the page's own YAML at line 382. It previously said "so read it before answering," which contradicted the model and was left over from removing the reading judge.

Six things below, most severe first. 1, 2 and 5 are mechanical; 3 is one sentence; 4 and 6 are judgment calls you may want to defer to a follow-up.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -172,47 +80,87 @@ export function actionTools(opts: ActionToolOptions): ActionTool[] {
...
+        `there is nothing to run. An executor is a physical binding: a ` +
+        `profile supplies one, and a profile that writes 'executor: null' ` +
+        `withdraws it. The action is still declared and still published; it ` +
+        `is only not performable here, and is performed somewhere else.`;
+  }
+  if (executor.kind !== 'sql') {
+    return `Action '${action.name}' is executed by ${
+               executor.kind
+                   .toUpperCase()}, which runs outside this transaction ` +
+        `and could not be rolled back if the commit failed. Supply a handler ` +
```
libei: **This tells the user to supply a `handler`, which this PR deletes — and it ships into the generated skill.**

After this change, `grep -rn 'ActionHandler\|handler' src` returns exactly two hits: a `// CLI command handlers` comment in `commands.ts`, and this string literal. The concept is gone; the advice to use it is not.

It isn't internal-only. The golden proves it reaches the artifact — `tests/libts/semantic/fixtures/actions_place_order.mcp.skill.golden.md:55`:

> `PlaceOrder` -- Action 'PlaceOrder' is executed by MCP, which runs outside this transaction and could not be rolled back if the commit failed. Supply a handler that performs the write as DML, or declare the action with a 'sql' executor.

So anyone generating a skill for an `mcp`/`rest`/`grpc` action is pointed at a seam that no longer exists in the tool. "Runs outside this transaction" has the same problem — there is no longer a transaction for it to run outside of.

The surviving half of the sentence is the good half: declare the action with a `sql` executor. Suggest reducing it to that.

toolbox/mdcode/docs/semantic-model/actions.md
```diff
@@ -399,9 +399,8 @@ That rule is the honest shape of a balance check and it is also one nothing here
 enforces: the balance is a row, a judge is shown the arguments and no more, and
 a rule it cannot settle it reports as not holding — which under `reject` refuses
 every call. Declared this way it is a policy on record and a schema constraint
-waiting to be written, not a control. Keep reading with that in mind; [a rule
-the judge can't settle](#a-rule-the-judge-cant-settle) is where it is worked
-through.
+waiting to be written, not a control. Keep reading with that in mind; [what a
+guard can see](#what-a-guard-can-see) explains why.
```
libei, line 404: **§2 still says kcmd executes the action. After this PR it doesn't.** (Anchored here because the lines themselves aren't in the diff — they're `actions.md:292` and `:303`, untouched by this commit, which is exactly why they're easy to miss.)

Line 292: "kcmd opens one transaction, runs the statements in the order you wrote them, and commits at the end."

Line 303, under the heading **A guard becomes a real gate**: "A `sql` action's guards settle before kcmd opens a transaction, and a call that doesn't clear them never reaches one, so a refusal leaves the store untouched."

The remaining commands are `cac`, `init`, `mcp`, `profiles`, `pull`, `push`, `skills-generate`. None runs an action or settles a guard. So the claim is false, and it is the load-bearing one — "a guard becomes a real gate" is the whole argument for carrying the write in the model.

It also contradicts this PR's own demo README at line 23, which states plainly that "no command line in this repository currently runs an action or settles its rules." Two files in one tree now say opposite things about the same tool.

toolbox/mdcode/src/libts/semantic/runtime/agent_tools.ts
```diff
@@ -494,104 +304,28 @@ export interface ModelTools {
...
- * start. Dropping it in silence is the one thing this does not make easy.
- */
-export function callableTools(tools: ModelTools): CallableTools {
-  const derived: DerivedTool[] = [...tools.lookups, ...tools.actions];
   return {
-    callable: derived.filter(tool => tool.runnable),
-    withheld: derived.filter(tool => !tool.runnable),
-    instruction: tools.instruction,
+    actions,
+    instruction: instructionFor(opts.runtime.model),
```
libei, line 313: **The generated instruction still teaches the deleted runtime's result contract.**

`instructionFor` (just below, unchanged here) tells the reading agent: "When a tool reports that a write did not happen, read the reason it gives and repeat it plainly; if it says a person has to decide, say so and stop" and "when a write did happen and the tool returns warnings…report both."

Those name specific artifacts of code deleted in this PR — `ToolResult.reason`, `ToolResult.warnings`, and the `escalate` appeal sentence that `askJudges` emitted. In the skills flow the agent writes through its own MCP database server, which returns SQL results, not the model's refusals or warnings. Nothing left in the tree can produce what this paragraph tells the agent to relay.

This ships in every `SKILL.md`, so it's the most-read paragraph the generator emits. Either trim it to what the skill flow actually produces, or say whose runtime is expected to honour it — the README is already careful to make that distinction, and this text isn't.

toolbox/mdcode/demo/semantic-model/skill/README.md
```diff
@@ -274,53 +274,27 @@ gcloud spanner databases execute-sql "$DATABASE" \
...
-
-  action  issue_credit  (IssueCredit)
-      Credit a customer against one order -- a late delivery, a coupon, a
-      shipping charge applied in error. The credit is added as a negative line
-      and the order total is recomputed from the lines.
-
-      This call is gated by CreditWithinOrderTotal, CreditUnderReviewThreshold,
-      CreditMemoNamesAServiceFailure, CreditIsNotSplitToAvoidReview:
-...
+$ kcmd profiles
```
libei, line 290: The swap to `kcmd profiles` is right, and marking the transcripts as recorded against the old CLI action runner is the honest call — that part reads well.

One thing the edit didn't reach: the headline caveat above it is now understated. Line 24 says "**The runtime** settles them when an application embeds it and gives it a judge, and no caller in this repository does today," and line 47 repeats it as "the runtime settles a rule only when its caller supplies a judge." Line 483 says an agent "should be handed these actions as tools by its own framework, which puts the action's guards to a judge before opening a transaction."

All three describe a runtime that exists but goes uncalled. This PR deletes it — `run_action.ts`, `judge.ts` and `gemini.ts` are gone, so there is nothing left to embed and no judge to give it. "No caller supplies a judge" and "the code that would settle a rule is no longer in this repository" are different admissions, and the second is the one that's true after this commit.

Line 483's guidance still stands as advice to a third-party framework; it just shouldn't read as describing something shipped here.

libei (commented):
Re-reviewed at `0117791`. The scope changed completely since my last pass — this now deletes the execution runtime rather than tidying it. Verified in a clean worktree: `tsc --noEmit` clean, full suite green (1001 tests, 0 fail; down from 1102, consistent with dropping `run_action.test.ts` and `gemini.test.ts`).

All six findings from the previous round are resolved, mostly by deletion — `applied: 'unknown'`, `whatToDo`, `callableTools` and `ASSUMED_JUDGE` now return zero hits across `src`, `tests`, `docs` and `demo`, and `kcmd agent-tools` is gone rather than untested. The new §7 is a real improvement, and I checked its four `Not runnable` reasons line by line against `whyRefusedWithoutRunning` — they match, including the non-advisory filter on both the blank-judgment and undeclared cases.

The theme of what's left is the one this PR is itself about: **deleting the runtime orphaned the prose that described it.** Four inline comments below. Two more clusters can't be anchored because the files aren't in the diff:

**Stale references to deleted modules.** `run_action` is named in `src/libts/gcp/alloydb.ts:296` and `:342`, `src/libts/semantic/ir.ts:493` ("see run_action.ts"), and `src/libts/semantic/runtime/store.ts:84`. `src/libts/semantic/runtime/runtime.ts:11` still describes deriving tools from "the same bound model `kcmd agent-tools` reads." `src/libts/semantic/skills.ts:25` contrasts the rendered tool with "the one that runs," which no longer exists. The one worth more than a comment fix is `tests/libts/semantic/actions.test.ts:1103` — it justifies its own narrowness with "these are the forms `run_action.test.ts` already drives through a `sql` [executor]," and that file is deleted here, so the coverage it defers to is gone.

**The write transport is now dead in production.** `beginReadWrite`, `executeSql`, `rollback` and `withSession` on `src/libts/gcp/spanner.ts` and `alloydb.ts` have no caller left in `src/` — `run_action.ts` was the only one. Their sole remaining exercise is their own unit tests in `tests/libts/gcp/spanner.test.ts`, which pass against a fake. That is the same "left behind with zero callers" condition the PR description gives as its reason for the deletions, so it reads as in scope; if you're keeping them deliberately as generic GCP clients for a future writer, worth saying so in the description so the next cleanup doesn't churn on it.

